PR #4, opened 2022-01-04
WIP: feat: display spendable amount on Send page

theborakompanioni:
Display spendable amount on the Send page.

Gotchas:
- Some utxos can be auto-frozen by jm. These are not spendable and cannot be "unfrozen" via the api at the moment.
  How should this be displayed to the user?

Still todo:
- reload available balances after payment

### DIFF
--- a/src/components/Send.jsx
+++ b/src/components/Send.jsx
@@ -1,8 +1,8 @@
 import React from 'react'
-import { useState } from 'react'
+import { useState, useEffect } from 'react'
 import { useLocation } from 'react-router-dom'
 import * as rb from 'react-bootstrap'
-import { serialize, ACCOUNTS } from '../utils'
+import { serialize, ACCOUNTS, btcToSats, valueToUnit, SATS } from '../utils'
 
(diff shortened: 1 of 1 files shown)


dergigi:
I'll review and test this PR as soon as the conflict is resolved.

theborakompanioni:
> I'll review and test this PR as soon as the conflict is resolved.

I am not happy with the approach here.

The thing is.. should we even try to determine the "spendable" amount in the ui?
It'd be good to get this value from the server, as any changes in the mechanism would make our calculations obsolete, maybe plain wrong.

dergigi:
> It'd be good to get this value from the server, as any changes in the mechanism would make our calculations obsolete, maybe plain wrong.

Agree, let's not do it client-side.

dnlggr:
> It'd be good to get this value from the server, as any changes in the mechanism would make our calculations obsolete, maybe plain wrong.

Yeah that probably make sense! I'm closing this PR then, feel free to reopen if needed.

PR #7, opened 2022-01-14
Add switch to show / hide balances

dnlggr:
Adds a simple switch to show / hide wallet balances on the main page.

Hiding sensitive information such as balances is one of the best practices [mentioned in the Bitcoin design guidelines](https://bitcoin.design/guide/onboarding/protecting-a-wallet/#quickly-hiding-balances). 

Nothing fancy, maybe we should make it possible to hide balances app-wide in the future but for now I think this might be good enough.

Screen_Recording_2022-01-14_at_11 41 41

### DIFF
--- a/src/components/CurrentWallet.jsx
+++ b/src/components/CurrentWallet.jsx
@@ -8,12 +8,18 @@ import DisplayUTXOs from './DisplayUTXOs'
 
 export default function CurrentWallet ({ currentWallet }) {
   const [walletInfo, setWalletInfo] = useState(null)
+  const [redactBalances, setRedactBalances] = useState(!(window.localStorage.getItem('jm-redactBalances') === 'false'))
   const [unit, setUnit] = useState(window.localStorage.getItem('jm-unit') || BTC)
   const [utxos, setUtxos] = useState(null)
   const [fidelityBonds, setFidelityBonds] = useState(null)
   const [showUTXO, setShowUTXO] = useState(false)
   const [alert, setAlert] = useState(null)
   const [isLoading, setIsLoading] = useState(false)
+
+  const setAndPersistRedactBalances = redactBalances => {
+    setRedactBalances(redactBalances)
+    window.localStorage.setItem('jm-redactBalances', redactBalances)
+  }
 
   const setAndPersistUnit = unit => {
     setUnit(unit)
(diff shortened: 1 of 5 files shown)


theborakompanioni:
+1 for hiding balances by default. :100:

src/components/CurrentWallet.jsx
```diff
@@ -8,13 +8,19 @@ import DisplayUTXOs from './DisplayUTXOs'
 
 export default function CurrentWallet ({ currentWallet }) {
   const [walletInfo, setWalletInfo] = useState(null)
+  const [redactBalances, setRedactBalances] = useState(!(window.localStorage.getItem('jm-redactBalances') === 'false'))
```
dennisreimann, line 11: > +1 for hiding balances by default. 100

Agreed, let's also rename the property to `showBalances`.

src/components/CurrentWallet.jsx
```diff
@@ -77,17 +83,18 @@ export default function CurrentWallet ({ currentWallet }) {
         </div>}
       {walletInfo && walletInfo?.total_balance &&
         <>
-          <p>Total Balance: {valueToUnit(walletInfo.total_balance, unit)}</p>
+          <p>Total Balance: {valueToUnit(walletInfo.total_balance, unit, redactBalances)}</p>
+          <rb.Form.Check type="switch" label="Show Balances" checked={!redactBalances} onChange={(e) => setAndPersistRedactBalances(!e.target.checked)} />
           <rb.Form.Check type="switch" label="Display amounts in SATS" checked={unit === SATS} onChange={(e) => setAndPersistUnit(e.target.checked ? SATS : BTC)} />
```
dennisreimann, line 88: Let's apply a margin at the botton here:

```suggestion
          <rb.Form.Check type="switch" label="Display amounts in SATS" checked={unit === SATS} onChange={(e) => setAndPersistUnit(e.target.checked ? SATS : BTC)} className="mb-4" />
```

dnlggr:
Yep agreed, it's already hidden by default!

src/components/CurrentWallet.jsx
dnlggr: Updated. ✅

dnlggr:
Thanks for the reviews guys! All feedback addressed. ✅

src/utils.js
```diff
@@ -28,7 +28,11 @@ export const btcToSats = value => Math.round(parseFloat(value) * 100000000)
 
 export const satsToBtc = value => parseInt(value, 10) / 100000000
 
-export const valueToUnit = (value, unit) => {
+export const valueToUnit = (value, unit, showValue = true) => {
```
dennisreimann, line 31: I'd say we should then default this one to false as well:

```suggestion
export const valueToUnit = (value, unit, showValue = false) => {
```

dnlggr: Yeah that's a good point! 👍  I initially thought I don't want to change it for other pages but this function is not used anywhere else than on the main wallet page so we can definitely change the default here. 

Updated. ✅

dennisreimann (approved):

PR #11, opened 2022-01-24
feat: add sat symbol & update balance UI

dnlggr:
Updates the balance text a bit to more closely resemble the [screens in Figma](https://www.figma.com/file/kfejZJFlwBywvLEnPEmJo1/JoinMarket-UI?node-id=2850%3A67638) and the Bitcoin Design guidelines on [units and symbols](https://bitcoin.design/guide/payments/units-and-symbols/) and on [protecting a wallet](https://bitcoin.design/guide/onboarding/protecting-a-wallet/).

Basically, I turned on some [font features](https://rsms.me/inter/#features) according to the Figma designs and added a better symbol for when the balance is hidden.

It's not perfect yet, I also want to add:
- The possibility to cycle through states (BTC, sats, hidden) by clicking on a balance. For that we need some global state and I didn't want to touch that quite yet.
- Formatting based on the user’s locale. Currently I hardcoded the _BTC_ symbol to be prefixed to the number (as it is done in the U.S. with USD amounts. However, I didn't know what to do with the _sats_ symbol as it looked strange when prefixed to a number so I kept it as a postfix. Open for suggestions on this one!

For a first version it should be fine though. Let me know what you think!

Not sure how we want to organize custom CSS going forward. I used a [CSS Module](https://create-react-app.dev/docs/adding-a-css-modules-stylesheet) but am open to do it differently or just add everything to `index.css` for now.

## 📸 

<table>
<tr>
<td width="33%">

</td>
<td width="33%">

</td>
<td width="33%">

</td>
</tr>
</table>


dnlggr:
See also https://github.com/joinmarket-webui/joinmarket-webui/issues/19.

dergigi:
Wonderful! 

Regarding the sat symbol: so far most interfaces still use "sat" or "sats," i.e. no symbol at all. Thunderhub—which looks quite neat in general I have to say—uses the sat symbol, and simply puts it at the end using a slightly dimmer font color:
Screenshot 2022-01-25 at 19 57 34

Most other interfaces, including bitcoin and lightning wallets, simply use "sat" or "sats" either before or after the number (usually after).
[1ML](https://1ml.com/):
Screenshot 2022-01-25 at 20 04 17

[LightningNetwork.plus](https://lightningnetwork.plus/):
Screenshot 2022-01-25 at 20 03 56

dnlggr:
Updated to use the sat symbol (resolves #19).

The [bitcoin design guidelines](https://bitcoin.design/guide/payments/units-and-symbols/) don't really use a specific sat symbol so maybe we should let the user pick if they want a symbol or just "sat(s)" in the upcoming settings page. Maybe @GBKS has some input on that? For now, I'd say let's keep it simple and just go with the symbol.

I also added digit groups using [`Intl.NumberFormat`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat) as suggested by the design guides. For now I hardcoded the formatting to the `en-US` locale but we could do more location specific things in the future.

Regarding[ browser compatibility of `Intl.NumberFormat`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat#browser_compatibility): It should be available on all somewhat modern browsers. Does someone have any experience with that? And a more broad question: What browser compatibility are we targeting in general?


<table>
<tr>
<td width="33%">

</td>
<td width="33%">

</td>
<td width="33%">

</td>
</tr>
</table>

src/index.css
```diff
@@ -178,6 +178,12 @@ Usage:
   src: url('./fonts/Inter-italic.var.woff2?v=3.19') format('woff2');
 }
 
+/* https://satsymbol.com/ */
+@font-face {
+  font-family: 'Satoshi Symbol';
+  src: url("./fonts/SatoshiSymbol.otf") format('opentype');
```
dergigi, line 184: :+1:

dergigi (approved):
Looks good to me :+1:

dergigi:
Great! Let's keep it as is for now. We can do some polishing for v4 later on, I'm sure we can improve things here and there. Different font color for the symbol would also be easier on the eyes, for example.

PR #21, opened 2022-01-26
feat: setup code formatting

dnlggr:
This is a request for comments on code formatting. I personally like having a good, opinionated formatter setup so that I don't have to worry about formatting, indenting, etc. as I write code. I'd love to hear your opinions. Don't want to force anything on anybody.

This PR sets up [Prettier](https://prettier.io/) which is—I believe—one of the standard JavaScript formatters.

I've added two npm scripts:
- `npm run lint`: Runs Prettier and prints out any files violating the code format
- `npm run format`: Runs Prettier and reformats any files violating the code format

I've also added husky which (optionally) let's you format all changed files as a git pre-commit hook. To activate this, simply run `npm run prepare`. That'll install the pre-commit hook.

We could also run code formatting via the GitHub Actions CI if we wanted to.

Let me know what you think! Happy to hear any feedback and concerns!


dennisreimann:
Can we keep the no semicolons and single quote settings?

package.json
```diff
@@ -37,14 +40,20 @@
...
+    "format": "prettier --write --no-error-on-unmatched-pattern src/**/*.{js,jsx,json,css,md}"
   },
   "eslintConfig": {
     "extends": [
       "react-app",
       "react-app/jest"
     ]
   },
+  "prettier": {
+    "printWidth": 120
```
dennisreimann, line 55:
```suggestion
    "printWidth": 120,
    "semi": false,
    "singleQuote": true
```

Pretty please :wink:

dergigi:
Big +1 for that on my side. I've used prettier before and had no issues with it.

I don't have strong opinions on the JS specifics, so no semicolons and single quote is fine with me.

dnlggr:
> Can we keep the no semicolons and single quote settings?

Absolutely!

dnlggr:
Updated to single quotes and no semicolons, let me know if I should add anything else! :)

dnlggr:
Does anyone have experience with setting up linting for an open source project? I have a Prettier plugin for my editor but I figured I'd also add the (optional) pre-commit hook option for anyone who prefers that. Ultimately though, we can't force anyone to use formatting of course. That's fine for me. The only thing that could be annoying is that if someone who has the formatter setup touches a file that wasn't formatted before (i.e. contributed by someone without a formatter) the whole file will get formatted and there will be lots of unrelated changes in the PR. We can either:

- Just live with that.
- Do some kind of CI formatting to make sure `master` is always formatted.

Any thoughts? I feel like I might be overthinking it. 😄

dennisreimann (approved):

dennisreimann:
@dnlggr I think having the pre-commit hook is fine. We can also add a linting/formatting check step to the build, so that PRs aren't mergeable without being properly formatted.

dnlggr:
> We can also add a linting/formatting check step to the build, so that PRs aren't mergeable without being properly formatted.

Sounds good! 👍

PR #29, opened 2022-01-28
feat: style navbar

### LINKED ISSUE #20: Header with tabs: Send, Receive, Earn
See the [Main V2](https://www.figma.com/file/kfejZJFlwBywvLEnPEmJo1/JoinMarket-UI?node-id=2850%3A68011) page in Figma:

Screenshot from 2022-01-26 16-15-34

Rough component outline:
- [x] Header component (added in #29)
- [x] Wallet name & balance (added in #29)
- [x] Send / Receive / Earn (added in #29)
- [x] Settings (gear icon) (added in #26)

dnlggr:
Resolves #20.

Should be pretty flexible and close to the design. Let me know what you think!

I added one additional button to the right of the navbar to show the "wallets selector" screen. In the designs this button isn't in the navbar but in the main screens. Since we didn't implement those screens yet I put the button in the navbar for now so that we have a way to still show the "wallet selector" screen.

## 📸 

Screenshot 2022-01-31 at 14 31 31

Screenshot 2022-01-31 at 14 31 28


dergigi:
Awesome! Looks great already! :orange_heart:

src/components/CurrentWallet.jsx
```diff
@@ -50,24 +48,17 @@ export default function CurrentWallet({ currentWallet }) {
     Promise.all([loadingWallet, loadingUtxos]).finally(() => setIsLoading(false))
 
     return () => abortCtrl.abort()
-  }, [currentWallet])
+  }, [currentWallet, setWalletInfo])
```
dnlggr, line 51: ESLint wanted me to add this as a dependency (which make sense I guess). I'm not quite sure if I need to wrap it in a `useCallback()` hook to memoize it, though. 

It doesn't cause endless re-renders (I checked that) but I also couldn't find out if the `useContext()` hook (which `setWalletInfo` comes from, see [here](https://github.com/joinmarket-webui/joinmarket-webui/pull/29/files#diff-95bcf45bfc1b1047b24c7448e4036322edfe04e6bcaec7838b986472954725a6R40)) guarantees a stable identity of its return value across re-renders. It seems like it does (otherwise endless re-renders would occur) but the docs didn't mention that.

@dennisreimann if you know of any standard way of handling that, please let me know! 🙏

dnlggr:
Damn, I forgot the activity indicators for joining and earning. I'll work on them now and push them either to this PR or make a separate PR depending on the reviews this received until then.

dnlggr:
Added the activity indicators. For the "Earn" tab I used the one from #20. For the "Coinjoin in Progress" I used the same we had before.


Screenshot 2022-01-31 at 16 57 58
Screenshot 2022-01-31 at 16 58 10
Screenshot 2022-01-31 at 16 58 39

theborakompanioni:
Looks nice and clean! Great work! Would love to get it merged as soon as possible to make Screenshots for Umbrel : D

One privacy issue that came to light when displaying the balance in the navbar: Suppose `showBalances` is `true` and a user locks a wallet - without closing the browser. Another user creates a new wallet -> the balance of the previously opened wallet will be leaked.

src/components/Balance.module.css
```diff
@@ -1,9 +1,14 @@
 .balance-wrapper {
   display: inline-flex;
-  align-items: baseline;
+  align-items: center;
+
+  min-height: 23px;
```
theborakompanioni, line 5: I really do not know if it is this value that causes the misbehaviour: In Chromium (v97) and in Brave (v1.33, based on Chromium v96) the wrapper div changes it size when toggling `showBalances` on the settings screen. Does not happen in Firefox (v95). Can you reproduce this?

dnlggr: Good find! The sat symbol and the "balance hidden" symbol had different heights. Fixed! ✅

dnlggr:
> One privacy issue that came to light when displaying the balance in the navbar: Suppose showBalances is true and a user locks a wallet - without closing the browser. Another user creates a new wallet -> the balance of the previously opened wallet will be leaked.

I would say we should probably not support a multi user setup—at least for now. Nevertheless this is an interesting issue! Probably related to how the wallet contents is only loaded in the `CurrentWallet` component. I'll have a look!

dnlggr:
> > One privacy issue that came to light when displaying the balance in the navbar: Suppose showBalances is true and a user locks a wallet - without closing the browser. Another user creates a new wallet -> the balance of the previously opened wallet will be leaked.
> 
> I would say we should probably not support a multi user setup—at least for now. Nevertheless this is an interesting issue! Probably related to how the wallet contents is only loaded in the `CurrentWallet` component. I'll have a look!

Should be fixed now! When locking a wallet we now also reset the `currentWalletInfo` not just the `currentWallet` state. ✅

theborakompanioni:
Sat symbol seems to have changed visually. 

Before:
image

After:
image

The previous one appears to be more friendly to human eyes. What do you think?

theborakompanioni (approved):
Found no more issues that are in the scope of this PR.
Visually a nice improvement and the code looks really great too.

:rocket:

dnlggr:
> Sat symbol seems to have changed visually.

Yeah we switched to the one from [bitcoinicons.com](https://bitcoinicons.com). I feel like this might even be a bug. I'll check that out and open an issue on their end if needed.

dnlggr:
> > Sat symbol seems to have changed visually.
> 
> Yeah we switched to the one from [bitcoinicons.com](https://bitcoinicons.com) a while ago. I feel like this might even be a bug. I'll check that out and open an issue on their end if needed.

Yep this is a bug in bitcoinicons.org. See https://github.com/BitcoinDesign/Bitcoin-Icons/issues/79. I'll fix that manually for us. But not in here.

theborakompanioni:
What do you think about removing the unfilled area until a balance is loaded?

image
image

dennisreimann:
The mobile version can be improved, but let's do that in a separate PR.

image

Is there a reason why the offcanvas got traded for the navbar collapse version? I think the offcanvas would work well and we could present it below the header. Maybe this is something we can discuss in the next meeting.

Other than that it works and it a great improvement!

dennisreimann:
Added a minor visual improvement to take the activity indicator out of the display flow, so that the bottom line doesn't seem a bit longer on the right when the Earn tab is active.

dennisreimann (approved):

dergigi (approved):
Great work! Massive improvement. Code looks good to me ✅ 

Tested it on Firefox 96.0.3 (64-bit) in both desktop and mobile mode.

dnlggr:
> Added a minor visual improvement to take the activity indicator out of the display flow, so that the bottom line doesn't seem a bit longer on the right when the Earn tab is active.

Awesome thanks!

> The mobile version can be improved, but let's do that in a separate PR.

Agreed! Will take care of that.

> Is there a reason why the offcanvas got traded for the navbar collapse version?

No not really, probably got changed more-or-less by accident when I re-did the navbar. I can switch it back, no problem. 👍

dnlggr:
Thanks for the reviews guys!

dnlggr:
Tracked all remaining open improvement ideas here: https://github.com/joinmarket-webui/joinmarket-webui/issues/41

PR #33, opened 2022-01-31
fix: initialize local storage with default values

### LINKED ISSUE #32: Initial settings error
User opening the app for the first time will have no default config stored in local storage leading to display issues (e.g. when showing balances).

Reproduce: Delete local storage data and reload.
Fix: Always take default values into account when initializing the settings for the first time.

theborakompanioni:
Always take default values into account when initializing the settings for the first time, as it might already contain some values (e.g. from `prefers-color-scheme` handling)

Fixes #32.

### DIFF
--- a/src/context/SettingsContext.jsx
+++ b/src/context/SettingsContext.jsx
@@ -1,10 +1,9 @@
 import React, { createContext, useReducer, useEffect, useContext } from 'react'
 import { BTC } from '../utils'
 
-const localStorageKey = 'jm-settings'
+const localStorageKey = window.JM.SETTINGS_STORE_KEY
 
 const initialSettings = {
-  theme: 'light',
   showBalance: false,
   unit: BTC,
 }
--- a/src/context/SettingsContext.jsx
+++ b/src/context/SettingsContext.jsx
@@ -4,6 +4,7 @@ import { BTC } from '../utils'
 const localStorageKey = 'jm-settings'
 
 const initialSettings = {
+  theme: 'light',
   showBalance: false,
   unit: BTC,
 }
@@ -22,7 +23,7 @@ const settingsReducer = (oldSettings, action) => {
 const SettingsProvider = ({ children }) => {
   const [settings, dispatch] = useReducer(
     settingsReducer,
-    JSON.parse(window.localStorage.getItem(localStorageKey)) || initialSettings
+    Object.assign({}, initialSettings, JSON.parse(window.localStorage.getItem(localStorageKey)))
   )
 
   useEffect(() => {


src/context/SettingsContext.jsx
```diff
@@ -4,6 +4,7 @@ import { BTC } from '../utils'
 const localStorageKey = 'jm-settings'
 
 const initialSettings = {
+  theme: 'light',
   showBalance: false,
   unit: BTC,
```
theborakompanioni, line 9:
```suggestion
  unit: SATS,
```
Since these default values are already touched.. what do you think about making sats the standard?

src/context/SettingsContext.jsx
```diff
@@ -4,6 +4,7 @@ import { BTC } from '../utils'
 const localStorageKey = 'jm-settings'
 
 const initialSettings = {
+  theme: 'light',
```
dennisreimann, line 7: This one's a bit tricky, because the default depends on the system preferences of the user. (see also index,html)

```suggestion
  theme: window.matchMedia('(prefers-color-scheme: dark)').matches ? window.JM.THEMES[1] : window.JM.THEMES[0];
```

src/context/SettingsContext.jsx
dnlggr: I feel like for JoinMarket most people use larger amounts (in the millions of sats) since super small amounts don't really make sense because of relatively high fees (compared to a lightning wallet where fees are in the single sat range at most). The [Bitcoin design guidelines](https://bitcoin.design/guide/payments/units-and-symbols/#recommended-interaction) say something along the lines of "sats for daily spending and BTC for savings applications". I guess JM is somewhere in between. I'd keep BTC the standard but definitely won't die on that hill. 😄

src/context/SettingsContext.jsx
dnlggr: I think we don't even need this as it already gets set [in `index.html`](https://github.com/joinmarket-webui/joinmarket-webui/blob/master/public/index.html).

src/context/SettingsContext.jsx
```diff
@@ -22,7 +23,7 @@ const settingsReducer = (oldSettings, action) => {
 const SettingsProvider = ({ children }) => {
   const [settings, dispatch] = useReducer(
     settingsReducer,
-    JSON.parse(window.localStorage.getItem(localStorageKey)) || initialSettings
+    Object.assign({}, initialSettings, JSON.parse(window.localStorage.getItem(localStorageKey)))
```
dnlggr, line 26: Good catch! That would only update when there's not settings at all (which is never the case since the theme gets set in `index.html` already—see above).

dnlggr (commented):
Nice, good catch! Thanks for cleaning that up! 👏 👍 

I don't think we need the theme change though. See inline comment.

src/context/SettingsContext.jsx
theborakompanioni: Yes, I did set it do 'light' even if it is done in index.html -> kind of has a documentary purpose. But I can remove it, if you want to. I kind of like to see all the default settings (unless overwritten, of course).

src/context/SettingsContext.jsx
theborakompanioni: I can understand this reasoning. Thanks for the explanation (and the link!). :muscle:

src/context/SettingsContext.jsx
dennisreimann: I think it makes sense to have it removed to not duplicate code that might get out of sync.

theborakompanioni: Removed! Thank you both for your inputs!

dnlggr (approved):

PR #37, opened 2022-02-01
feat: developer docs

dnlggr:
Moved the development related part of the readme to a dedicated DEVELOPING.md. Also added some more details on teh code formatting approach and a link to some troubleshooting tips. Thanks to @dennisreimann for the hint! 🙏

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,6 +42,7 @@
     "test": "react-scripts test",
     "eject": "react-scripts eject",
     "setup-format-hook": "husky install",
+    "lint": "prettier --check --no-error-on-unmatched-pattern src/**/*.{js,jsx,json,css,md}",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -41,7 +41,7 @@
     "build": "react-scripts build",
     "test": "react-scripts test",
     "eject": "react-scripts eject",
-    "prepare": "husky install",
+    "setup-format-hook": "husky install",
```
dennisreimann, line 44: I think we could make this a [`postinstall` script](https://docs.npmjs.com/cli/v6/using-npm/scripts#npm-install), to ensure it gets set up for everyone. Not sure if this is the ideal way to solve this, but I think it should be somehow automated.

dnlggr: Agreed. Initially I didn't want to force it on anybody but if we all agree on it, I'm fine with automating the install! Then we can just forget about it.

dennisreimann (approved):

PR #38, opened 2022-02-01
feat: do not autocomplete password when creating wallet

theborakompanioni:
There are some pros and cons for password auto-completion, however on the create wallet page it should propably not prefill with an existing password - what do you think? 
Most modern browsers will optionally open a dialog to choose from.

Before:
image

After:
image

See [info about autocomplete attribute and login fields from Mozilla](https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion#the_autocomplete_attribute_and_login_fields)


dennisreimann (approved):
Good catch!

dergigi (approved):
LGTM ✅

PR #39, opened 2022-02-01
feat: Adapt favicon color based on system theme

dennisreimann:
Display a light version of the favicon if the system theme is dark (and hence there's a dark browser window), use a dark version if the system theme is light. 

Unfortunately this doesn't work in Chrome (and potentially other Webkit-based browsers), but it works at least in Firefox and it doesn't make things worse for other browsers, as the dark version will always be displayed (as before),


theborakompanioni (approved):
Neat! Tested with Firefox (:heavy_check_mark:), Brave (:heavy_check_mark:) and Chromium (:x:).

dergigi (approved):
LGTM ✅  

Tested on Firefox.

PR #40, opened 2022-02-01
docs: Docker README improvements

dennisreimann:
Minor typo fixes and an update for the rebuild command: I had to use the `--no-cache` option to get the current latest master version of JoinMarket, without it it was using the previous, cached version of their master.


theborakompanioni (approved):
Thank you! :pray:

dergigi (approved):

PR #48, opened 2022-02-04
fix: navbar padding

dnlggr:
Fixes inconsistent padding in the navbar which caused ugly jumping effects when switching from the non-active to the active state of the wallet preview in the navbar.

🎥 **Before:**

Screen_Recording_2022-02-04_at_09 29 03

🎥 **After:**

Screen_Recording_2022-02-04_at_09 28 37


theborakompanioni (approved):
lgtm! :+1:

dennisreimann (approved):

PR #49, opened 2022-02-04
fix: display amount of addresses correctly

theborakompanioni:
Quick fix to display the correct amount per address.

Small bug seems to have been introduced while refactoring that will led to the balance of addresses always end up displaying the whole account balance.


dnlggr (approved):

dnlggr:
Thanks!

PR #50, opened 2022-02-04
Specter Integration

k9ert:
With our newest `v1.8.x` release there is the possibility to integrate via extensions. Components which are client-only are specifically easy to integrate. I spend about 2h creating this. I don't had a joinmarket-dev-env running so this might even not work at all and i'm also not very familiar with react.

Here is a gif showing how it looks like. In the `README.md` at the very bottom, it's described how to get it to run.

joinmwebui-specter

What do you think about such an integration? Would it be interesting to collaborate here?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,8 @@
 .env.production.local
 
 npm-debug.log*
+
+# python stuff
+.env
+joinmarketwebui/static
+__pycache__
--- a/README.md
+++ b/README.md
@@ -128,3 +128,18 @@ npm install && npm start
 ## 👨‍💻 Developing
 
 See [DEVELOPING.md](DEVELOPING.md) for developer docs.
+
+## Specter integration
+
+
+```
+npm run build
+virtualenv --python=python3 .env
+# . ./.env/bin/activate
+# pip3 install -r requirements.txt --pre
+. ../specter-desktop/.env/bin/activate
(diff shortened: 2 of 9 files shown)


dergigi:
@k9ert can you explain how this integration will work in detail, from the user's perspective?

k9ert:
To be honest, i don't know yet. One the one hand side, this is not much different than an integration with the node-providers (umbrel/rapsiblitz/...). Specter from that angle is just a way of shipping the software to the user. We had some discussion about security concerns but i think i've found a way to overcome them with a [piggiback-extension](https://github.com/cryptoadvance/specter-desktop/pull/1561). If we would keep it like that, one possible user-story would be like:
* The user is downloading one of the specter-apps for his platform (windows/öinux/macOS)
* The user starts specter
* Like in the above animated gif he'll find the joinmarket-WEBUI in the extensions
* He can use it right away if he fulfilled the other dependencies
* There is somewhere a button which brings him back to specter

Potentially, we can do more integrations even if this stays a piggyback-extension:
* We could pass Bitcoin-Core credentials to the webui
* Ups, that's a short list

Maybe let me explain/discuss in the next devcall.

dergigi:
Thank you for the background info.

> Maybe let me explain/discuss in the next devcall.

Sounds good! Looking forward to it!

theborakompanioni:
Hey @k9ert, are you still working on this?
I'd close this for now - please reopen if you want to continue! :pray:

PR #51, opened 2022-02-04
feat: privacy levels

### LINKED ISSUE #27: Revamp wallet view: display "accounts" as privacy levels
See the current design in [Figma](https://www.figma.com/file/kfejZJFlwBywvLEnPEmJo1/JoinMarket-UI?node-id=2850%3A68011):

Screenshot 2022-01-28 at 15 53 21

In short:
* Hide derivation paths and addresses from the user
* What is currently called "Account 0" is the lowest level, i.e. "one shield"

We can add the complexity that this will hide later on if we like, but the current view is all kinds of confusing, so this would be an improvement. 

Note that this concept/design is by no means final, since we can't always map "privacy level" to "accounts" (mixdepths) in a truthful one-to-one manner. But again: it's a big improvement from the current UI and should be easier to understand than mixdepths.

dnlggr:
Resolves #27. ✨ 

This is a first step towards a more automated, less manual UX. By hiding derivation paths and addresses from the user and mapping accounts/mix depths to privacy shields as it is [in Figma](https://www.figma.com/file/kfejZJFlwBywvLEnPEmJo1/JoinMarket-UI?node-id=2850%3A68011).

When you click _Deposit_, you'll be shown a fresh address from (always) mix depth 0.

I also added a settings toggle to switch from this "magic wallet view" (please tell me if you have a better name for this) and the "advanced wallet view" (which is the same we had before, showing all mix depths, addresses, etc.).

For now, mix depths are mapped directly to privacy levels. We're aware that this mapping may not always be correct (e.g. the user could manually deposit non-private coins to any mix depth; the tumbler cycles through the mix depths which can lead to coins in mix depth 0 which are in fact more private than coins that have just been deposited there. See also #27. We should definitely document this and think about how we want to go froward from here.

We should also think about how to visualize the different privacy levels when withdrawing. Currently, the user can simply pick from any account when sending coins.

This is very simple and doesn't have much functionality yet. It's something we can build upon, though. Props to @GBKS for the UI. I think it looks super fresh and clean. 👏 

Let me know what you think!

## 📸 
Screenshot 2022-02-04 at 15 34 53
Screenshot 2022-02-04 at 15 35 09
Screenshot 2022-02-04 at 15 52 09
Screenshot 2022-02-04 at 16 00 18


src/components/DisplayAccounts.jsx
```diff
@@ -49,14 +49,15 @@ export default function DisplayAccounts({ accounts, ...props }) {
...
-                  {entries.map(({ address, amount, hd_path: hdPath, labels }) => (
+                  {entries.map(({ address, amount, hd_path: hdPath, labels, status }) => (
                     <rb.Row key={address} className="w-100 mt-3">
                       <rb.Col xs="auto">
                         <code className="text-break">{hdPath}</code>
                       </rb.Col>
                       <rb.Col>
                         <code className="text-break">{address}</code>{' '}
                         {labels && <span className="badge bg-info">{labels}</span>}
+                        {status && <span className="badge bg-info">{status}</span>}
```
dnlggr, line 60: The `satus` of an address is if it is _new_ or _used_. This must've been the `label` in older API versions. I'm pretty sure I saw that info in the UI before. The label seems to be empty now. I added the status to the UI to make sure it's shown in any case.

src/components/Receive.jsx
```diff
@@ -71,6 +73,7 @@ const Receive = ({ currentWallet }) => {
           onChange={(e) => setAccount(parseInt(e.target.value, 10))}
           style={{ maxWidth: '21ch' }}
           required
+          disabled={!settings.useAdvancedWalletMode}
```
dnlggr, line 76: If in "magic wallet mode", don't let the user choose any other mix depth than 0️⃣  for receiving.

theborakompanioni: Great idea! Have you thought about not displaying the select box in the first place (in this mode)?

src/components/CurrentWalletMagic.jsx
```diff
@@ -0,0 +1,159 @@
...
+  const filledShields = Array(level + 1)
+    .fill()
+    .map((_, index) => {
+      return <Sprite key={index} symbol="shield-filled" width="24" height="30" />
+    }) // todo: no IE support. fine?
+  const outlienedShields = Array(5 - filledShields.length)
+    .fill()
+    .map((_, index) => {
+      return <Sprite key={index} symbol="shield-outline" width="24" height="30" />
+    }) // todo: no IE support. fine?
```
theborakompanioni, line 47: >  "no IE support. fine?"

I'd say it's fine. 

We have this in `package.json`:
```
"browserslist": {
    "production": [
      ">0.2%",
      "not dead",
      "not op_mini all"
    ],
    "development": [
      "last 1 chrome version",
      "last 1 firefox version",
      "last 1 safari version"
    ]
  },
```

theborakompanioni:
Looking really great so far. Excellent  :muscle: 

Two notes from a quick and dirty review for now: 

- Shortly before loading indicator appears the view is briefly rendered ("flickers")
- Have you tried not showing "only" a loading indicator while fetching data? 
  - Should we maybe display some kind of placeholder (e.g. something like [this](https://github.com/zalog/placeholder-loading))
    the first time, then show loading indicator + "previous" content while loading new data (similar to the `CurrentWalletAdvanced` screen)?

src/components/Receive.jsx
dnlggr: I thought about it yeah. We should probably revamp both the send / receive screens a bit according to Figma designs and improve wording as well as things such as hiding accounts when they're not needed. Let's talk about that in today's call!

dnlggr:
> Should we maybe display some kind of placeholder animation (e.g. something like [this](https://github.com/zalog/placeholder-loading))
the first time, then show loading indicator + "previous" content while loading new data (similar to the CurrentWalletAdvanced screen)?

That is a very good point! I'll definitely track that as separate issue. 🎫

dnlggr:
> > Shortly before the loading indicator appears the view is briefly rendered ("flickers")
>
> True. No idea why that happens. Will look into it!

Alright, I think that should be fixed now. The reason was that by `isLoading` was `false` by default and then set to `true` in the `useEffect()` hook immediately after render. But for the short amount of time between the first render and the hook being called it was `false` causing the UI to be rendered only to be replaced with the loading indicator shortly after.

I set `isLoading` to default to `true` which should fix the issue.

src/components/Receive.jsx
dnlggr: I tracked https://github.com/joinmarket-webui/joinmarket-webui/issues/57.

src/components/CurrentWalletMagic.jsx
```diff
@@ -0,0 +1,159 @@
...
+
+const PrivacyLevel = ({ level, balance }) => {
+  const settings = useSettings()
+
+  const filledShields = Array(level + 1)
+    .fill()
+    .map((_, index) => {
+      return <Sprite key={index} symbol="shield-filled" width="24" height="30" />
+    })
+  const outlienedShields = Array(5 - filledShields.length)
```
dergigi, line 43: I just noticed that we can't always assume a mix depth of 5, since this can be configured by the user:

> The value of mixdepth runs from 0..M-1, where M is the number of mixdepths chosen by the user; by default 5.

See the [JoinMarket docs](https://github.com/JoinMarket-Org/JoinMarket-Docs/blob/master/High-level-design.md#wallets).

It's probably fine for now, but we might want to consider this at some point in the future.

dergigi (approved):
Nice! Looks really good! 💖 

Tested on Firefox in both desktop and mobile modes, with all settings variations. Code looks good to me, the only thing we should refactor at some point in the future is the hard-coded maximum mix depth.

src/components/CurrentWalletMagic.jsx
dnlggr: That is a very good point! It was an easy fix to make this dynamic so I figured I'd do it right away. We still use hard coded accounts in the Send/Receive views. Once we update those (in #57) we can make it dynamic there as well.

PR #52, opened 2022-02-04
fix: hidden balance icon alignment and size

dnlggr:
Properly centers the "hidden" icon and size it relative to the current active font size.

This is important for when balances are shown in larger font such as in the "privacy level screen" in #51.


dergigi (approved):
Looks good to me ✅

PR #62, opened 2022-02-07
feat: update create wallet flow

dnlggr:
Updates the "create wallet" flow according to the designs in [Figma](https://www.figma.com/file/kfejZJFlwBywvLEnPEmJo1/JoinMarket-UI?node-id=2850%3A70409).

It's just a UI refresh more or less. The only difference in business logic is that the wallet doesn't get opened until the seed and password have been confirmed by the user.

One thing I wanted to add is that the seed and password are hidden by default and the user needs to click to unveil them. Similar to how we hide the balance by default. Thoughts on that?

## 📸 

Screenshot 2022-02-07 at 17 43 03
Screenshot 2022-02-07 at 17 43 12


dergigi:
> One thing I wanted to add is that the seed and password are hidden by default and the user needs to click to unveil them. Similar to how we hide the balance by default. Thoughts on that?

We should make it considerably difficult to show the seed and show various warnings, since it is a fatal action, potentially. [Phoenix](https://phoenix.acinq.co/) does this well, for example, we could do it in a similar way.

src/components/CreateWallet.jsx
```diff
@@ -52,57 +21,25 @@ export default function CreateWallet({ currentWallet, startWallet }) {
...
+  return (
+    <>
       <rb.Form onSubmit={onSubmit} validated={validated} noValidate>
-        <h1>Create Wallet</h1>
-        {alert && <rb.Alert variant={alert.variant}>{alert.message}</rb.Alert>}
-        <rb.Form.Group className="mb-3" controlId="walletName">
+        <rb.Form.Group className="mb-4" controlId="walletName">
           <rb.Form.Label>Wallet Name</rb.Form.Label>
-          <rb.Form.Control name="wallet" style={{ maxWidth: '20em' }} required />
+          <rb.Form.Control name="wallet" placeholder="Your wallet..." required />
```
theborakompanioni, line 29:
```suggestion
          <rb.Form.Control name="wallet" placeholder="Your wallet..." disabled={isCreating} required />
```

It takes a few seconds till a wallet is created. A user should not be able to add more to the input while the wallet is in creation. However, I do not know if this is a user-friendly way to solve this. What do you think?

src/components/CreateWallet.jsx
```diff
@@ -52,57 +21,25 @@ export default function CreateWallet({ currentWallet, startWallet }) {
...
+          <rb.Form.Control.Feedback>Looks good!</rb.Form.Control.Feedback>
           <rb.Form.Control.Feedback type="invalid">Please set a wallet name.</rb.Form.Control.Feedback>
         </rb.Form.Group>
-        <rb.Form.Group className="mb-3" controlId="password">
+        <rb.Form.Group className="mb-4" controlId="password">
           <rb.Form.Label>Password</rb.Form.Label>
           <rb.Form.Control
             name="password"
             type="password"
+            placeholder="Choose a secure password..."
```
theborakompanioni, line 38:
```suggestion
            placeholder="Choose a secure password..."
            disabled={isCreating}
```

Same as above (especially important for a password input, I guess).

src/components/CreateWallet.jsx
```diff
@@ -116,10 +53,136 @@ export default function CreateWallet({ currentWallet, startWallet }) {
...
+      <p className="mb-4">
+        <rb.Form.Switch onChange={onSwitch} label="I've written down the information above." />
+      </p>
+      <rb.Button
+        variant="dark"
+        type="submit"
+        disabled={!userConfirmed}
+        onClick={() => userConfirmed && walletConfirmed()}
+      >
+        Fund Wallet
```
theborakompanioni, line 90:
```suggestion
        Fund wallet
```
In Figma (if I am on the correct version) "wallet" is not capitalized.
I do not think it is relevant.. it was the only real difference I saw to an otherwise pixel-perfect implementation :muscle:

theborakompanioni:
> > One thing I wanted to add is that the seed and password are hidden by default and the user needs to click to unveil them. Similar to how we hide the balance by default. Thoughts on that?
> Just to be clear, I think it's fine for now as is, the above could be a future improvement.

I agree with that.

Tested on Firefox/Brave :heavy_check_mark:  mobile/desktop :heavy_check_mark:  darkmode/lightmode :heavy_check_mark: . 
Looks superb yet again. Pixel-perfect.

src/components/CreateWallet.jsx
dnlggr: Sure that's a very good point. The request to create the wallet is already sent off at that point so it wouldn't change anything technically but definitely a good addition from a UX perspective imo! 👌

dnlggr:
> We should make it considerably difficult to show the seed and show various warnings, since it is a fatal action, potentially. [Phoenix](https://phoenix.acinq.co/) does this well, for example, we could do it in a similar way.

Totally agree. The one difference from us to Phoenix is that we only show the seed once immediately after wallet creation. Still, if someone sees that seed they can still later on steal you funds after you funded your wallet. I tracked https://github.com/joinmarket-webui/joinmarket-webui/issues/65 to add that. I'll try to get this in soon—I think we should already have that for the first release.

dnlggr:
Thanks for the reviews guys! I'll merge this later today unless there's still some feedback to be addressed?

dergigi:
> Totally agree. The one difference from us to Phoenix is that we only show the seed once immediately after wallet creation. Still, if someone sees that seed they can still steal you funds later on after you funded your wallet. 

That's right. What's usually done in that case is that the user is quizzed, to make sure that the seed was written down. We should definitely make sure to do this and get it right, because loss of funds can occur if the user fails to write it down.

It say it's not super-duper critical since the first release will be an early beta release anyway, but still.

> I think we should already have that for the first release.

I agree!

> Thanks for the reviews guys! I'll merge this later today

Go from my side ✅

PR #67, opened 2022-02-08
refactor: collect api requests in single file

### LINKED ISSUE #60: Unify API access
Currently, each API request is assembled and sent by every component itself. As a developer, it would be nice if this could happen clearly from a single service. This makes it easier to understand when, where and why requests are made.

See the [Open API spec for jmwalletd](https://raw.githubusercontent.com/JoinMarket-Org/joinmarket-clientserver/master/docs/api/wallet-rpc.yaml).

I already tried to generate a client from the specification with [openapi-generator](https://github.com/OpenAPITools/openapi-generator). Namely used generator [javascript](https://openapi-generator.tech/docs/generators/javascript),
[javascript-flowtyped](https://openapi-generator.tech/docs/generators/javascript-flowtyped), [typescript](https://openapi-generator.tech/docs/generators/typescript), [typescript-axios](https://openapi-generator.tech/docs/generators/typescript-axios), [typescript-fetch](https://openapi-generator.tech/docs/generators/typescript-fetch). But these look very verbose and bloated, the easiest way is probably just to move all API accesses in one file for now.

What do you think? Any benefits of generating the client at this stage of development? Or is a simple, single file better for now?

theborakompanioni:
This change really just moves all `fetch` calls to a single file `libs/JmWalletApi.js`.
It does not aim to be feature-complete, nor does it add any feature, increase configurability or something else.
So this can definitely be improved in the future, but is out of scope for now.

I'd need feedback on the method naming (e.g. just `walletAddressNew`) and params, as I did not include `signal` param in every call (because it has not been used in specific requests) and conversion between `name` and `walletname` feels also a bit cumbersome.

Closes #60

### DIFF
--- a/src/libs/JmWalletApi.js
+++ b/src/libs/JmWalletApi.js
@@ -61,8 +61,6 @@ const walletUnlock = async ({ walletname }, { password }) => {
 }
 
 const walletUtxos = async ({ walletname, token, signal }) => {
-  const wallettype = 'sw-fb'
-
   return await fetch(`/api/v1/wallet/${walletname}/utxos`, {
     headers: { ...Authorization(token) },
     signal,
--- a/src/libs/JmWalletApi.js
+++ b/src/libs/JmWalletApi.js
@@ -14,8 +14,8 @@ const getSession = async ({ signal }) => {
   return await fetch(`/api/v1/session`, { signal })
 }
 
-const getAddressNew = async ({ walletname, token, accountNr, signal }) => {
-  return await fetch(`/api/v1/wallet/${walletname}/address/new/${accountNr}`, {
+const getAddressNew = async ({ walletName, token, accountNr, signal }) => {
+  return await fetch(`/api/v1/wallet/${walletName}/address/new/${accountNr}`, {
     headers: { ...Authorization(token) },
     signal,
   })
@@ -27,21 +27,21 @@ const getWalletAll = async ({ signal }) => {
   })
 }
 
-const postWalletCreate = async ({ walletname, password }) => {
-  const wallettype = 'sw-fb'
+const postWalletCreate = async ({ walletName: name, password }) => {
+  const walletname = name.endsWith('.jmdat') ? name : `${name}.jmdat`
 
   return await fetch(`/api/v1/wallet/create`, {
     method: 'POST',
     body: JSON.stringify({
-      password,
+      wallettype: 'sw-fb',
       walletname,
-      wallettype,
+      password,
     }),
   })
 }
 
-const getWalletDisplay = async ({ walletname, token, signal }) => {
-  return await fetch(`/api/v1/wallet/${walletname}/display`, {
+const getWalletDisplay = async ({ walletName, token, signal }) => {
+  return await fetch(`/api/v1/wallet/${walletName}/display`, {
     headers: { ...Authorization(token) },
     signal,
   })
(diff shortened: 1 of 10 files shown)


src/libs/JmWalletApi.js
```diff
@@ -0,0 +1,129 @@
+/**
+ * Simple collection of all api requests to jmwalletd.
+ *
+ * It is not tried to be feature-complete, but to represent only what is really used.
```
dnlggr, line 4:
```suggestion
 * This is not aiming to be feature-complete.
```

src/libs/JmWalletApi.js
```diff
@@ -0,0 +1,129 @@
+/**
+ * Simple collection of all api requests to jmwalletd.
```
dnlggr, line 2:
```suggestion
 * Simple collection of api requests to jmwalletd.
```

dnlggr (commented):
Nice I was already thinking that we might soon need something like that. Perfect timing! 👏 👌 

> So this can definitely be improved in the future, but is out of scope for now.

I agree, all functions in a file is good enough for now. We can make this more configurable later.

> I'd need feedback on the method naming

I usually like names that immediately tell you exactly what is going on by looking at the call site. For REST/HTTP requests like we have here we could go with a combination of HTTP method + resource. E.g. `getSession`, `getAddressNew`, `postWalletCreate`, `getWalletDisplay`, `postWalletUnlock` etc. If we turn the API into more of a service later on we can still introduce more natural reading names but since this is just a collection of raw HTTP requests, I'd stay as "close to the metal" as possible for now.

> conversion between name and walletname feels also a bit cumbersome.

Yep. Luckily JS got a bit better in that regard. You can now have something similar to named parameters as other languages have them. I haven't tried this myself yet but checkout [this article](https://simonsmith.io/destructuring-objects-as-function-parameters-in-es6#renaming-the-extracted-values). I'd say let's pick a good descriptive name for the parameters at the call site and use whatever name we need for the API request only inside the function. I hope this works. 🤞 😄 

Let me know what you think! Great work!

theborakompanioni:
> For REST/HTTP requests like we have here we could go with a combination of HTTP method + resource. E.g. `getSession`, `getAddressNew`, `postWalletCreate`, `getWalletDisplay`, `postWalletUnlock` etc. [...] I'd stay as "close to the metal" as possible for now.

It is a bit confusing at first for two non-idempotent `GET` requests `getWalletLock` and `getMakerStop`, but it is definitely more descriptive raw request wise now - thanks!

I'll add a comment to these two methods, so at least the intention behind it is clear. :+1:

> 
> > conversion between name and walletname feels also a bit cumbersome.
> 
> Yep. Luckily JS got a bit better in that regard. You can now have something similar to named parameters as other languages have them. I haven't tried this myself yet but checkout [this article](https://simonsmith.io/destructuring-objects-as-function-parameters-in-es6#renaming-the-extracted-values). I'd say let's pick a good descriptive name for the parameters at the call site and use whatever name we need for the API request only inside the function. I hope this works. crossed_fingers smile

Thank you! JS definitely got quite some cool features lately. I'll try - please report if I end up making it more confusing : D

dnlggr:
> It is a bit confusing at first for two non-idempotent GET requests getWalletLock and getMakerStop, but it is definitely more descriptive "raw request wise" now - thanks!

Good point on those. Imo we can also drop the `get` if it doesn't read nicely and the function is still descriptive.

dnlggr (approved):
LGTM ✅ 

Eventually it would be nice to have the Joinmarket backend realated naming like `walletname` or `cj_fee` only in the request bodies or headers and default to JS style camelCase everywhere else. But that's a bit out of scope now and something we can tackle in the future if we want to.

PR #70, opened 2022-02-08
fix: Invalid DOM property 'class' in App

theborakompanioni:
Fixes a small warning  "Invalid DOM property `class`. Did you mean `className`?".

For anyone also wondering why a warning is given even when resulting DOM is exactly the same: https://www.geeksforgeeks.org/why-react-uses-classname-over-class-attribute/


secondl1ght (approved):

secondl1ght:
Good point, I know using class still works but I meant to use className instead. I'll go back through my other code as well and update everything to className. Thanks!

PR #73, opened 2022-02-09
feat: display basic yield generator report

### LINKED ISSUE #47: Display yield generator activity
Since https://github.com/JoinMarket-Org/joinmarket-clientserver/issues/1122 has been merged (https://github.com/JoinMarket-Org/joinmarket-clientserver/pull/1148), it is possible to show the yield generator activity to the user.

A simple view would already enhance the user experience by a lot.

theborakompanioni:
Display he yield generator report as simple table.

image

Needs a better representation but I suppose it is better than nothing. Hopefully someone with a sense for design can improve the visualization.

Closes #47.


theborakompanioni:
Should we display it in `useAdvancedWalletMode` only for now?

dnlggr:
> Should we display it on useAdvancedWalletMode only for now?

Or make it toggleable on the earn page. Something like a "show report" button maybe?

theborakompanioni:
> Should we display it on `useAdvancedWalletMode` only for now?

:heavy_check_mark: I don't want to make designers unhappy. I guess we get away by displaying it only in "advance" mode.

> Or make it toggleable on the earn page. Something like a "show report" button maybe?

image

"Display report" / "Show report"?

dnlggr:
> Or make it toggleable on the earn page. Something like a "show report" button maybe?

I'd use a button for that. A toggle (which is basically just a glorified checkbox) for showing hiding UI is not really common afaik.

theborakompanioni:
> I'd use a button for that. A toggle (which is basically just a glorified checkbox) for showing hiding UI is not really common afaik.

image
image

dergigi (commented):
Nice! Looks good to me 👍 One small suggestion to improve the error wording.

What's the easiest way to generate data for this report on regtest?

theborakompanioni:
> Nice! Looks good to me +1 One small suggestion to improve the error wording.

:+1:

> What's the easiest way to generate data for this report on regtest?

You can start the marker service from a wallet with funds (e.g. after `./init-setup.sh`): Then you will immediately see the "Connected" line.

To get actual data you'd do a coinjoin and then change the `target` in `./setupProxy.js` to route `/api/` to `https://localhost:29183` , which is the second instance that coinjoined with you. I think there is no live reload so you'd need to run `npm start` again.

PS: I want to use our `standalone` docker image as base for the secondary instance in the future, then we can have the most recent version on port `8081` and do better regression testing.

PR #76, opened 2022-02-09
feat: updated send page

dnlggr:
Updates the send page to look nicer and more close to Figma. We could improve the UI after sending or while joining  quite a bit but not for this first release.

## 📸 

Screenshot 2022-02-09 at 12 53 17

### DIFF
--- a/src/components/Send.jsx
+++ b/src/components/Send.jsx
@@ -5,6 +5,7 @@ import * as rb from 'react-bootstrap'
 import PageTitle from './PageTitle'
 import ToggleSwitch from './ToggleSwitch'
 import { useCurrentWalletInfo, useSetCurrentWalletInfo, useCurrentWallet } from '../context/WalletContext'
+import { useSettings } from '../context/SettingsContext'
 import { serialize } from '../utils'
 import * as Api from '../libs/JmWalletApi'
 
@@ -65,6 +66,7 @@ export default function Send({ currentWallet }) {
   const wallet = useCurrentWallet()
   const walletInfo = useCurrentWalletInfo()
   const setWalletInfo = useSetCurrentWalletInfo()
+  const settings = useSettings()
 
   const location = useLocation()
   const [validated, setValidated] = useState(false)
@@ -219,7 +221,7 @@ export default function Send({ currentWallet }) {
                     .sort((lhs, rhs) => lhs.account - rhs.account)
                     .map(({ account, account_balance: balance }) => (
                       <option key={account} value={account}>
-                        Account {account} ({balance})
+                        Account {account} {settings.showBalance && `(\u20BF${balance})`}
                       </option>
                     ))}
                 </rb.Form.Select>
--- a/src/components/Send.jsx
+++ b/src/components/Send.jsx
@@ -1,16 +1,99 @@
-import React from 'react'
+import React, { useEffect } from 'react'
 import { useState } from 'react'
 import { useLocation } from 'react-router-dom'
 import * as rb from 'react-bootstrap'
+import PageTitle from './PageTitle'
+import { useCurrentWalletInfo, useSetCurrentWalletInfo, useCurrentWallet } from '../context/WalletContext'
 import { serialize, ACCOUNTS } from '../utils'
 
+const CollaboratorsSelector = () => {
+  const [numCollaborators, setNumCollaborators] = useState(6) // Todo: Sane default
+  const [usesCustomNumCollaborators, setUsesCustomNumCollaborators] = useState(false)
+
+  return (
+    <rb.Form.Group controlId="collaborators">
+      <rb.Form.Label className="mb-0">Choose a number of collaborators {numCollaborators}</rb.Form.Label>
+      <div className="mb-2">
+        <rb.Form.Text className="text-secondary">
+          A higher number is better for privacy, but also increases the fee.
+        </rb.Form.Text>
+      </div>
+      <div className="d-flex flex-row flex-wrap" style={{ gap: '0.5rem' }}>
+        {[3, 5, 6, 9].map((number, index) => {
+          return (
+            <rb.Button
+              key={index}
+              variant="white"
+              className={`py-2 px-2 border rounded text-center${
+                !usesCustomNumCollaborators && numCollaborators === number && ' border-dark border-2'
+              }`}
+              style={{ width: '6rem' }}
+              onClick={() => {
+                setUsesCustomNumCollaborators(false)
+                setNumCollaborators(number)
+              }}
+            >
+              {number}
+            </rb.Button>
+          )
+        })}
+        <rb.Form.Control
+          name="collaborators"
+          type="number"
+          min={1}
+          placeholder="Other"
+          className={`py-2 px-2 border rounded text-center${usesCustomNumCollaborators && ' border-dark border-2'}`}
+          style={{ width: '6rem' }}
+          onChange={(e) => {
+            setUsesCustomNumCollaborators(true)
+            setNumCollaborators(e.target.value)
+          }}
+          onClick={(e) => {
+            setUsesCustomNumCollaborators(true)
+            setNumCollaborators(e.target.value)
+          }}
+        />
+      </div>
+      <rb.Form.Control.Feedback type="invalid">Please set the counterparties.</rb.Form.Control.Feedback>
+    </rb.Form.Group>
+  )
+}
+
 export default function Payment({ currentWallet }) {
+  const wallet = useCurrentWallet()
+  const walletInfo = useCurrentWalletInfo()
+  const setWalletInfo = useSetCurrentWalletInfo()
+
   const location = useLocation()
   const [validated, setValidated] = useState(false)
   const [alert, setAlert] = useState(null)
   const [isSending, setIsSending] = useState(false)
   const [isCoinjoin, setIsCoinjoin] = useState(false)
   const [account, setAccount] = useState(parseInt(location.state?.account, 10) || 0)
+
+  useEffect(() => {
+    if (walletInfo) {
+      return
+    }
+
+    const abortCtrl = new AbortController()
+    const { name, token } = wallet
+    const opts = {
+      headers: { Authorization: `Bearer ${token}` },
+      signal: abortCtrl.signal,
+    }
+
+    setAlert(null)
+
+    fetch(`/api/v1/wallet/${name}/display`, opts)
+      .then((res) => (res.ok ? res.json() : Promise.reject(new Error(res.message || 'Loading wallet failed.'))))
+      .then((data) => setWalletInfo(data.walletinfo))
+      .catch((err) => {
+        !abortCtrl.signal.aborted && setAlert({ variant: 'danger', message: err.message })
+      })
+
+    return () => abortCtrl.abort()
+  }, [wallet, setWalletInfo, walletInfo])
 
   const sendPayment = async (account, destination, amount_sats) => {
     const { name, token } = currentWallet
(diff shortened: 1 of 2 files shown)


src/components/Send.jsx
```diff
@@ -1,17 +1,103 @@
...
+        <rb.Form.Text className="text-secondary">
+          A higher number is better for privacy, but also increases the fee.
+        </rb.Form.Text>
+      </div>
+      <div className="d-flex flex-row flex-wrap">
+        {collaboratorsSelection.map((number, index) => {
+          return (
+            <rb.Button
+              key={index}
+              variant="white"
```
theborakompanioni, line 30:
```suggestion
              variant={settings.theme === 'light' ? 'white' : 'dark'}
```

Content of the buttons are not visible in dark mode. This fixes the visibility for now, but borders are still not quite right in dark mode.

src/components/Send.jsx
```diff
@@ -1,17 +1,103 @@
...
+                setNumCollaborators(number)
+              }}
+            >
+              {number}
+            </rb.Button>
+          )
+        })}
+        <rb.Form.Control
+          type="number"
+          min={1}
```
theborakompanioni, line 45:
```suggestion
          min={1}
          max={999}
```
I do not want to keep users from providing any number they want, but more than three digits distorts the interface. However, maybe even `99` as maximum is still sane.

dnlggr: Yeah let's go for 99. Good point!

dnlggr:
Damn I forgot to test for dark mode. My mistake. Will update now. 🤦‍♂️

src/components/Send.jsx
```diff
@@ -1,17 +1,103 @@
...
+          }}
+          onClick={(e) => {
+            if (e.target.value) {
+              setUsesCustomNumCollaborators(true)
+              setNumCollaborators(e.target.value)
+            }
+          }}
+        />
+      </div>
+      <rb.Form.Control.Feedback type="invalid">Please set the counterparties.</rb.Form.Control.Feedback>
```
theborakompanioni, line 61:
```suggestion
      {usesCustomNumCollaborators && <rb.Form.Control.Feedback type="invalid">Please set the counterparties.</rb.Form.Control.Feedback>}
      </div>
```
I have not managed to get this message to be displayed otherwise.

However:
- If you provide an invalid value (e.g. `-1`), the default value is shown as "selected", but the form cannot be sent without visual feedback

src/components/Send.jsx
```diff
@@ -1,17 +1,104 @@
...
-export default function Payment({ currentWallet }) {
+const CollaboratorsSelector = ({ numCollaborators, setNumCollaborators }) => {
+  const settings = useSettings()
+  const [usesCustomNumCollaborators, setUsesCustomNumCollaborators] = useState(false)
+
+  const collaboratorsSelection = [3, 5, 6, 9]
+
+  return (
+    <rb.Form.Group className="collaborators-selector">
+      <rb.Form.Label className="mb-0">Choose a number of collaborators ({numCollaborators})</rb.Form.Label>
```
dnlggr, line 20: Naming suggestions?

src/components/Send.jsx
```diff
@@ -1,17 +1,104 @@
...
+    }
+
+    setAlert(null)
+
+    fetch(`/api/v1/wallet/${name}/display`, opts)
+      .then((res) => (res.ok ? res.json() : Promise.reject(new Error(res.message || 'Loading wallet failed.'))))
+      .then((data) => setWalletInfo(data.walletinfo))
+      .catch((err) => {
+        !abortCtrl.signal.aborted && setAlert({ variant: 'danger', message: err.message })
+      })
```
theborakompanioni, line 98: Can now also use `Api.getWalletDisplay` if you want to.

src/components/Send.jsx
```diff
@@ -1,17 +1,104 @@
...
+            }
+          }}
+        />
+      </div>
+      <rb.Form.Control.Feedback type="invalid">Please set the counterparties.</rb.Form.Control.Feedback>
+    </rb.Form.Group>
+  )
+}
+
+export default function Send({ currentWallet }) {
```
theborakompanioni, line 66: How exactly does this `currentWallet` behave now compared to the `wallet` from `useCurrentWallet()`?

src/components/Send.jsx
dnlggr: Updated.

src/components/Send.jsx
```diff
@@ -92,65 +180,88 @@ export default function Payment({ currentWallet }) {
...
+                  defaultValue={account}
+                  onChange={(e) => setAccount(parseInt(e.target.value, 10))}
+                  required
+                  className="number"
+                >
+                  {walletInfo.accounts
+                    .sort((lhs, rhs) => lhs.account - rhs.account)
+                    .map(({ account, account_balance: balance }) => (
+                      <option key={account} value={account}>
+                        Account {account} {settings.showBalance && `(\u20BF${balance})`}
```
theborakompanioni, line 232: Balance is always shown as BTC.. which is kind of strange when I have my settings set to `sats` and the next input field wants me to insert a value in sats. 
What do you think about displaying "xxx sats" when `settings.unit` is `SATS`?

src/components/Send.jsx
dnlggr: Oh good catch, that's a remainder from the old version where the wallet was passed in as a prop. Removed.

src/components/Send.jsx
dnlggr: Nice! 🙌 Will do.

src/components/Send.jsx
theborakompanioni: :+1:

src/components/Send.jsx
dnlggr: Let's track that as an improvement if ok for you. We can't use the `<Balance>` component inside an `<option>` unfortunately so we'll probably need some trickery. 🪄

src/components/Send.jsx
dnlggr: Updated. ✅ 

> If you provide an invalid value (e.g. -1), the default value is shown as "selected", but the form cannot be sent without visual feedback

Can you elaborate on that. Could not reproduce it. I feel like I might be missing something.

src/components/Send.jsx
dnlggr: Tracked https://github.com/joinmarket-webui/joinmarket-webui/issues/78.

src/components/Send.jsx
theborakompanioni: > Updated. white_check_mark
:+1:
> Can you elaborate on that. Could not reproduce it. I feel like I might be missing something.

My wording was not clear - here are other steps to also reproduce it:
- insert a 'abc' in the "other" input -> feedback is shown now
- select a number e.g. 6 (but leave 'a'!) -> feedback disappears
- try sending the form -> form cannot be sent

dnlggr: Oh I get it now! Yeah you're right that whole part needs a little work still. Will update this tomorrow. 😄

src/components/Send.jsx
```diff
@@ -1,20 +1,111 @@
...
-export default function Payment({ currentWallet }) {
+const CollaboratorsSelector = ({ numCollaborators, setNumCollaborators }) => {
+  const settings = useSettings()
+  const [usesCustomNumCollaborators, setUsesCustomNumCollaborators] = useState(false)
+
+  const collaboratorsSelection = [3, 5, 6, 9]
+
+  return (
+    <rb.Form.Group className="collaborators-selector">
+      <rb.Form.Label className="mb-0">Choose a number of collaborators ({numCollaborators})</rb.Form.Label>
```
dergigi, line 20:
```suggestion
      <rb.Form.Label className="mb-0">Number of collaborators: {numCollaborators}</rb.Form.Label>
```

src/components/Send.jsx
```diff
@@ -92,65 +184,88 @@ export default function Payment({ currentWallet }) {
...
+              Loading
+            </div>
+          </rb.Col>
+        </rb.Row>
+      ) : (
+        <rb.Row className="send justify-content-center">
+          <rb.Col md={10} lg={8} xl={6}>
+            <PageTitle
+              title="Send bitcoin"
+              subtitle="You can mix your bitcoin before sending them for better privacy." // Todo: Copy.
```
dergigi, line 202: I'll do my best to come up with a better copy. I really don't like "mix" - note to self, use something along the lines of "Collaborative transactions increase the privacy of yourself and others."

dergigi (commented):
Beautiful! 😍 Looks great already. Superb job!

dergigi:
One thing I noticed is that the font rendering changes when you open the drop-down menu. Not critical at all, but I guess we'd have to use a custom drop-down (not the system UI one) to fix this.

Before expansion:
Screenshot 2022-02-09 at 22 02 08

After:
Screenshot 2022-02-09 at 22 02 02

src/components/Send.jsx
dnlggr: 💯 Agreed!

dnlggr:
> One thing I noticed is that the font rendering changes when you open the drop-down menu. Not critical at all, but I guess we'd have to use a custom drop-down (not the system UI one) to fix this.

Yeah good find! Unfortunately, `<select>`s are a bit limited in what they can do. Font customization might be possible though—I'll check tomorrow. If it's not easily possible to change the font, I'll add this to #78 which likely needs a custom dropdown as well.

src/components/Send.jsx
```diff
@@ -92,65 +184,88 @@ export default function Payment({ currentWallet }) {
...
+            />
+            {alert && <rb.Alert variant={alert.variant}>{alert.message}</rb.Alert>}
+            <rb.Form onSubmit={onSubmit} validated={validated} noValidate>
+              <rb.Form.Group className="mb-4" controlId="destination">
+                <rb.Form.Label>Recipient</rb.Form.Label>
+                <rb.Form.Control
+                  name="destination"
+                  placeholder="Enter address..."
+                  defaultValue=""
+                  className="number"
```
theborakompanioni, line 212: `className` of `number` for input "address" not sounding quite optimal here.
Although I like the looks of it! : -)

dnlggr: Fair point! :smile

src/components/Send.jsx
```diff
@@ -1,20 +1,162 @@
...
+
+const isValidAmount = (candidate) => {
+  const parsed = parseInt(candidate, 10)
+  return !isNaN(parsed) && parsed > 0
+}
+
+const isValidNumCollaborators = (candidate) => {
+  const parsed = parseInt(candidate, 10)
+  return !isNaN(parsed) && parsed >= 1 && parsed <= 99
+}
```
dnlggr, line 28: These aren't very strong validation rules. They're basically the same as before just extracted from the built-in form validation to give us a bit more fine grained control. We can update those in the future to e.g. only allow valid looking addresses etc.

dnlggr:
@theborakompanioni I updated the whole validation logic. Went into a bit of a rabbit hole. I hope I didn't make it worse. 😅 Can you please test that part again? 🙏 Let me know if something doesn't make sense.

src/components/Send.jsx
```diff
@@ -1,20 +1,168 @@
...
 
-export default function Payment({ currentWallet }) {
+// not cryptographically random
+const pseudoRandomNumber = (min, max) => {
+  return Math.round(Math.random() * (max - min)) + min
+}
+
+const isValidAddress = (candidate) => {
+  return typeof candidate === 'string' && !(candidate === '')
+}
```
dergigi, line 18: Just FYI, some parts of a lib I've worked on in the past might be of interest if we ever want to do real address validation. See [this](https://github.com/swan-bitcoin/xpub-tool/blob/0c2cd9f090665f8eb23c935fed28837488444e03/packages/xpub-lib/src/validation.js) and [that](https://github.com/swan-bitcoin/xpub-tool/blob/0c2cd9f090665f8eb23c935fed28837488444e03/packages/xpub-lib/src/validation.test.js#L125).

dergigi (commented):
Looks good to me, we can deal with any smaller issues in a separate PR imho. I'd say we can merge this as soon as @theborakompanioni tested it on his end.

Address validation would be good to avoid potentially destructive actions, but I assume the API will err if we send ridiculous inputs to it?

theborakompanioni:
> Looks good to me, we can deal with any smaller issues in a separate PR imho. I'd say we can merge this as soon as @theborakompanioni tested it on his end.

Yeah, looks really great. Smaller issues in a separate PR is fine with me of course :+1:
The only thing I found is kind of a blocker though: If a user hits the "send" button, a success message is shown and the form is reset. But clicking the "send" button a second time will send a request with the same information again. The user is unaware of this as the form does not contain any information anymore. Worse yet, the button can be clicked multiple times..

This is really the biggest issue that came to light. The current version in `master` does not have this behavior.

> Address validation would be good to avoid potentially destructive actions, but I assume the API will err if we send ridiculous inputs to it?

Keeping someone from entering _obvious_ invalid values is good. 
Yes, the server has some sanity checks here: [jmclient/taker_utils.py#L55-L63](https://github.com/JoinMarket-Org/joinmarket-clientserver/blob/50f5228498c670258740cd3cab485bcb39e177a3/jmclient/jmclient/taker_utils.py#L55-L63) (this is a "direct send")

dnlggr:
> If a user hits the "send" button, a success message is shown and the form is reset. But clicking the "send" button a second time will send a request with the same information again. The user is unaware of this as the form does not contain any information anymore. Worse yet, the button can be clicked multiple times..

Oh I know what this is. Will fix ASAP. Thanks so much for finding! Why does this damn screen have to be so annoying. 😅

dnlggr:
Thanks @theborakompanioni for finding this, should be fixed now!

theborakompanioni (approved):
:rocket:

PR #77, opened 2022-02-09
feat: hide sensitive info on wallet create

### LINKED ISSUE #65: Hide seed and password by default on wallet creation confirmation screen
Maybe also do a seed quiz. See discussion in https://github.com/joinmarket-webui/joinmarket-webui/pull/62.

dnlggr:
Resolves #65.

I've made it so that the user needs to reveal the sensitive info at least once for the "Fund wallet" button to become active.

## 📸 

Screenshot 2022-02-09 at 14 42 42
Screenshot 2022-02-09 at 14 42 44

### DIFF
--- a/src/components/CreateWallet.jsx
+++ b/src/components/CreateWallet.jsx
@@ -62,32 +62,43 @@ const WalletCreationForm = ({ createWallet, isCreating }) => {
 
 const WalletCreationConfirmation = ({ createdWallet, walletConfirmed }) => {
   const [userConfirmed, setUserConfirmed] = useState(false)
-
-  const onToggle = (isToggled) => {
-    setUserConfirmed(isToggled)
-  }
+  const [revealSensitiveInfo, setRevealSensitiveInfo] = useState(false)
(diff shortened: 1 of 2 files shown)


src/components/CreateWallet.jsx
```diff
@@ -96,14 +107,14 @@ const WalletCreationConfirmation = ({ createdWallet, walletConfirmed }) => {
   )
 }
 
-const Seedphrase = ({ seedphrase }) => {
+const Seedphrase = ({ seedphrase, isBlurred }) => {
```
theborakompanioni, line 110:
```suggestion
const Seedphrase = ({ seedphrase, isBlurred = true }) => {
```
Not necessary, but it would be a nice pro private default.

theborakompanioni (approved):
lgtm :heavy_check_mark: 

Important enhancement! Nice.

src/components/CreateWallet.jsx
dnlggr: Let's do it.

Issue #69, opened 2022-02-08
Add shields to other views


secondl1ght:
We talked about in the call today adding the shields to wherever the Accounts are referenced to create a visual link between the accounts and the privacy levels. I just wanted to check that this is what was decided before implementing anything.

dergigi:
Yes, I think showing the shields wherever we sensibly can would be a good idea. I also liked that a better name emerged organically, we seem to be calling them "privacy levels" which works quite well imho.

secondl1ght:
Sounds good, so we can add the shields and also rename them from Accounts to Privacy Level 0 etc.

I will put in a PR for the renaming - for the shields it looks like the only other place the accounts are referenced is in the Send and Receive dropdown selections. 

I looked at the code and it may be hard to add the shields here because these dropdowns are populated by mapping an array of numbers. Does anyone have an idea on how to get them into these dropdowns through refactoring?

dergigi:
Given that we want to [get rid of the shields](https://github.com/joinmarket-webui/joinmarket-webui/wiki/Representation-of-Mixdepths) in the long run, I guess we can close this issue?

PR #79, opened 2022-02-09
Change accounts to privacy levels

secondl1ght:
This was suggested in issue #69, I have changed the UI references to Accounts to Privacy Levels. I think this looks really good and drives home the meaning to the users better.

image
image

Please let me know if I have missed any other references to the Account. Thanks!


dnlggr (commented):
Nice, looks great! 💯 

I'd say let's do "Privacy Levels" in the normal mode but keep "Accounts" in advanced mode for now (since there are tons of references to accounts in this mode and "advanced" users can probably handle it). You can switch on `settings.useAdvancedWalletMode` to decide on which one to show.

PR #82, opened 2022-02-10
feat: update earn page

theborakompanioni:
Also fixes 1 of 2 points in https://github.com/joinmarket-webui/joinmarket-webui/issues/87.

### DIFF
--- a/src/components/Earn.jsx
+++ b/src/components/Earn.jsx
@@ -201,17 +201,24 @@ export default function Earn({ currentWallet, makerRunning }) {
         <rb.Form onSubmit={onSubmit} validated={validated} noValidate>
           {!makerRunning && !isWaiting && (
             <>
-              <rb.Form.Group className="mb-3" controlId="offertype">
-                <rb.Form.Check
-                  type="switch"
-                  label="Relative offer"
-                  checked={isRelOffer}
-                  onChange={(e) => setAndPersistOffertype(e.target.checked ? OFFERTYPE_REL : OFFERTYPE_ABS)}
-                />
-              </rb.Form.Group>
+              {(settings.useAdvancedWalletMode || !isRelOffer) && (
+                <rb.Form.Group className="mb-3" controlId="offertype">
+                  <rb.Form.Check
+                    type="switch"
+                    label="Relative offer"
+                    checked={isRelOffer}
+                    onChange={(e) => setAndPersistOffertype(e.target.checked ? OFFERTYPE_REL : OFFERTYPE_ABS)}
+                  />
+                </rb.Form.Group>
+              )}
               {isRelOffer ? (
                 <rb.Form.Group className="mb-3" controlId="feeRel">
-                  <rb.Form.Label>Relative Fee (percent)</rb.Form.Label>
+                  <rb.Form.Label className="mb-0">Relative Fee (percent)</rb.Form.Label>
+                  <div className="mb-2">
+                    <rb.Form.Text className="text-secondary">
+                      As a percentage of the amounts you help others mix.
+                    </rb.Form.Text>
+                  </div>
                   <rb.Form.Control
                     type="number"
                     name="feeRel"
@@ -239,19 +246,21 @@ export default function Earn({ currentWallet, makerRunning }) {
                   <rb.Form.Control.Feedback type="invalid">Please provide an absolute fee.</rb.Form.Control.Feedback>
                 </rb.Form.Group>
               )}
-              <rb.Form.Group className="mb-3" controlId="minsize">
-                <rb.Form.Label>Minimum amount in SATS</rb.Form.Label>
-                <rb.Form.Control
-                  type="number"
-                  name="minsize"
-                  required
-                  step={1000}
-                  value={minsize}
-                  min={0}
-                  onChange={(e) => setAndPersistMinsize(e.target.value)}
-                />
-                <rb.Form.Control.Feedback type="invalid">Please provide a minimum amount.</rb.Form.Control.Feedback>
-              </rb.Form.Group>
+              {settings.useAdvancedWalletMode && (
+                <rb.Form.Group className="mb-3" controlId="minsize">
+                  <rb.Form.Label>Minimum amount in SATS</rb.Form.Label>
+                  <rb.Form.Control
+                    type="number"
+                    name="minsize"
+                    required
+                    step={1000}
+                    value={minsize}
+                    min={0}
+                    onChange={(e) => setAndPersistMinsize(e.target.value)}
+                  />
+                  <rb.Form.Control.Feedback type="invalid">Please provide a minimum amount.</rb.Form.Control.Feedback>
+                </rb.Form.Group>
+              )}
             </>
           )}
           <rb.Button variant="dark" type="submit" disabled={isSending || isWaiting}>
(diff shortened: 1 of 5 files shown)


theborakompanioni:
Question: Should the relative value be display as real percentage? e.g. if user inputs 0.001, then show `Relative fee (0.1%)` as input label. Or should we generally input as percentage? Is the value in Figma 3% (factor input) or 0.03% (percentage input)?

src/components/Earn.jsx
```diff
@@ -189,109 +197,138 @@ export default function Earn({ currentWallet, makerRunning }) {
...
+            {!makerRunning && !isWaiting && (
+              <>
+                {settings.useAdvancedWalletMode && (
+                  <rb.Form.Group className="mb-3" controlId="offertype">
+                    <rb.Form.Check
+                      type="switch"
+                      label="Relative offer"
+                      checked={isRelOffer}
+                      onChange={(e) => setAndPersistOffertype(e.target.checked ? OFFERTYPE_REL : OFFERTYPE_ABS)}
+                    />
```
dnlggr, line 218: 🎛 Let's use our custom `ToggleSwitch` here what do you think?

src/components/Earn.jsx
```diff
@@ -189,109 +197,138 @@ export default function Earn({ currentWallet, makerRunning }) {
...
+                      max={0.1}
+                      step={0.0001}
+                      required
+                      onChange={(e) => setAndPersistFeeRel(e.target.value)}
+                    />
+                    <rb.Form.Control.Feedback type="invalid">Please provide a relative fee.</rb.Form.Control.Feedback>
+                  </rb.Form.Group>
+                ) : (
+                  <rb.Form.Group className="mb-3" controlId="feeAbs">
+                    <rb.Form.Label>Absolute fee in sats</rb.Form.Label>
```
dnlggr, line 244: Should we add a subtitle here as well to match that one of the relative offer?

src/components/Send.jsx
```diff
@@ -304,7 +304,7 @@ export default function Send() {
                   name="amount"
                   type="number"
                   value={amount || ''}
-                  className="number"
+                  className="slashed-zeroes"
```
dnlggr, line 307: Perfect!

src/components/Earn.jsx
```diff
@@ -189,109 +197,138 @@ export default function Earn({ currentWallet, makerRunning }) {
...
-                aria-hidden="true"
-                className="ms-2 me-1"
-              />
-            )}
-          </rb.Button>
-          {isShowReport && <YieldgenReport lines={yieldgenReportLines} />}
-        </div>
+        <rb.Row className="justify-content-center mt-5 mb-3">
+          <rb.Col md={10} lg={8} xl={6}>
+            <h6>Report</h6>
```
dnlggr, line 302:
```suggestion
```

I'd leave this out. It's clear by the button anyway.

dnlggr (commented):
Nice, good job! Looks great! ✨  See my comments inline. Most of them are NITs.

src/components/Earn.jsx
```diff
@@ -189,109 +197,135 @@ export default function Earn({ currentWallet, makerRunning }) {
...
+                      type="number"
+                      name="feeRel"
+                      required
+                      step={0.0001}
+                      value={feeRel}
+                      min={0}
+                      max={0.1}
+                      onChange={(e) => setAndPersistFeeRel(e.target.value)}
+                    />
+                    <rb.Form.Control.Feedback type="invalid">Please provide a relative fee.</rb.Form.Control.Feedback>
```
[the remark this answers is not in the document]
dnlggr: Just saw you comment. I'd go with percentage. In the future we could even use a cool slider or something like that. 🎚️

src/components/Earn.jsx
```diff
@@ -189,109 +197,135 @@ export default function Earn({ currentWallet, makerRunning }) {
...
+                      onChange={(e) => setAndPersistOffertype(e.target.checked ? OFFERTYPE_REL : OFFERTYPE_ABS)}
+                    />
+                  </rb.Form.Group>
+                )}
+                {isRelOffer ? (
+                  <rb.Form.Group className="mb-3" controlId="feeRel">
+                    <rb.Form.Label className="mb-0">Relative fee</rb.Form.Label>
+                    <div className="mb-2">
+                      <rb.Form.Text className="text-secondary">
+                        As a percentage of the amounts you help others with improved privacy.
```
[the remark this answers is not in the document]
theborakompanioni: Thought of something like that too. Definitely worth keeping in mind.

src/components/Earn.jsx
```diff
@@ -189,109 +197,135 @@ export default function Earn({ currentWallet, makerRunning }) {
...
-                  onChange={(e) => setAndPersistFeeAbs(e.target.value)}
-                />
-                <rb.Form.Control.Feedback type="invalid">Please provide an absolute fee.</rb.Form.Control.Feedback>
-              </rb.Form.Group>
+    <div className="earn">
+      <rb.Row className="justify-content-center">
+        <rb.Col md={10} lg={8} xl={6}>
+          <PageTitle
+            title="Earn bitcoin"
+            subtitle="By making your bitcoin available for others, you help them improve their privacy and you can also earn a yield."
```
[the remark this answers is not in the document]
theborakompanioni: Done. Guess we should address phrases and wording in general before the next release. Not resolving the conversation so @dergigi has a chance to comment.

src/components/Earn.jsx
```diff
@@ -179,7 +213,8 @@ export default function Earn({ currentWallet, makerRunning }) {
 
     if (isValid) {
       if (makerRunning === false) {
-        await startMakerService(feeAbs, feeRel, offertype, minsize)
+        const feeRelFactor = percentageToFactor(feeRel)
```
dnlggr, line 216: Probably doesn't make a difference but I feel like it would be cleaner to do that conversion before setting the state. Basically keep the state a factor and keep the UI-related things on the UI layer. Meaning converting percent -> factor before setting the state and factor -> percent after reading the state before it is being rendered. Only if it's not a huge hassle though. What do you think—tell me if it doesn' t make sense.

src/components/Earn.jsx
```diff
@@ -34,35 +39,59 @@ const YieldgenReport = ({ lines }) => {
...
 
+  const minRelFee = 0
+  const maxRelFee = 10.0
+
+  const percentageToFactor = (val, precision = 6) => {
+    // Value cannot just be divided
+    // e.g. ✗ 0.0027 / 100 == 0.000027000000000000002
+    // but: ✓ Number((0.0027 / 100).toFixed(6)) = 0.000027
+    return Number((val / 100).toFixed(precision))
+  }
```
dnlggr, line 93: Reminds me that we still have no unit tests. 😅

theborakompanioni: Yeah, I guess we should get this going sooner than later.
Let's get v0.0.3 out and then introduce basic unit tests next?

src/components/Earn.jsx
theborakompanioni: Way more comprehensible! Will do.

dnlggr (approved):
Nice, LGTM ✅

PR #84, opened 2022-02-11
fix: extract error message of text/html response

### LINKED ISSUE #83: Improve error message server responds with `text/html`
Improve error message when server returns `500 Internal Server Error` as `text/html` (e.g. when trying to send with too little funds ) - we display "JSON.parse: unexpected character at line 1 column 1 of the JSON data" as we try to parse json.

theborakompanioni:
Fixes #83

This is only done on the "send" page currently. Have not experienced this behaviour on other requests than `/direct-send`.

Before:

After:


dnlggr (approved):
LGTM! ✅

PR #86, opened 2022-02-12
fix: display info text on send page if service is running

theborakompanioni:
Do you have any ideas on the best way to phrase?
It should be informative and let the user know what to do.

Currently it is..
- ..when the maker is running: 
  - ~~🛈 Sending as collaborative transaction is currently disabled: Earn is active. Stop the service in order to send as collaborative transaction.~~
  - "Earn is active. Stop the service in order to send collaborative transactions."

- ..when a taker is running:
  - ~~🛈 Sending as collaborative transaction is currently disabled: A transaction is already in progress. Please wait until this operation is finished. This will only take a few minutes.~~
  - ~~A collaborative transaction is currently in progress.~~
  - "A collaborative transaction is in progress."

The info takes a while to appear - this will be resolved once the necessary data is available sooner (e.g. by reintegrating the websocket connection or requesting session information more timely).

Fixes 1 of 2 points in https://github.com/joinmarket-webui/joinmarket-webui/issues/87.

### DIFF
--- a/src/components/Send.jsx
+++ b/src/components/Send.jsx
@@ -302,7 +302,7 @@ export default function Send({ makerRunning, coinjoinInProcess }) {
                 🛈{' '}
                 <small className="text-secondary">
                   {makerRunning && <>Earn is active. Stop the service in order to send collaborative transactions.</>}
-                  {coinjoinInProcess && <>A collaborative transaction is in progress.</>}
+                  {coinjoinInProcess && <>A collaborative transaction is currently in progress.</>}
                 </small>
               </div>
             </rb.Fade>


src/components/Send.jsx
```diff
@@ -286,7 +296,19 @@ export default function Send() {
               title="Send bitcoin"
               subtitle="Collaborative transactions increase the privacy of yourself and others."
             />
+
+            <rb.Fade in={!isCoinjoinOptionEnabled} mountOnEnter={true} unmountOnExit={true}>
+              <div className="mb-4 border border-1 p-3" style={{ borderRadius: '.25rem' }}>
+                🛈{` `}
```
dnlggr, line 302: Same comments as in #82. 😉

src/components/Send.jsx
```diff
@@ -286,7 +296,19 @@ export default function Send() {
...
+                <small className="text-secondary">
+                  {makerRunning && <>Earn is active. Stop the service in order to send collaborative transactions.</>}
+                  {coinjoinInProcess && <>A collaborative transaction is in progress.</>}
```
dnlggr, line 305:
```suggestion
                  {coinjoinInProcess && <>A collaborative transaction is currently in progress.</>}
```

src/components/Send.jsx
```diff
@@ -339,12 +361,14 @@ export default function Send() {
...
               </rb.Form.Group>
-              <rb.Form.Group controlId="isCoinjoin" className={`${isCoinjoin ? 'mb-3' : ''}`}>
-                <ToggleSwitch
-                  label="Send as collaborative transaction for improved privacy"
-                  onToggle={(isToggled) => setIsCoinjoin(isToggled)}
-                />
-              </rb.Form.Group>
+              {isCoinjoinOptionEnabled && (
+                <rb.Form.Group controlId="isCoinjoin" className={`${isCoinjoin ? 'mb-3' : ''}`}>
+                  <ToggleSwitch
```
dnlggr, line 366: That's nice!

dnlggr (commented):
Nice!

src/components/Send.jsx
theborakompanioni: In this case `npm run format` will insert `{' '}`.

src/components/Send.jsx
theborakompanioni: Okay, had that in the previous version (see the phrasing history, ha!) but thought it sounds kind of strange when the user _just_ initiated the transaction. Will add "currently" (also matches the wording on the Earn page)!

dnlggr (approved):

PR #89, opened 2022-02-14
fix: navigate to root on unmapped path

### LINKED ISSUE #81: Empty page is rendered for `/wallet` when wallet is locked
When the wallet is locked, all paths except `/` will render an empty page.

This is true for paths that usually lead to something, such as `/wallet`, `/send`, `/earn`, `/receive`, `/settings`, but also for non-existent paths such as `/this-is-a-path-that-leads-nowhere`.

We should either:
* Show a 404 page, or
* Redirect the user to the wallets page, i.e. `/`

Screenshot:
Screenshot 2022-02-09 at 21-52-13 JoinMarket

theborakompanioni:
Navigates to "/" on unmapped paths, e.g. `/unmapped`, `/this-is-a-path-that-leads-nowhere`, but also on `/send`, `/earn`, etc if no wallet is active.

Closes #81

### DIFF
--- a/src/components/App.jsx
+++ b/src/components/App.jsx
@@ -1,5 +1,5 @@
 import { useState, useEffect, useCallback } from 'react'
-import { Route, Routes } from 'react-router-dom'
+import { Route, Routes, Navigate } from 'react-router-dom'
 import * as rb from 'react-bootstrap'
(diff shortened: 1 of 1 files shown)


dnlggr (approved):
Good addition! ✅

PR #92, opened 2022-02-14
feat: add alpha disclaimer

### LINKED ISSUE #90: More prominent "Alpha/Beta" disclaimer
We currently have an "this is alpha software, be prepared to loose everything" disclaimer—but it's hidden somewhere in the onboarding reel so most users probably won't read it.

We should have a more dominant disclaimer that's visible in the app (at least on the send page but I would even show it on all pages). The disclaimer should make it clear that this is not "top-notch privacy for your bitcoin" (yet). We need to be careful not to make false promises in these early stages where we offer only limited features (e.g. no tumbler, only simple spend-joins) Especially, since we're targeting a less technical audience we need to be careful not to make them believe that a simple spend with 2 collaborators makes their coins perfectly private for example. See also the [recent discussions in our group chat](https://t.me/JoinMarketWebUI/986).

Adding this to the 0.0.3 release I think it's important to discuss before we release the first (alpha/beta) version.

dnlggr:
Resolves #90.

Text is just something off the top of my head. Let me know if you have a better idea please!

Semi happy with the looks but probably good enough.

## 📸 
Screenshot 2022-02-14 at 18 24 11
Screenshot 2022-02-14 at 18 24 16
Screenshot 2022-02-14 at 18 29 38

### DIFF
--- a/src/components/App.jsx
+++ b/src/components/App.jsx
@@ -25,6 +25,7 @@ export default function App() {
   const [makerRunning, setMakerRunning] = useState()
   const [connectionError, setConnectionError] = useState()
   const [coinjoinInProcess, setCoinjoinInProcess] = useState()
+  const [showAlphaWarning, setShowAlphaWarning] = useState(false)
   const settings = useSettings()
 
   const startWallet = useCallback(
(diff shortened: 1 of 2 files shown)


theborakompanioni (approved):
lgtm :heavy_check_mark:

PR #93, opened 2022-02-15
fix: condition check during wallet confirmation

theborakompanioni:
Noticed this during testing the docker setup. 
The checked `name` property is actually from the function `createWallet` rather than the reference to variable `createdWallet`.
Seems to invoke an error in the packaged application but does not show up during development.


dnlggr (approved):
Wow good find.

PR #97, opened 2022-02-15
fix: footer spacing and links

dnlggr:
This should've gone into #92 but I forgot to push the commit. 🤦‍♂️


theborakompanioni (approved):
:heavy_check_mark:

PR #98, opened 2022-02-15
fix: fix and dry up font featues

### LINKED ISSUE #96: Parts of an address might be rendered as math
Just noticed a small issue: if an address contains an "x" between two numbers it will be rendered as a mathematical symbol:

Screenshot 2022-02-15 at 11 18 21

An easy fix would be to render every address inside a `<code>` tag or something, like this: `bcrt1q6wyvqnm2daajashl0jdl8l84ghwadsg4g37x30`

dnlggr:
Resolves #96.

Explicitly turns off the [contextual alternates font feature](https://rsms.me/inter/#features/calt) and dries up the "slashed-zeroes" CSS.

## 📸 

Screenshot 2022-02-15 at 11 49 25


src/components/Send.jsx
```diff
@@ -299,15 +299,18 @@ export default function Send({ makerRunning, coinjoinInProcess }) {
 
             <rb.Fade in={!isCoinjoinOptionEnabled} mountOnEnter={true} unmountOnExit={true}>
               <div className="mb-4 p-3 border border-1 rounded">
-                🛈{' '}
```
theborakompanioni, line 302: Remove on [Earn](https://github.com/joinmarket-webui/joinmarket-webui/blob/master/src/components/Earn.jsx#L244) page as well?

theborakompanioni (approved):
One small nit.

Otherwise LGTM :heavy_check_mark:

src/components/Send.jsx
dnlggr: Removed. What was that even?

theborakompanioni: It's an info sign. [Unicode Character 'CIRCLED INFORMATION SOURCE' (U+1F6C8)](https://www.fileformat.info/info/unicode/char/1f6c8/index.htm)

PR #99, opened 2022-02-15
dev: use joinmarket-webui-dev-standalone in regtest setup

theborakompanioni:
Changes to regtest environment:
- Use `joinmarket-webui/joinmarket-webui-dev-standalone` for second jm container
  - This will expose a UI on port `29080` - with basic auth enabled (user and password `joinmarket`). helpful for regression tests
- Use `--docker-install` flag for `latest` setup for first jm container
  - Deliberately not using `standalone` as base because everyone should be able to run any server version quickly
- Remove image based on `btcpayserver/joinmarket:v0.9.3`


theborakompanioni:
Hmm.. somehow I got the `jmwalletd` daemon to exit (seen this behaviour in previous version, but thought it got resolved):
```
jm_regtest_joinmarket2 | 2022-02-17 01:40:54,944 INFO exited: jmwalletd (exit status 0; expected)
```
Trying to reproduce it. I am investigating what exactly led to this and if it has been introduced by any changes on our side.

docker/regtest/dockerfile-deps/joinmarket/webui-standalone/Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM ghcr.io/joinmarket-webui/joinmarket-webui-dev-standalone:master@sha256:8e4beed972c9bf41fa6607bfad576ef75c6e73e47e978c96580bf8fd28751bbb
```
dnlggr, line 1: Does it make sense to wait with merging this and use a tag from `joinmarket-webui-dev-standalone`?

theborakompanioni: I did not think about tagging `dev-standalone` at all. I even thought it might be good to remove the digest, so we can always trigger a release on demand and get all devs to get a new version automatically (if they rebuild).
Maybe we should settle on a strategy here. I have not come up with a proper release strategy for the docker images like you have for this repo. What approach would you desire most?

dnlggr: Good questions. I think you're right it might make sense to just use whatever is the latest pushed release here... Theat would mean if we run [this](https://github.com/joinmarket-webui/joinmarket-webui-docker/blob/master/.github/workflows/create-and-publish-docker-dev-manually.yml) GitHub workflow, and I rebuild locally, it would be on the latest release right?

theborakompanioni: > That would mean if we run [this](https://github.com/joinmarket-webui/joinmarket-webui-docker/blob/master/.github/workflows/create-and-publish-docker-dev-manually.yml) GitHub workflow, and I rebuild locally, it would be on the latest release right?

Yes, this is correct. Currently there are no tags used and we can change anything related to docker to our needs on demand. I was thinking about tagging the docker images with vYYYYMMDD with a version table in the readme (I think @dergigi  mentioned something like this a while ago). But for the `dev` releases we can do something completely different, of course. I will remove the image checksum, if that is okay.

theborakompanioni: I will verify that this works as expected (getting the new version on rebuild when checksum is removed) and report here.

theborakompanioni: > I will verify that this works as expected (getting the new version on rebuild when checksum is removed) and report here.

Need to run `docker-compose build --pull --no-cache` to pull the latest `dev-standalone` - adapted the readme accordingly.

dnlggr (approved):
Looks good to me! ✅  Are we good to merge?

PR #104, opened 2022-02-16
fix: enable fetching manifest with credentials

theborakompanioni:
With this change the `site.webmanifest` file can be fetched on a server with an auth mechanism enabled.

From [developer.mozilla.org](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/crossorigin#example_webmanifest_with_credentials):
> The use-credentials value must be used when fetching a [manifest](https://developer.mozilla.org/en-US/docs/Web/Manifest) that requires credentials, even if the file is from the same origin.

For more on why this is needed on `rel="manifest"` see [here](https://github.com/w3c/manifest/issues/535#issuecomment-435739223).


dergigi (approved):
Looks good to me ✅

PR #108, opened 2022-02-17
feat: update wallets page

theborakompanioni:
More consistence in regards to other pages.

- Use one row and col for the whole page

- Alert not using full width anymore

- Slightly increased "Create new wallet" button when no wallet exists.

- The `PageTitle` is not centered anymore, which deviates from the Figma screen.
Personally, I think it looks better to have the headlines all in the same place (e.g. switching to the "Create Wallet" screen and back seemed a little awkward before) What do you think?

Before/After without wallets:
image
image

Before/After with wallets:
image
image

### DIFF
--- a/src/components/App.jsx
+++ b/src/components/App.jsx
@@ -119,11 +119,7 @@ export default function App() {
           <rb.Alert variant="danger">No connection to backend: {connectionError}.</rb.Alert>
         ) : (
           <Routes>
-            <Route
-              exact
-              path="/"
-              element={<Wallets currentWallet={currentWallet} startWallet={startWallet} stopWallet={stopWallet} />}
-            />
+            <Route exact path="/" element={<Wallets startWallet={startWallet} stopWallet={stopWallet} />} />
             <Route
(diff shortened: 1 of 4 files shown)


dnlggr (commented):
For the empty state I'd definitely keep the heading centered! It only looks good left aligned if the rest of the page has a left-ish feel to it (cf. the send page where there's lots of left aligned text in the form vs. the main wallets page where there whole page is centered and thus the heading is as well). For the empty state, the rest of the text and the button is centered so the heading looks better if its centered as well.

If there's multiple wallets I can life with it either way in that case. I personally like it centered but definitely not such a clear case here. Looks good either way.

The alert not being full width is definitely a good change! 👌

theborakompanioni:
> If there's multiple wallets I can life with it either way in that case. I personally like it centered but definitely not such a clear case here. Looks good either way.

Thank for the feedback. Changed it back to centered heading. Hope it is okay to just try things out and ask for your opinion. Centered is perfectly fine with me. :+1:

PR #110, opened 2022-02-17
feat: freeze/unfreeze utxos and fix style on dark theme

theborakompanioni:
I didn't want to cheat a feature in before the next version. I was just fixing the display in dark mode... and I thought... why not?

I'm happy to just fix the visual defect if you want. However, this would at least provide a proper workaround for users just trying things or testing to recover when they mistakenly sent to an address twice.

Before/After UTXO:
image
image

Before/After Accounts:


dnlggr (approved):
✅ ☀️ 🌑 🥶

PR #116, opened 2022-02-21
Take minimum_makers config var value into account on Send page

### LINKED ISSUE #101: Take `minimum_makers` config var value into account on Send page
When initiating a collaborative transaction the number of collaborators can be provided.
It seems, based on [these comments](https://github.com/getumbrel/umbrel/pull/1216#issuecomment-1030928317), this value must not be less than what is specified in config variable `minimum_makers`. Default is `4`.

A viable solution is to `GET` the value of this variable and use it as min value for the form on the Send page.

theborakompanioni:
Closes #101 

This PR fetches the `minimum_makers` config variable and uses it as minimum amount of collaborators on the Send page.
This means a taker operation cannot be started with a lower value than what is specified in the config file.

One little caveat is that the initial value for the amount collaborators is randomized between `[5, 7]` (min +1, min +3), which can lead to the initial value not being shown as active button in the `CollaboratorsSelector` component. 

Is this acceptable for now? What do you think?

<br />
<br />

Screenshot for `minimum_makers := 2`:

<br />

Also, the approach of parsing an error message from `text/html` is removed as this has been fixed in [joinmarket-clientserver#1170](https://github.com/JoinMarket-Org/joinmarket-clientserver/issues/1170)

Edit: this also does not take into account if the value is greater than `99` - which is the default maximum amount. Guess that is an unreasonable value and can therefore be ignored.

### DIFF
--- a/src/components/Send.jsx
+++ b/src/components/Send.jsx
@@ -27,9 +27,9 @@ const isValidAmount = (candidate) => {
   return !isNaN(parsed) && parsed > 0
 }
 
-const isValidNumCollaborators = (candidate) => {
+const isValidNumCollaborators = (candidate, minNumCollaborators) => {
   const parsed = parseInt(candidate, 10)
-  return !isNaN(parsed) && parsed >= 1 && parsed <= 99
+  return !isNaN(parsed) && parsed >= minNumCollaborators && parsed <= 99
 }
 
 const extractErrorMessage = async (response, fallbackReason = 'Unknown Error - No exact reasons are available : (') => {
@@ -53,20 +53,20 @@ const extractErrorMessage = async (response, fallbackReason = 'Unknown Error - N
   return message || fallbackReason
 }
 
-const CollaboratorsSelector = ({ numCollaborators, setNumCollaborators }) => {
+const CollaboratorsSelector = ({ numCollaborators, setNumCollaborators, minNumCollaborators }) => {
   const settings = useSettings()
 
   const [usesCustomNumCollaborators, setUsesCustomNumCollaborators] = useState(false)
 
   const validateAndSetCustomNumCollaborators = (candidate) => {
-    if (isValidNumCollaborators(candidate)) {
+    if (isValidNumCollaborators(candidate, minNumCollaborators)) {
       setNumCollaborators(candidate)
     } else {
       setNumCollaborators(null)
     }
   }
 
-  const defaultCollaboratorsSelection = [3, 5, 6, 7, 9]
+  const defaultCollaboratorsSelection = [0, 2, 3, 4, 6].map((val) => val + minNumCollaborators)
 
   return (
     <rb.Form noValidate className="collaborators-selector">
@@ -101,9 +101,9 @@ const CollaboratorsSelector = ({ numCollaborators, setNumCollaborators }) => {
           })}
           <rb.Form.Control
             type="number"
-            min={1}
+            min={minNumCollaborators}
             max={99}
-            isInvalid={!isValidNumCollaborators(numCollaborators)}
+            isInvalid={!isValidNumCollaborators(numCollaborators, minNumCollaborators)}
             placeholder="Other"
             defaultValue=""
             className={`p-2 border border-1 rounded text-center${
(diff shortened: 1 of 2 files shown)


theborakompanioni:
This is what the `CollaboratorsSelector` component looks like with the default value `4` for `minimum_makers`:
image

src/libs/JmWalletApi.js
```diff
@@ -155,6 +155,23 @@ const postFreeze = async ({ walletName, token }, { utxo, freeze = true }) => {
   })
 }
 
+/**
+ * Get the value of a specific config setting. Note values are always returned as string.
```
dnlggr, line 159:
```suggestion
 * Get the value of a specific config setting. Note that values are always returned as string.
```

src/libs/JmWalletApi.js
```diff
@@ -155,6 +155,23 @@ const postFreeze = async ({ walletName, token }, { utxo, freeze = true }) => {
...
+ *
+ * @returns an object with property `configvalue` as string
+ */
+const postConfigGet = async ({ walletName, token, signal }, { section, field }) => {
```
dnlggr, line 163: Why did they make this a POST call. 😅 😭 Really stretches our naming scheme...

Should we start making exceptions and just go for `getConfig` or so?

src/components/Send.jsx
```diff
@@ -139,22 +118,24 @@ export default function Send({ makerRunning, coinjoinInProcess }) {
...
   const [isCoinjoinOptionEnabled, setIsCoinjoinOptionEnabled] = useState(!makerRunning && !coinjoinInProcess)
+  const [minNumCollaborators, setMinNumCollaborators] = useState(4) // default value from an unchanged joinmarket.cfg (last check 2022-02-20)
 
   const initialDestination = null
   const initialAccount = 0
   const initialAmount = null
-  const initialNumCollaborators = () => {
-    return pseudoRandomNumber(5, 7)
+  const initialNumCollaborators = (minValue) => {
+    return minValue + pseudoRandomNumber(1, 3)
```
dnlggr, line 131: The way we're shipping the UI, the default value will remain 4 right? In the regtest setup it's 1 (naturally) so the default could be as low as 2. That feels super weird (but doesn't matter in the regtest setup of course). Just wanted to double check that we won't ship a version that recommends a low value such as 2 to the user.

dnlggr (commented):
Nice change! Yeah I'd say we can live with the 99 being hardcoded for now. 👍

src/libs/JmWalletApi.js
theborakompanioni: As you please. I do not have a strong opinion until we generate the client from the openapi spec : D
But if we deviate from the current naming scheme, we should reconsider all other method names too.

Should we just ditch the http method prefix and reorder?
e.g. `postMakerStart` -> `startMaker`
`postDirectSend` -> `sendDirect`
?

I'd do that in a separate PR if that is okay with you.
(Let's not touch files that have nothing to do with the task at hand)

src/components/Send.jsx
theborakompanioni: The default value is whatever the server repo chooses. We cannot control what a user will subsequently change the value to. In the future we might also let the user change the value in the UI, so we should expect it to have any arbitrary value.

But to answer your question: Yes, we do not change the value in our `standalone` image and hence it will take the current default, which is `4`. But of course the user can always overwrite it by providing a `jm_minimum_makers` env variable.

src/libs/JmWalletApi.js
dnlggr: Let's forget it for now and do a refactor of the API stuff another time. Then we can also improve other things like drying up response parsing etc.

src/components/Send.jsx
theborakompanioni: Great idea! Thank you!

Updated! Is this what you had in mind?

dnlggr (approved):
Cool! ✅

src/components/Send.jsx
dnlggr: Yeah I think that's a pretty nice solution now. 👍

PR #122, opened 2022-02-23
feat: readd websocket connection

### LINKED ISSUE #46: Fix websocket connection
Bring back the non-functional websocket connection, which got removed in joinmarket-webui/joinmarket-webui#45.

When I converted the prototype I couldn't figure out whether it is a problem in the web server/proxy or in the API's web socket which makes it fail. However, making this web socket connection work would be great, so that we can receive event updates via the API.

theborakompanioni:
This PR reintroduces the websocket connection to the backend. Closes #46.

The backend listens on port 28283 for incoming websocket connections.
See the [jmwalletd websocket documentation](https://github.com/JoinMarket-Org/joinmarket-clientserver/blob/v0.9.5/docs/JSON-RPC-API-using-jmwalletd.md#websocket).

A reverse proxy must pass all incoming connections with path `/jmws` to the backend server.
`/jmws` was chosen, as `/ws` is used by [`WebpackDevServer` by default](https://github.com/facebook/create-react-app/blob/v5.0.0/packages/react-dev-utils/webpackHotDevClient.js#L66). We could change that, but having it on a different path has some benefits.

The websocket connection, as it is currently implemented, is responsible for two things:
- Updating the maker/taker state
- Updating the connection indicator

#### Updating the maker/taker state
Before this PR the maker/taker state has only been updated by polling the `/session` api endpoint.
As the endpoint is polled only every 10 seconds (at the time of writing), this resulted in a user not being informed for some time whether an action had actually been performed. Now, the user is informed in (near) real-time.

#### Updating the connection indicator
Once the websocket connection is lost or closed, the connection indicator is updated. Also resulting in (near) real-time information to the user.

### Approach
As a websocket connection can be established without any authentication, it will be created when the application is loaded.
Once a wallet has been created or unlocked, the token is sent to the server, enabling receiving sensitive information.
When a wallet is closed or locked, the websocket connection is  kept open and reused once another wallet is unlocked.

If the connection is closed, an attempt is made to reopen the connection after a short delay. This delay is currently set to 1 second.

TODO:
- [x] Establish websocket connection
- [x] Update taker/maker state based on event data
- [x] Handle disconnects
- [x] Connection indicator based on websocket state in addition to state of request polling
- [x] Documentation (code, PR, etc.)
- [x] Update docker setup (see https://github.com/joinmarket-webui/joinmarket-webui-docker/pull/21)

Manual tests:
- [x] Reconnect after connection loss
- [x] Connect after initial failing connection
- [x] Successful connection with the (currently used) basic auth setup
- [x] Messages received correctly when switching wallets

#### Known Issues
- Shortly before unloading the window (e.g on a page refresh), the "No connection to backend" message is briefly displayed. This is expected as the websocket connection is closed, but it is rather unattractive. Guess it can be lived with (?).
- A "normal" transaction is received multiple times via websocket (consistently 4 times during testing), a collaborative transaction is received multiple times every 5 seconds (everytime the backend calls bitcoin core's `listtransactions` rpc method). We don't use this information or display it to the user anywhere, so it might not that big of a problem for now.

### DIFF
--- a/docker/regtest/docker-compose.yml
+++ b/docker/regtest/docker-compose.yml
@@ -124,10 +124,12 @@ services:
       - 8080  # payjoin server
       - 62601 # obwatch
       - 27183 # joinmarketd
-      - 28183 # jmwalletd
+      - 28183 # jmwalletd api
+      - 28283 # jmwalletd websocket
     ports:
       - "62601:62601"
       - "28183:28183"
+      - "28283:28283"
     volumes:
       - "joinmarket_datadir:/root/.joinmarket"
       - "nbxplorer_datadir:/root/.nbxplorer"
--- a/docker/regtest/docker-compose.yml
+++ b/docker/regtest/docker-compose.yml
@@ -176,9 +176,11 @@ services:
       - 8080  # payjoin server
       - 62601 # obwatch
       - 27183 # joinmarketd
-      - 28183 # jmwalletd
+      - 28183 # jmwalletd api
+      - 28283 # jmwalletd websocket
     ports:
       - "29183:28183"
+      - "29283:28283"
       - "29080:80"
     volumes:
       - "joinmarket2_datadir:/root/.joinmarket"
--- a/src/components/App.jsx
+++ b/src/components/App.jsx
@@ -33,6 +33,7 @@ export default function App() {
 
   const [makerRunning, setMakerRunning] = useState()
   const [connectionError, setConnectionError] = useState()
+  const [websocketConnected, setWebsocketConnected] = useState()
   const [coinjoinInProcess, setCoinjoinInProcess] = useState()
   const [showAlphaWarning, setShowAlphaWarning] = useState(false)
   const settings = useSettings()
@@ -75,11 +76,7 @@ export default function App() {
   // update the connection indicator based on the websocket connection state
   useEffect(() => {
     const websocketError = websocketState !== WebSocket.CONNECTING && websocketState !== WebSocket.OPEN
-    if (!websocketError) {
-      setConnectionError(null)
-    } else {
-      setConnectionError('Websocket is disconnected.')
-    }
+    setWebsocketConnected(!websocketError)
   }, [websocketState])
 
   useEffect(() => {
@@ -110,9 +107,6 @@ export default function App() {
         })
         .catch((err) => {
           if (!abortCtrl.signal.aborted) {
-            // set the connection error message from the http request as it
-            // might contain more useful information for the user than just
-            // "Websocket is disconnected", e.g. "Gateway Timeout"
             setConnectionError(err.message)
             resetState()
           }
@@ -267,8 +261,8 @@ export default function App() {
             </rb.Nav.Item>
           </div>
           <div className="d-flex order-0 order-md-2 flex-1 justify-content-center justify-content-md-end align-items-center">
-            <span className={`mx-1 ${connectionError ? 'text-danger' : 'text-success'}`}>•</span>
-            <span className="text-secondary">{connectionError ? 'Disconnected' : 'Connected'}</span>
+            <span className={`mx-1 ${websocketConnected ? 'text-success' : 'text-danger'}`}>•</span>
+            <span className="text-secondary">{websocketConnected ? 'Connected' : 'Disconnected'}</span>
           </div>
         </rb.Container>
       </rb.Nav>
--- a/src/components/App.jsx
+++ b/src/components/App.jsx
@@ -11,12 +11,14 @@ import CurrentWalletAdvanced from './CurrentWalletAdvanced'
 import Settings from './Settings'
 import Navbar from './Navbar'
 import { useSettings } from '../context/SettingsContext'
-import { useWebsocket } from '../context/WebsocketContext'
+import { useWebsocket, useWebsocketState } from '../context/WebsocketContext'
 import { useCurrentWallet, useSetCurrentWallet, useSetCurrentWalletInfo } from '../context/WalletContext'
 import { getSession, setSession, clearSession } from '../session'
 import * as Api from '../libs/JmWalletApi'
-
 import Onboarding from './Onboarding'
+
+// interval in milliseconds for periodic session requests
+const SESSION_REQUEST_INTERVAL = 10_000
 
 export default function App() {
   const currentWallet = useCurrentWallet()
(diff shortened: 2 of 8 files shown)


docker/regtest/docker-compose.yml
```diff
@@ -124,10 +124,12 @@ services:
       - 8080  # payjoin server
       - 62601 # obwatch
       - 27183 # joinmarketd
-      - 28183 # jmwalletd
+      - 28183 # jmwalletd api
+      - 28283 # jmwalletd websocket
```
dennisreimann, line 128: This doesn't look right :)

docker/regtest/docker-compose.yml
```diff
@@ -124,10 +124,12 @@ services:
...
     ports:
       - "62601:62601"
       - "28183:28183"
+      - "28283:28283"
```
dennisreimann, line 132: This addition neither.

dennisreimann (commented):
I think the port numers are messed up right now

docker/regtest/docker-compose.yml
theborakompanioni: Hey @dennisreimann .. Thanks for looking over it (still a draft!)

Can you elaborate?

From the docs:
> [...] which with defaults will start serving the RPC over https:// on port 28183, and a (secure) websocket server (wss://) on port 28283.

https://github.com/JoinMarket-Org/joinmarket-clientserver/blob/v0.9.5/docs/JSON-RPC-API-using-jmwalletd.md#introduction---how-to-start-the-server

theborakompanioni: > Yikes, it was just me being blind. Sorry for bothering!

You do not bother me. Quite the opposite. It's good to know you are watching over it.
I'd be happy if you were a reviewer : D
Will ping you once it is ready!

theborakompanioni:
PR is ready for review now. Will update the docker setup to handle the `/jmws` path accordingly, if you feel this approach is okay.

dennisreimann (approved):
tACK. Great work, thanks for bringing it back :)

One minor idea as I think we should not have the debug messages in production - otherwise this is great!

Regarding the "known issues": I think this is fine for now and we can address them in smaller, separate PRs if needed.

theborakompanioni:
During tests with the docker image the websocket gets terminated reproducibly with `Connection Closed: 1006` (and reopened subsequently - but causing a visual effect for the user). Please refrain from merging for now. I fear this needs further investigation : |

Will report back.

src/setupProxy.js
```diff
@@ -2,8 +2,7 @@ const { createProxyMiddleware } = require('http-proxy-middleware')
 
 module.exports = (app) => {
   app.use(
-    '/api/',
-    createProxyMiddleware({
```
dnlggr, line 6: Does this change anything or is it just style? (Either is fine, just wondering.)

theborakompanioni:
> During tests with the docker image the websocket gets terminated reproducibly with `Connection Closed: 1006`

This was due to the default nginx inactivity timeouts of 60s.

> Will report back.

Made some adaptions based on the findings.
Most notably:
- set only the connection indicator (in the footer) based on the websocket connection state
- send a keep alive message every 30 seconds if possible (the token - server will close the connection on any other message)
- simple linear back-off delays for retries when no connection can be made (to a maximum delay of 10 seconds)

Most of the issues I have experienced seem to be fixed (but I will keep testing!).
However, these are rather significant changes, hence another thorough review would be needed : |

Any suggestions are highly appreciated.

See the related Draft PR for docker: https://github.com/joinmarket-webui/joinmarket-webui-docker/pull/21

src/context/WebsocketContext.jsx
```diff
@@ -0,0 +1,159 @@
...
+    const onStateChange = () => setWebsocketState(websocket.readyState)
+
+    websocket.addEventListener('open', onStateChange)
+    websocket.addEventListener('close', onStateChange)
+
+    return () => {
+      websocket && websocket.removeEventListener('close', onStateChange)
+      websocket && websocket.removeEventListener('open', onStateChange)
+    }
+  }, [websocket])
```
dnlggr, line 80: Is `websocketState` ever different from using `websocket.readyState` directly?

src/components/App.jsx
```diff
@@ -43,6 +54,31 @@ export default function App() {
...
+
+    websocket.addEventListener('message', onWebsocketMessage)
+
+    return () => websocket && websocket.removeEventListener('message', onWebsocketMessage)
+  }, [websocket, onWebsocketMessage])
+
+  // update the connection indicator based on the websocket connection state
+  useEffect(() => {
+    const websocketError = websocketState !== WebSocket.CONNECTING && websocketState !== WebSocket.OPEN
+    setWebsocketConnected(!websocketError)
```
dnlggr, line 79: Technically we could also just do `setWebsocketConnected(websocketState === WebSocket.OPEN)` right? Since a connecting state isn't really connected. Just a nit though.

src/context/WebsocketContext.jsx
```diff
@@ -0,0 +1,159 @@
...
+
+    websocket.onclose = () => {
+      console.debug('websocket connection closed')
+    }
+
+    websocket.onmessage = (event) => {
+      const data = JSON.parse(event?.data)
+      console.debug('websocket message', data)
+    }
+  }
```
dnlggr, line 45: I'd be in favor of not dumping each message in the console (as it's really a lot of messages). One can also view the messages in the Network tap of browser dev tools (at least in Firefox, probably also in other browsers). That would keep the console clear.

dnlggr (commented):
Seems to works really nice! Just a couple of questions inline.

Regarding the flickering "dicsonnected" state when refreshing the page, I'd say that's fine.

src/context/WebsocketContext.jsx
theborakompanioni: Yes, small differences exist. `websocketState` will only once be `connecting`, subsequently only changes between `open` and `closed`. It would be great to have a way to react on state changes without registering event listeners in every component that wants to act on them. Do you know how that can be achieved?

src/setupProxy.js
theborakompanioni: `/api/` definition works fine - but `/jmws` did not work otherwise. Wanted to be consistent and changed the `/api/` route as well. Maybe there is something else to be done - I really just left it as it is as soon as it worked.

src/context/WebsocketContext.jsx
dnlggr: No not really. Let's keep it like this then! 👍

src/setupProxy.js
dnlggr: Was just wondering what the difference in the APIs was. But if it works it works! :)

dnlggr (approved):
LGTM! ✅ Nice to have instant feedback for those things now. 👌

PR #124, opened 2022-02-23
feat: allow for deployment in subdirectory

dennisreimann:
Right now the only way to deploy the web UI is with a separate host, it does not work in a subdirectory/path like `/joinmarket/`. I came across this trying to integrate the Web UI into the BTCPay Server setup. We want to avoid the complexity of setting up different hosts (and SSL certs) per service, that's why we deploy services in paths under the main domain.  

This PR determines the deployment base path depending on whether `PUBLIC_URL` is set. We aren't making use of the [`PUBLIC_URL` environment variable](https://create-react-app.dev/docs/adding-custom-environment-variables/#referencing-environment-variables-in-the-html), which is perfectly fine as it would require us to set it on build time and different integrations might want to use different paths anyways. 

I've tested this setting up a local proxy which pipes requests from `http://localhost:4000/joinmarket/` to `http://localhost:3000/` in order to check whether or not the asset URLs and API requests still work. From my tests the approach in `index.html` works even if `PUBLIC_URL` is set. It feels a bit hacky to get to the base path that way, but I did not find any other way and it looks like it works reliably. 

The determined path is then used to set the base for the API requests and act as the [`basename` for React Router](https://reactrouterdotcom.fly.dev/docs/en/v6/api#router). The asset paths work just fine because we've already set the [`homepage` property in package.json](https://create-react-app.dev/docs/deployment/#step-1-add-homepage-to-packagejson).

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -2,6 +2,8 @@
 <html lang="en">
   <head>
     <meta charset="utf-8" />
+    <link rel="base-url" href=".">
+    <link rel="public-url" href="%PUBLIC_URL%">
     <link rel="apple-touch-icon" sizes="180x180" href="%PUBLIC_URL%/apple-touch-icon.png">
     <link rel="icon" href="%PUBLIC_URL%/favicon.svg">
     <link rel="manifest" href="%PUBLIC_URL%/site.webmanifest" crossorigin="use-credentials">
@@ -12,6 +14,14 @@
     <title>JoinMarket</title>
     <script>
       window.JM = { SETTINGS_STORE_KEY: 'jm-settings', THEMES: ['light', 'dark'], THEME_ROOT_ATTR: 'data-theme' }
+      // determine base path depending on whether PUBLIC_URL is set
+      const baseUrl = document.querySelector('link[rel="base-url"]').href
+      const publicUrl = document.querySelector('link[rel="public-url"]').href
+      const base = baseUrl.length > publicUrl.length ? baseUrl : publicUrl
+      window.JM.PUBLIC_PATH = base
+        .replace(`${window.location.protocol}//${window.location.host}`, '') // remove domain part
+        .replace(/\/$/, '') // remove trailing slash
+      // theme
       const settings = JSON.parse(window.localStorage.getItem(JM.SETTINGS_STORE_KEY) || '{}')
       const userColorMode = settings.theme;
       const systemColorMode = window.matchMedia('(prefers-color-scheme: dark)').matches ? JM.THEMES[1] : JM.THEMES[0];
--- a/src/components/Sprite.jsx
+++ b/src/components/Sprite.jsx
@@ -3,7 +3,7 @@ import React from 'react'
 export default function Sprite({ symbol, className, ...props }) {
   return (
     <svg role="img" className={`d-inline-block sprite sprite-${symbol}${className ? ` ${className}` : ''}`} {...props}>
-      <use href={`${process.env.PUBLIC_URL}/sprite.svg#${symbol}`}></use>
+      <use href={`${window.JM.PUBLIC_PATH}/sprite.svg#${symbol}`}></use>
     </svg>
   )
 }
--- a/src/index.jsx
+++ b/src/index.jsx
@@ -10,7 +10,7 @@ import '@ibunker/bitcoin-react/dist/index.css'
 import './index.css'
 
 ReactDOM.render(
-  <BrowserRouter>
+  <BrowserRouter basename={window.JM.PUBLIC_PATH}>
     <SettingsProvider>
       <WalletProvider>
         <App />
--- a/src/libs/JmWalletApi.js
+++ b/src/libs/JmWalletApi.js
@@ -10,6 +10,7 @@
  * 'x-jm-authorization' so that the reverse proxy can apply its own
  * authentication mechanism.
  */
+const BASE_PATH = `${window.JM.PUBLIC_PATH}/api`
 
 /**
  * Construct a bearer authorization header object for the given token.
@@ -25,18 +26,18 @@ const Authorization = (token) => {
 }
 
 const getSession = async ({ signal }) => {
-  return await fetch(`/api/v1/session`, { signal })
+  return await fetch(`${BASE_PATH}/v1/session`, { signal })
 }
 
 const getAddressNew = async ({ walletName, token, accountNr, signal }) => {
-  return await fetch(`/api/v1/wallet/${walletName}/address/new/${accountNr}`, {
+  return await fetch(`${BASE_PATH}/v1/wallet/${walletName}/address/new/${accountNr}`, {
     headers: { ...Authorization(token) },
     signal,
   })
 }
 
 const getWalletAll = async ({ signal }) => {
-  return await fetch(`/api/v1/wallet/all`, {
+  return await fetch(`${BASE_PATH}/v1/wallet/all`, {
     signal,
   })
 }
@@ -44,7 +45,7 @@ const getWalletAll = async ({ signal }) => {
 const postWalletCreate = async ({ walletName: name, password }) => {
   const walletname = name.endsWith('.jmdat') ? name : `${name}.jmdat`
 
-  return await fetch(`/api/v1/wallet/create`, {
+  return await fetch(`${BASE_PATH}/v1/wallet/create`, {
     method: 'POST',
     body: JSON.stringify({
       wallettype: 'sw-fb',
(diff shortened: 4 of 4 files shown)


dnlggr:
I tried to test it with the setup you mentioned but I think I'm missing something.

I'm running a proxy (NGINX) with the following config:

```
server {
    listen 4000;

    location /joinmarket {
        proxy_pass http://127.0.0.1:3000;
    }
}
```

and I'm starting the web UI with:

```
PUBLIC_URL="http://localhost:4000/joinmarket" npm start
```

With this I get the web UI served on `:4000`, but the API requests don't go through. Again, I feel like I'm missing something on my end though. Pretty 🧠 -fried already. I'll try again tomorrow!

dennisreimann:
`PUBLIC_URL` is supposed to be a path, rather than an URL, yeah that naming makes sense. 

`PUBLIC_URL=/joinmarket npm start` it is.

dnlggr (approved):
I couldn't really reproduce it in a subdirectory like `/joinmarket`. Somehow the UI was served correctly but the`/api` requests were not working. Again I think I just misunderstand something. But it's definitely doesn't break anything if `PUBLIC_URL` is not set, so ✅ from me.

dennisreimann:
@dnlggr let's not merge it then, I'll post the nginx config so that we can reproduce and verify it works.

dennisreimann:
Added an extra nginx for testing the subdirectory deployment. Steps to reproduce:

For dev mode we need to set `PUBLIC_URL`, because somehow it does not honor the `homepage: "."` setting and does not use the `.` as a base in dev mode. Hence we cheat and set `PUBLIC_URL=/joinmarket`, which is accepted (note that a plain `/` also does not work). That's also why the ngixn setup is `/joinmarket/extrapath` …

The alternative to jumping through the extra `/joinmarket` path hoop would have been to do a production build and set up the proxy in the nginx — maybe this is something we want to include as a reference?

Ok, with all that out of the way:

1. `docker-compose up`
2. `PUBLIC_URL=/joinmarket npm start` 
3. Visit `http://localhost:3000/joinmarket/` (dev server) and see it works
4. Visit `http://localhost:8000/joinmarket/extrapath/` (nginx) and see it works
5. Restart `npm start` without `PUBLIC_URL` 
6. Visit `http://localhost:3000/joinmarket/` (dev server) and see it still works as before

src/setupProxy.js
```diff
@@ -1,27 +1,33 @@
...
   app.use(
-    '/ws/',
+    `${PUBLIC_URL}/ws/`,
     createProxyMiddleware({
       target: 'wss://localhost:28183',
       changeOrigin: true,
       secure: false,
       ws: true,
+      onProxyReq: (proxyReq, req, res) => {
+        proxyReq.path = proxyReq.path.replace(PUBLIC_URL, '')
```
theborakompanioni, line 29: Would using `pathRewrite` for this be more natural?
A [custom function](https://github.com/chimurai/http-proxy-middleware/blob/master/recipes/pathRewrite.md#custom-rewrite-function) can be defined which we can be reused for every existing and future route.

What do you think?

dennisreimann: Yes, I think it can be cleaned up – for now this is just what I came up with to quickly test it. I'm not sure if we'd want to keep the contents of my last commit, as it is  intended for testing purposes only.

theborakompanioni:
This somehow does not work on my current dev machine. I think it is a cgroup issue as it seems to be a problem with docker's file permissions : / I look into it more deeply in the coming days.

Thank you for providing the links. Seems to be a small rabbit hole of its own..
But I have a question, though.

> The alternative to jumping through the extra `/joinmarket` path hoop would have been to do a production build and set up the proxy in the nginx — maybe this is something we want to include as a reference?

Does this mean we have to know the path the app is deployed on at build time?

dennisreimann:
> Does this mean we have to know the path the app is deployed on at build time?

Nope, that's what I referred to in the initial description: We aren't using `PUBLIC_URL` in production, it is only used here for testing as the dev mode behaves differently than the production build in case `PUBLIC_URL` isn't set: `npm run build` produced an index.html that contains a '.' in that case (this is a result of the `"homepage": "."` setting, as intended), whereas `npm start` does not set this and produces absolute paths, which is a bug I think.

src/setupProxy.js
dennisreimann: I can rebase and clean it up once the websocket PR is merged or simply remove the last commit once you had a chance to verify it. Depends on whether or not we want to introduce the nginx into the regtest setup - it is only needed for testing this use case and possibly providing a reference for the integration, but we also got the docker repo for that.

theborakompanioni:
> Nope, that's what I referred to in the initial description: We aren't using `PUBLIC_URL` in production

Thanks for the clarification. :pray: 

I hope to fix my setup over the weekend and provide feedback next week, if that is okay.
Otherwise, if that's too late, we can merge it and fix any upcoming issues before the next release.

dennisreimann:
Nope, let's not rush this as it also feels a bit hacky and we should have verified and understood it at least.

dnlggr:
Ok testing according to the described test setup works almost for me now except that for the nginx hosted version on port 8000 the sprite gets loaded at the wrong path. It tries to load `http://localhost:8000/joinmarket/sprite.svg` (since `/joinmarket` is passed in as `PUBLIC_URL` I think) but it should be loading it at `http://localhost:8000/joinmarket/extrapath/sprite.svg`. I think that's just something off in the test setup though. Trying to reproduce it in a production setup now.

Let me summarize my understanding of the whole thing just to make sure I'm not completely off:

If we want host the whole thing under `https://btcpay.example.com/joinmarket`, then `document.querySelector('link[rel="base-url"]').href` would be equal to exactly that URL. After the replacing magic, `window.JM.PUBLIC_PATH` would be simply `/joinmarket`.
That means that `/joinmarket` would be prepended to all API request, used as basepath for the router etc.

And when we want to host at the root level (e.g. `https://example.com`), `window.JM.PUBLIC_PATH` would be the empty string `""` and nothing would change compared to right now.

dnlggr:
Ok I think I got it now. Sorry for the long feedback loop I don't know why I was so confused about this. 😅 

What I did to test was this:

1️⃣ Started the regtest setup to serve the `jmwalletd` backend on `https://127.0.0.1:28183`:

``` sh
docker-compose up
```

2️⃣ Built the UI for production:

``` sh
npm run build
```

3️⃣ Served the built UI with nginx under `http://localhost:7500`:

``` nginx
server {
    listen 7500;

    root /(...)/joinmarket-webui/build;
    index index.html;

    location /api/ {
        proxy_set_header Authorization $http_x_jm_authorization;
        proxy_set_header x-jm-authorization "";

        proxy_pass https://127.0.0.1:28183;
    }
}
```

4️⃣ Tested that everything works normally and the app is reachable under `http://localhost:7500`:

Screenshot 2022-02-25 at 14 28 05

Looks good! ✅ 

5️⃣ Started a second nginx that serves the UI on `http://localhost:7501/joinmarket` (via reverse proxy to the first server):

``` nginx
server {
    listen 7501;

    location /joinmarket/ {
        proxy_pass http://localhost:7500/;
    }
}
```

4️⃣ Tested that everything works normally and the app is reachable under `http://localhost:7501/joinmarket`:

Screenshot 2022-02-25 at 14 28 22

Looks good! ✅ 

---

Does that make sense?

dnlggr:
Sounds good! 🚀

dennisreimann:
> It tries to load http://localhost:8000/joinmarket/sprite.svg (since /joinmarket is passed in as PUBLIC_URL I think) but it should be loading it at http://localhost:8000/joinmarket/extrapath/sprite.svg. I think that's just something off in the test setup though

You are right -- fixed it in [e837dbc](https://github.com/joinmarket-webui/joinmarket-webui/pull/124/commits/e837dbc4150b2332a91fdf87cbe81accad8691b8). Thanks for all your feedback!

theborakompanioni (approved):
> Thanks for all your feedback!

utACK

Sorry, I did not get my things in order. Trusting the thorough review of @dnlggr and @dennisreimann's experience :100: :100: :100:

dnlggr:
Tested again, everything seems to work just fine. Merging this now! ⚾ 🛣

dnlggr:
Damn, there's something I missed, @theborakompanioni found it.

When you're on e..g the Send page (`/send`) and refresh, then `window.JM.PUBLIC_PATH` is going to be `/send`. That'll break API requests as they're trying to request `${window.JM.PUBLIC_PATH}/api/...`.

dnlggr:
I'll have a closer look after lunch.

dnlggr:
> I'll have a closer look after lunch.

Couldn't come up with a better solution that works dynamically without hardcoding a subdirectory. Do you have any ideas on how to do it so that when reloading `index.html` doesn't wrongly use the current route (e.g. `/send`) as subdirectory @dennisreimann?

PR #129, opened 2022-02-28
feat: allow for deployment in subdirectory

dennisreimann:
One more try at #124.

@dnlggr the [9a5904e](https://github.com/joinmarket-webui/joinmarket-webui/pull/129/commits/9a5904e51abe17bc754166f2f533042209df01b9) commit solves the issue mentioned in you last comment of the previous PR for me.

### DIFF
--- a/docker/regtest/docker-compose.yml
+++ b/docker/regtest/docker-compose.yml
@@ -2,6 +2,7 @@ version: "3"
 
 services:
   nginx:
+    container_name: jm_regtest_nginx_test_basepath
     image: nginx
     volumes:
       - ./dockerfile-deps/joinmarket/nginx:/etc/nginx/templates
--- a/docker/regtest/docker-compose.yml
+++ b/docker/regtest/docker-compose.yml
@@ -5,7 +5,7 @@ services:
     container_name: jm_regtest_nginx_test_basepath
     image: nginx
     volumes:
-      - ./dockerfile-deps/joinmarket/nginx:/etc/nginx/templates
+      - ./dockerfile-deps/nginx:/etc/nginx/templates
     ports:
       - "8000:80"
     extra_hosts:
--- a/docker/regtest/docker-compose.yml
+++ b/docker/regtest/docker-compose.yml
@@ -1,6 +1,14 @@
 version: "3"
 
 services:
+  nginx:
+    image: nginx
+    volumes:
+      - ./dockerfile-deps/joinmarket/nginx:/etc/nginx/templates
+    ports:
+      - "8000:80"
+    extra_hosts:
+      - "host.docker.internal:host-gateway"
 
   bitcoind:
     container_name: jm_regtest_bitcoind
--- a/docker/regtest/dockerfile-deps/joinmarket/nginx/default.conf.template
+++ b/docker/regtest/dockerfile-deps/joinmarket/nginx/default.conf.template
@@ -1,0 +1,18 @@
+# To test the reverse proxy setup, run the dev server using:
+#
+# PUBLIC_URL=/joinmarket npm start
+#
+# And visit http://localhost:8000/joinmarket/extrapath/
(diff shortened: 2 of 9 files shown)


dnlggr (commented):
Nice, I tested it again and it seems to work! 👏 

@theborakompanioni Can you please double check the websockets? I think we probably need to update the websocket endpoint as well right?

dennisreimann:
Oh yes, I'll need to rebase in case it works for all of us.

dennisreimann:
Rebased and added the nginx with testing instructions. I can remove the nginx commit, but now it is simple and clean enough that we could also keep it. Looking forward to your feedback and hopefully it works this time :)

docker/regtest/docker-compose.yml
```diff
@@ -1,6 +1,14 @@
 version: "3"
 
 services:
+  nginx:
+    image: nginx
```
theborakompanioni, line 5: Currently the container will be named `regtest_nginx_1` - all others start with `jm_regtest`. Should we add it here too and maybe provide a meaningful description like "test_basepath"?

```suggestion
    container_name: jm_regtest_nginx_test_basepath
    image: nginx
```

Just a nit. Not important.

docker/regtest/docker-compose.yml
```diff
@@ -1,6 +1,14 @@
...
+    volumes:
+      - ./dockerfile-deps/joinmarket/nginx:/etc/nginx/templates
```
theborakompanioni, line 7:
```suggestion
      - ./dockerfile-deps/nginx:/etc/nginx/templates
```
Can we move the nginx dir from `dockerfile-deps/joinmarket/nginx` directory one level up into `dockerfile-deps/nginx` ?

theborakompanioni (approved):
> Rebased and added the nginx with testing instructions. I can remove the nginx commit, but now it is simple and clean enough that we could also keep it. 
:+1:

> Looking forward to your feedback and hopefully it works this time :)

Again, utack for the docker stuff. I am sorry that I can only test the standard setup - which works great, and the reload issue seems fixed.

docker/regtest/docker-compose.yml
dennisreimann: Thanks, done in [d020b9b](https://github.com/joinmarket-webui/joinmarket-webui/pull/129/commits/d020b9b6f2085dae9bc723c64f6ea5530428c862).

dnlggr:
Tested using my previous testing method (see #124), the nginx setup from this PR and normal `npm start` dev mode. Everything seems to work as expected (Except for the websocket in the nginx setup, but that's an nginx setup issue right? So nothing wrong with the implementation.)

Looks good to me. I'd say we don't necessarily need the testing setup in master—I feel like once this works we won't need it again (and if we do, we can look up this PR). But I also don't mind if we merge it. Your choice. :)

dnlggr (approved):

dennisreimann:
> Except for the websocket in the nginx setup, but that's an nginx setup issue right?

The last commit was supposed to also make it work. What issue did you encounter?

dnlggr:
For the version on port 3000 hosted by `npm start` everything is fine, but for ngingx hosted versions, I'm getting:

```
WebSocket connection to 'ws://localhost:8000/joinmarket/jmws' failed: 
```

Is that working for you guys?

dennisreimann:
Yes, my guess is that you are visiting `localhost:8000/joinmarket`, but it needs to be `localhost:8000/joinmarket/extrapath/` (also note the trailing slash). Might that be the case?

image

image

dnlggr:
Ok I retested everything and it works now (including the websockets)! I must've either had not the latest commit checked out or missed the trailing `/`. In any case, it works now. Sorry for the confusion—that was my fault.

dennisreimann:
No worries, thanks to both of you for testing this and bearing with me on that feature :D

Issue #140, opened 2022-03-02
Remove privacy levels from receive screen in magic wallet mode


secondl1ght:
I saw on Telegram that a user was confused about the privacy levels being shown on the Receive screen and why the input was disabled. @theborakompanioni made a good point that this really should not be shown to the user because it is disabled anyway and you should always send to Account 0. I am submitting a PR to hide this from the user in magic wallet mode while still allowing advanced users to send to any account they choose.

secondl1ght:
Added with PR #141

PR #141, opened 2022-03-02
Remove privacy levels on receive screen in magic mode

secondl1ght:
As described in issue #140

image


theborakompanioni:
Is this okay with you @dnlggr ? Think we talked about it before and left it in. I'd be in favor of removing, simply because of the user-unfriendliness of disabled input elements.

dnlggr (approved):
LGTM ✅

PR #150, opened 2022-03-04
fix: update suggested number of collaborators

### LINKED ISSUE #134: Clarify default number of counterparties
The JM usage guide [mentions a default of 5-7](https://github.com/JoinMarket-Org/joinmarket-clientserver/blob/master/docs/USAGE.md#try-out-a-coinjoin-using-sendpaymentpy) counterparties which is what we implemented. Adam mentioned [on Telegram](https://t.me/JoinMarketWebUI/1220) that it should be 9.

Tracking this here so we don't forget that. I'd say let's err on the side of caution and go with the higher number here.

dnlggr:
Resolves #134.

Updates the suggested number of collaborators to be 8-10 or higher. This is the actual default value, not 5-7 like we had previously.

### DIFF
--- a/src/components/Send.jsx
+++ b/src/components/Send.jsx
@@ -11,7 +11,7 @@ import * as Api from '../libs/JmWalletApi'
 // initial value for `minimum_markers` from the default joinmarket.cfg (last check on 2022-02-20 of v0.9.5)
 const MINIMUM_MAKERS_DEFAULT_VAL = 4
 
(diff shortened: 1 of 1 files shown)


dennisreimann (approved):
ACK, +1 on the randomization!

theborakompanioni (approved):
tack! Thank you.
Any objections to merging?

PR #152, opened 2022-03-04
Warn on missing config vars

### LINKED ISSUE #135: Add warning about missing `max_cj_fee_rel`/`max_cj_fee_abs` settings
From @AdamISZ 
> i encountered a tricky error, but it's a function of me as user: i got 'error, config var not present' when starting the coinjoin. it took me a while to track down that it was because max_cj_fee_rel/abs was not present that caused it. users will have default `joinmarket.cfg`s at least to start with, so i hope they won't see that problem as they won't know how to fix it.

In the standalone versions of Jam the config variables `max_cj_fee_rel` and `max_cj_fee_abs` are populated (e.g. see [Umbrel](https://github.com/getumbrel/umbrel/blob/v0.4.15/apps/jam/docker-compose.yml#L29-L30) and [Citadel](https://github.com/runcitadel/apps/blob/f0743a968984174dcf92e8ec37a346b2ed0614ae/apps/jam/app.yml#L51-L52)). As these variables have no default values, users connecting to an existing `jmwalletd` instance will only ever see the message `config var not present` if they are missing. As a user I want a meaningful message about what is wrong or what I can do about it if manual adaptions are required.

Easiest workaround for that would be to always try to retrieve these values first.
It'd be also an option to set these variables according to user input, but these changes would not be persistent.
Assuming a user connects to an already running daemon, he or she is capable of adapting the config file as well.

theborakompanioni:
Closes #135.

This is really just a workaround for users connecting Jam to existing backends.
This will never happen in our integrations like Umbrel, as we predefine max fee values.
Additional requests to check for existing max fee values are only made if the taker operation reported the error `Action cannot be performed, config vars are not set.`.

I did not find a good place to make the check and opted to do it just after the original error happens. If we check for more things like that it might be good to do that on initial application start, but I refrained from doing so for this uncommon case for now. Feedback on how to make it more clear/readable is highly appreciated.

### DIFF
--- a/src/components/Send.jsx
+++ b/src/components/Send.jsx
@@ -113,11 +113,11 @@ const CollaboratorsSelector = ({ numCollaborators, setNumCollaborators, minNumCo
   )
 }
 
-const enhanceTakerErrorMessageIfPossible = async (requestContext, httpStatus, errorMessage) => {
+const enhanceTakerErrorMessageIfNecessary = async (requestContext, httpStatus, errorMessage) => {
   const configExists = (section, field) => Api.postConfigGet(requestContext, { section, field }).then((res) => res.ok)
 
-  const isConfigVarError = httpStatus === 409 && errorMessage === 'Action cannot be performed, config vars are not set.'
-  if (isConfigVarError) {
+  const tryEnhanceMessage = httpStatus === 409
+  if (tryEnhanceMessage) {
     const maxFeeSettingsPresent = await Promise.all([
       configExists('POLICY', 'max_cj_fee_rel'),
       configExists('POLICY', 'max_cj_fee_abs'),
@@ -126,10 +126,11 @@ const enhanceTakerErrorMessageIfPossible = async (requestContext, httpStatus, er
       .catch(() => false)
 
     if (!maxFeeSettingsPresent) {
-      return `
+      const maxFeeSettingsMissingMessage = `
         Config variables 'max_cj_fee_rel' and 'max_cj_fee_abs' must be set in your joinmarket.cfg in order to send collaborative transactions. 
         Consider adding them to your config manually.
       `
+      return `${errorMessage} ${maxFeeSettingsMissingMessage}`
     }
   }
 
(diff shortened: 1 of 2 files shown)


src/components/Send.jsx
```diff
@@ -113,6 +113,29 @@ const CollaboratorsSelector = ({ numCollaborators, setNumCollaborators, minNumCo
   )
 }
 
+const enhanceTakerErrorMessageIfPossible = async (requestContext, httpStatus, errorMessage) => {
+  const configExists = (section, field) => Api.postConfigGet(requestContext, { section, field }).then((res) => res.ok)
+
+  const isConfigVarError = httpStatus === 409 && errorMessage === 'Action cannot be performed, config vars are not set.'
```
dnlggr, line 119: Checking for an exact match with the error message might be a bit tricky as those can probably change without warning. I understand that there's probably no better way to do it. Just wanted to bring it up.

src/components/Send.jsx
```diff
@@ -113,6 +113,29 @@ const CollaboratorsSelector = ({ numCollaborators, setNumCollaborators, minNumCo
...
+  if (isConfigVarError) {
+    const maxFeeSettingsPresent = await Promise.all([
+      configExists('POLICY', 'max_cj_fee_rel'),
+      configExists('POLICY', 'max_cj_fee_abs'),
+    ])
+      .then((arr) => arr.every((e) => e === true))
```
dnlggr, line 125: Does this work?

```suggestion
      .then((arr) => arr.every((e) => e))
```

src/components/Send.jsx
```diff
@@ -113,6 +113,29 @@ const CollaboratorsSelector = ({ numCollaborators, setNumCollaborators, minNumCo
...
+      .catch(() => false)
+
+    if (!maxFeeSettingsPresent) {
+      return `
+        Config variables 'max_cj_fee_rel' and 'max_cj_fee_abs' must be defined in your configuration in order to send collaborative transactions.
```
dnlggr, line 130:
```suggestion
        Config variables 'max_cj_fee_rel' and 'max_cj_fee_abs' must be set in your joinmarket.cfg in order to send collaborative transactions. 
```

dnlggr (commented):
Nice, definitely a good additon

src/components/Send.jsx
theborakompanioni: Yeah, thinking the same. Do you think we should test for the config vars on any `409` error and just append the custom message?

theborakompanioni: Done.

dnlggr (approved):

PR #154, opened 2022-03-04
Fix readme links

secondl1ght:
I noticed that the Installation and Development links in the README were broken, so I updated them so that when you click on them they take you to the section again.


dnlggr (approved):
Thanks for the fix. 🧡

PR #158, opened 2022-03-07
fix: clear UI wallet if locking fails with status 401

### LINKED ISSUE #157: Improve lock/unlock mechanism
Got several support requests from Users being legitimately confused about wallet locking/unlocking.

- [x] Wallet with active maker is no be displayed as "Inactive", even if no token is available
- [x] Warn user that locking wallet will abort a taker operation (if running)
- [x] Warn user that locking wallet will abort a maker operation (if running)

User journey is really a bit strange currently.. closing a tab and "re-unlocking" wallet does not stop the maker - What should we aim at here? Would it be good to inform the user in a modal with the ability to keep the maker running and "just logout" (removing only the token, not locking the wallet)?

theborakompanioni:
This PR is a preparation for #157.

Most notable changes:
- add a "healthy" state to the websocket, which fixes the linear backoff interval if the token becomes invalid.
  socket would be opened correctly and error counter reset, shortly before the connection is closed by the server.
  Now, the error counter is reset only if the socket stays open at least short amount of time (currently 1000ms).

- Cleans the wallet if a user wants to lock the wallet with an invalid token. It will subsequently still display "Invalid Credentials" which might be confusing, but it's better than being trapped with an invalid token.


theborakompanioni:
#157 requires a little more refactoring. Having these changes included in the next bugfix release would help a lot.
Ready for review.

src/components/Wallet.jsx
```diff
@@ -45,33 +46,35 @@ export default function Wallet({ name, currentWallet, startWallet, stopWallet, s
...
 
       const res = await Api.getWalletLock({ walletName, token })
-      if (res.ok) {
-        const { walletname: lockedWalletName, already_locked } = await res.json()
+      const body = await res.json()
+
+      // on status OK or UNAUTHORIZED, stop the wallet and clear all local
+      // information. token might became invalid or another one has been
+      // issued for the same wallet, etc.
+      // in any case, user has no access to the wallet anymore.
```
dnlggr, line 61:
```suggestion
      // On status OK or UNAUTHORIZED, stop the wallet and clear all local
      // information. The token might have become invalid or another one might have been
      // issued for the same wallet, etc.
      // In any case, the user has no access to the wallet anymore.
```

src/components/Wallet.jsx
```diff
@@ -45,33 +46,35 @@ export default function Wallet({ name, currentWallet, startWallet, stopWallet, s
...
+      if (res.ok || res.status === 401) {
         stopWallet()
+      }
+
+      if (!res.ok) {
+        setAlert({ variant: 'danger', message: body.message })
+      } else {
+        const { walletname: lockedWalletName, already_locked } = body
```
dnlggr, line 69: I'd personally switch the branches here. The error case is less important and should come second imo. Not a hard change request though.

dnlggr (commented):
Nice, good change. The error message still saying "Invalid Credentials" is fine imo.

Issue #136, opened 2022-03-02
Show seed option in settings page


abhiShandy:
User should have an option to verify his seed backup by checking it on settings/wallet page

secondl1ght:
I started working on this but im having a bit of a hard time finding where the seed phrase is stored, can someone point me in the right direction?

theborakompanioni:
> I started working on this but im having a bit of a hard time finding where the seed phrase is stored, can someone point me in the right direction?

A `GET` to `/getseed` - see here: https://github.com/JoinMarket-Org/joinmarket-clientserver/blob/master/docs/api/wallet-rpc.yaml#L451-L476

secondl1ght:
Thanks, so I will need to manually make a request to api/v1/wallet/{walletname}/getseed? Or is there an API method I can call?

I tried entering it in the browser and get this response:

![image](https://user-images.githubusercontent.com/85003930/156410796-69f60749-214e-402e-a835-5f46ee7b8bac.png)

theborakompanioni:
Currently `/getseed` is not used so you'd have to add it to https://github.com/joinmarket-webui/joinmarket-webui/blob/master/src/libs/JmWalletApi.js - it does not work from the browsers address bar, as you need to send the jwt token in the authentication header. Just add it to the client and it should work.

secondl1ght:
Awesome thanks I will give this a try!

secondl1ght:
Hmm I am still getting the same message in the console even after adding it to the API and calling it from the client with the token and wallet name. I must be doing something wrong, I tried looking at the code for other API calls to see how it works.

secondl1ght:
I tested other API methods using the same auth token using `curl` in the terminal and they are working, but the `getSeed` one is getting the 405 response. Is it possible that we need to add some more code to the backend as well because we have not used this call before?

secondl1ght:
Just giving an update on this: I tried getting the new `getSeed` API working but ran into some difficulties, @theborakompanioni is going to put in a separate PR to get this added because he has it working. After I have that code I will use the new API call to finish this issue, I have the front-end pretty much ready just need the API call working.

secondl1ght:
PR is in #160

secondl1ght:
This has been merged

Issue #100, opened 2022-02-15
Improve wording


dnlggr:
We could/should go over the whole app and double check wording as most of it was done ad hoc during development. At least I didn't really put a lot of thought into it while developing. 😅

dnlggr:
cc @dergigi for this.

dnlggr:
This has been done through several inidividual PRs e.g. #277, #280.

PR #160, opened 2022-03-08
feat: add display seed phrase option to settings

secondl1ght:
As talked about in issue #136 I have added the option for the user to display their seed phrase backup after wallet initialization on the Settings screen. Big shout-out to @theborakompanioni for helping me debug this and get the new `getSeed` API call working! As part of this I created a new Seedphrase component to avoid code duplication and I also updated some global CSS styles affecting those elements. Thanks!

Here is how it looks:

 
image

image

image


src/components/Settings.jsx
```diff
@@ -1,11 +1,17 @@
...
 import PageTitle from './PageTitle'
+import Seedphrase from './Seedphrase'
+import ToggleSwitch from './ToggleSwitch'
 import { useSettings, useSettingsDispatch } from '../context/SettingsContext'
 import { SATS, BTC } from '../utils'
+import * as Api from '../libs/JmWalletApi'
 
 export default function Settings({ currentWallet }) {
+  const [seed, setSeed] = useState('')
+  const [showSeed, setShowSeed] = useState(false)
```
dnlggr, line 13:
```suggestion
  const [showingSeed, setShowingSeed] = useState(false)
```

Or something similar that makes it immediately clear what this flag does.

src/components/Settings.jsx
```diff
@@ -86,6 +92,44 @@ export default function Settings({ currentWallet }) {
...
+            e.preventDefault()
+            setRevealSensitiveInfo(false)
+            const { name: walletName, token } = currentWallet
+            const res = await Api.getSeed({ walletName, token })
+            if (res.ok) {
+              const { seedphrase } = await res.json()
+              setSeed(seedphrase)
+              setShowSeed(!showSeed)
+            } else {
+              console.warn('Could not retrieve seedphrase')
```
dnlggr, line 111: Let's use a user-visible warning here.

src/components/Settings.jsx
```diff
@@ -86,6 +92,44 @@ export default function Settings({ currentWallet }) {
...
+            }
+          }}
+        >
+          <Sprite symbol="mnemonic" width="24" height="24" className="me-2" />
+          {showSeed ? 'Hide' : 'Show'} seed phrase backup
```
dnlggr, line 116:
```suggestion
          {showSeed ? 'Hide' : 'Show'} seed phrase
```

src/components/Settings.jsx
```diff
@@ -86,6 +92,44 @@ export default function Settings({ currentWallet }) {
...
+        <br />
+
+        <rb.Button
+          variant="outline-dark"
+          className="border-0 mb-2 d-inline-flex align-items-center"
+          onClick={async (e) => {
+            e.preventDefault()
+            setRevealSensitiveInfo(false)
+            const { name: walletName, token } = currentWallet
+            const res = await Api.getSeed({ walletName, token })
```
dnlggr, line 105: We shouldn't load the seed on every button click. When the user clicks the button to hide the seedphrase, we don't want to load it again. We want to  load it only when the button is clicked to reveal the seed.

src/components/Settings.jsx
```diff
@@ -86,6 +92,44 @@ export default function Settings({ currentWallet }) {
...
+            </div>
+            <div className="mb-2">
+              <ToggleSwitch
+                label="Reveal sensitive information"
+                onToggle={(isToggled) => {
+                  setRevealSensitiveInfo(isToggled)
+                }}
+              />
+            </div>
+          </div>
```
dnlggr, line 131: Just a thought: Maybe a pop up that shows the seed with some additional warning and confirmation step before would be a better fit UI-wise here. But the current one is fine as well. I would add some warning text to the toggle, though.

dnlggr (commented):
👏

secondl1ght:
Thanks for the comments, I will make these updates!

src/components/Settings.jsx
secondl1ght: Ok I dont see the difference between these 2 personally but I will update to use `showingSeed`. :)

src/components/Settings.jsx
secondl1ght: What kind of warning text would you like for the toggle? I was thinking 'reveal sensitive information' lets the user know that it is important, should we add something on top of this?

secondl1ght:
Ok I made all the updates suggested, I just have a question about one comment about the warning message which I replied to.

secondl1ght:
Also here is how the error message looks when the API call fails:

image

secondl1ght:
One thing I noticed is that other places in the app use `setAlert` so communicate with the user, do we want to use that for the 'Could not retrieve seedphrase.' message or do you like the way it is?

src/components/Seedphrase.jsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react'
+
+export default function Seedphrase({ seedphrase, isBlurred = true }) {
+  return (
+    <div className="seedphrase slashed-zeroes d-flex flex-wrap">
+      {seedphrase.split(' ').map((seedWord, index) => (
+        <div key={index} className="d-flex py-2 ps-2 pe-3">
+          <span className="seedword-index text-secondary text-end">{index + 1}</span>
+          <span className="text-secondary">.&nbsp;</span>
+          <span className={isBlurred ? 'blurred-text' : ''}>{isBlurred ? 'random' : seedWord}</span>
```
dnlggr, line 10:
```suggestion
          <span className={isBlurred ? 'blurred-text' : ''}>{isBlurred ? 'abcdef' : seedWord}</span>
```

Unrelated but while we're touching this I just remembered that we wanted to change this to a word that's not in the BIP39 wordlist.

src/libs/JmWalletApi.js
```diff
@@ -21,6 +21,7 @@ const BASE_PATH = `${window.JM.PUBLIC_PATH}/api`
  * @param {string} token the bearer token
  * @returns an object containing the authorization header
  */
+
```
dnlggr, line 24:
```suggestion
```

src/libs/JmWalletApi.js
```diff
@@ -62,6 +63,13 @@ const getWalletDisplay = async ({ walletName, token, signal }) => {
   })
 }
 
+const getSeed = async ({ walletName, token, signal }) => {
```
dnlggr, line 66:
```suggestion
const getWalletSeed = async ({ walletName, token, signal }) => {
```

src/components/Settings.jsx
```diff
@@ -1,11 +1,18 @@
...
+import ToggleSwitch from './ToggleSwitch'
 import { useSettings, useSettingsDispatch } from '../context/SettingsContext'
 import { SATS, BTC } from '../utils'
+import * as Api from '../libs/JmWalletApi'
 
 export default function Settings({ currentWallet }) {
+  const [seed, setSeed] = useState('')
+  const [showingSeed, setShowingSeed] = useState(false)
+  const [revealSensitiveInfo, setRevealSensitiveInfo] = useState(false)
+  const [error, setError] = useState(false)
```
dnlggr, line 15:
```suggestion
  const [seedError, setSeedError] = useState(false)
```

src/components/Settings.jsx
```diff
@@ -1,11 +1,18 @@
...
+import Seedphrase from './Seedphrase'
+import ToggleSwitch from './ToggleSwitch'
 import { useSettings, useSettingsDispatch } from '../context/SettingsContext'
 import { SATS, BTC } from '../utils'
+import * as Api from '../libs/JmWalletApi'
 
 export default function Settings({ currentWallet }) {
+  const [seed, setSeed] = useState('')
+  const [showingSeed, setShowingSeed] = useState(false)
+  const [revealSensitiveInfo, setRevealSensitiveInfo] = useState(false)
```
dnlggr, line 14:
```suggestion
  const [revealSeed, setRevealSeed] = useState(false)
```

src/components/Settings.jsx
```diff
@@ -86,6 +93,54 @@ export default function Settings({ currentWallet }) {
...
+              }
+            } else {
+              setShowingSeed(!showingSeed)
+            }
+          }}
+        >
+          <Sprite symbol="mnemonic" width="24" height="24" className="me-2" />
+          {showingSeed ? 'Hide' : 'Show'} seed phrase
+        </rb.Button>
+        {error && (
```
dnlggr, line 124:
```suggestion
        {showingSeed && error && (
```

Let's only show the error in this case.

src/components/Settings.jsx
```diff
@@ -86,6 +93,54 @@ export default function Settings({ currentWallet }) {
...
+          <div className="text-danger" style={{ marginLeft: '1rem' }}>
+            Could not retreive seedphrase.
+          </div>
+        )}
+        {showingSeed && (
```
dnlggr, line 129:
```suggestion
        {showingSeed && !error && (
```

dnlggr (commented):
Nice good work! 👏  Almost there. Just a couple smaller things—naming mostly.

For the warning message I was thinking making it clear again that this is a potentially destructive thing to do. But let's tackle that separately (see #100).

src/components/Settings.jsx
theborakompanioni: > Just a thought: Maybe a pop up that shows the seed with some additional warning and confirmation step before would be a better fit UI-wise here. 

Would [`<rb.Modal>`](https://react-bootstrap.github.io/components/modal/#vertically-centered) would be a good fit for that?

src/components/Settings.jsx
secondl1ght: I dont think this will work because we only set `showingSeed` to be true if the API call works, if it fails then we do not set it to true because there is no seed to display and we show the error instead.

secondl1ght:
I resolved the conflicts and merged from master but I am getting this error now:

image

Did anyone get this when testing the PR that introduced the language support?

src/components/Settings.jsx
dnlggr: Ok 👍

src/components/Settings.jsx
```diff
@@ -112,6 +121,54 @@ export default function Settings({ currentWallet }) {
             </rb.Dropdown.Item>
           </rb.Dropdown.Menu>
         </rb.Dropdown>
+
+        <br />
+
+        <rb.Button
```
dnlggr, line 127: Last thing, can you remove the `<br />` wrap everything in a `<div>` please? The layout is off otherwis:



Screenshot 2022-03-09 at 17 49 53

src/components/Settings.jsx
dnlggr: We can do that in an improvement issue as well though.

src/components/Settings.jsx
secondl1ght: Yes sorry I was not able to view the UI yet to see this because of my error.

secondl1ght:
Hmm that is weird let me keep trying.

secondl1ght:
@theborakompanioni fixed it with #168

src/components/Settings.jsx
secondl1ght: Ok sounds good

secondl1ght:
Ok all comments have been resolved so I think this is ready for final review and merging! I will put in another issue to track the discussion on turning this into a modal.

public/sprite.svg
```diff
@@ -21,5 +21,6 @@
...
 <symbol id="caret-up" viewBox="0 0 24 24" fill="currentColor"><path fill-rule="evenodd" d="M12.248 8.237a.25.25 0 00-.354 0L5.53 14.601a.75.75 0 11-1.06-1.06l6.363-6.364a1.75 1.75 0 012.475 0l6.364 6.364a.75.75 0 01-1.06 1.06l-6.364-6.364z" clip-rule="evenodd"></path></symbol>
 <symbol id="caret-down" viewBox="0 0 24 24" fill="currentColor"><path fill-rule="evenodd" d="M4.47 9.4a.75.75 0 011.06 0l6.364 6.364a.25.25 0 00.354 0L18.612 9.4a.75.75 0 011.06 1.06l-6.364 6.364a1.75 1.75 0 01-2.475 0L4.47 10.46a.75.75 0 010-1.06z" clip-rule="evenodd"></path></symbol>
+<symbol id='mnemonic' xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" fill="currentColor" data-v-4fa90e7f=""><path fill-rule="evenodd" d="M5 6.5A1.5 1.5 0 016.5 5h11A1.5 1.5 0 0119 6.5v11a1.5 1.5 0 01-1.5 1.5h-11A1.5 1.5 0 015 17.5v-11zm1.5 0H9V8H6.5V6.5zM9 9.667H6.5v1.5H9v-1.5zm-2.5 3.166H9v1.5H6.5v-1.5zM9 16H6.5v1.5H9V16zm1.75-9.5h2.5V8h-2.5V6.5zm2.5 3.167h-2.5v1.5h2.5v-1.5zm-2.5 3.166h2.5v1.5h-2.5v-1.5zM13.25 16h-2.5v1.5h2.5V16zM15 6.5h2.5V8H15V6.5zm2.5 3.167H15v1.5h2.5v-1.5zM15 12.833h2.5v1.5H15v-1.5zM17.5 16H15v1.5h2.5V16z" clip-rule="evenodd"></path></symbol>
```
dnlggr, line 24:
```suggestion
<symbol id='mnemonic' xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" fill="currentColor"><path fill-rule="evenodd" d="M5 6.5A1.5 1.5 0 016.5 5h11A1.5 1.5 0 0119 6.5v11a1.5 1.5 0 01-1.5 1.5h-11A1.5 1.5 0 015 17.5v-11zm1.5 0H9V8H6.5V6.5zM9 9.667H6.5v1.5H9v-1.5zm-2.5 3.166H9v1.5H6.5v-1.5zM9 16H6.5v1.5H9V16zm1.75-9.5h2.5V8h-2.5V6.5zm2.5 3.167h-2.5v1.5h2.5v-1.5zm-2.5 3.166h2.5v1.5h-2.5v-1.5zM13.25 16h-2.5v1.5h2.5V16zM15 6.5h2.5V8H15V6.5zm2.5 3.167H15v1.5h2.5v-1.5zM15 12.833h2.5v1.5H15v-1.5zM17.5 16H15v1.5h2.5V16z" clip-rule="evenodd"></path></symbol>
```

PR #162, opened 2022-03-08
Fix page reloads

### LINKED ISSUE #161: Fix page reloads
As a user, when refreshing the page, I want it to behave like a normal site and not redirect me to `/` every damn time.
Reproduce: Press F5 on a subpage like `/send`.

theborakompanioni:
Closes #161 .

This PR adapts the wallet initialization.

Before these changes, the `currentWallet` was initialized with `null` and subsequently updated with the values of a session object in an `useEffect` callback. This causes the route setup for components that need a loaded wallet to redirect to `/` on a page refresh, because they are conditionally added with `{currentWallet && (<Route element= [...] />)}`) and `currentWallet` is always `null` on initial rendering.

After these changes, the `currentWallet` will be initialized directly from the session object, if available.
If no session is available, the behaviour stays the same - it will redirect to `/`. However, if a session is available, it will be able to reload the page normally.

I was not sure at first if some components depend on the wallet being initialized "lazily".. but it seems there is really no difference, except page reloads are now working. Would be great if someone can verify this.

Edit: Some `update on unmounted component` errors are also fixed. These are now more easily reproduced by being able to fresh pages.

### DIFF
--- a/src/components/CurrentWalletMagic.jsx
+++ b/src/components/CurrentWalletMagic.jsx
@@ -63,7 +63,7 @@ const PrivacyLevel = ({ numAccounts, level, balance }) => {
 export default function CurrentWalletMagic() {
   const settings = useSettings()
   const settingsDispatch = useSettingsDispatch()
-  const wallet = useCurrentWallet()
+  const currentWallet = useCurrentWallet()
   const walletInfo = useCurrentWalletInfo()
   const setWalletInfo = useSetCurrentWalletInfo()
 
@@ -72,7 +72,7 @@ export default function CurrentWalletMagic() {
 
   useEffect(() => {
     const abortCtrl = new AbortController()
-    const { name: walletName, token } = wallet
+    const { name: walletName, token } = currentWallet
 
     setAlert(null)
     setIsLoading(true)
@@ -83,10 +83,10 @@ export default function CurrentWalletMagic() {
       .catch((err) => {
         !abortCtrl.signal.aborted && setAlert({ variant: 'danger', message: err.message })
       })
-      .finally(() => setIsLoading(false))
+      .finally(() => !abortCtrl.signal.aborted && setIsLoading(false))
 
     return () => abortCtrl.abort()
-  }, [wallet, setWalletInfo])
+  }, [currentWallet, setWalletInfo])
 
   return (
     <div className="privacy-levels">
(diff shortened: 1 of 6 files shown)


src/context/WalletContext.jsx
```diff
@@ -1,9 +1,21 @@
...
 
+const initialWalletFromSession = () => {
+  const session = getSession()
+  return !(session && session.name && session.token)
+    ? null
+    : {
+        name: session.name,
+        token: session.token,
+      }
+}
```
dnlggr, line 15:
```suggestion
  return session && session.name && session.token
    ? {
        name: session.name,
        token: session.token,
      }
    : null
```

src/components/CurrentWalletMagic.jsx
```diff
@@ -83,10 +83,10 @@ export default function CurrentWalletMagic() {
       .catch((err) => {
         !abortCtrl.signal.aborted && setAlert({ variant: 'danger', message: err.message })
       })
-      .finally(() => setIsLoading(false))
+      .finally(() => !abortCtrl.signal.aborted && setIsLoading(false))
```
dnlggr, line 86: Ah I see, that caused the "unmounted component state change" warnings, right?

dnlggr (commented):
Awesome 🚀  Tested and works so much better now! 👏 😄 

> There is still an updated triggered which leads useEffect hooks be evaluated and e.g. causes previous unfinished requests to be aborted. However, this happens only the first time the page is loaded.

Can you elaborate a bit on this?

src/components/CurrentWalletMagic.jsx
theborakompanioni: Yep, think we still have a few of them. Low priority, but we'll become more and more aware in code reviews, I guess.

src/context/WalletContext.jsx
theborakompanioni: I try to keep the longer expressions at the end, think it is easier for the eyes. But I get your point. Applied!

theborakompanioni:
> Can you elaborate a bit on this?

Yeah, what seems to happen is that `App` will update `currentWallet` _with the same values_ in `startWallet` due to the initialization in here: https://github.com/joinmarket-webui/joinmarket-webui/blob/v0.0.3/src/components/App.jsx#L86-L91. This in turn will trigger the abort action of the `useEffect` hooks as `currentWallet` holds a different object. Have to figure out if this part is now obsolete, as it seems to really just be a one time action. Will test that and report back.

dnlggr (approved):
LGTM 🎉

PR #165, opened 2022-03-08
fix: receive address copy button not working

### LINKED ISSUE #159: The copy button in the recieve tab does not work.
When trying to copy the receiving address, the "Copy" button does not work.

I am using the app on Umbrel, on Firefox on a Mac.

neb-b:
resolves #159 

I tested this in brave, safari, and firefox using an http tunnel with [ngrok](https://ngrok.com/), which should be similar to the app running in umbrel

While working on this I discovered that `document.execCommand` is deprecated, but still has good browser support, and  [umbrel is using the same pattern for it's copy behavior](https://github.com/getumbrel/umbrel-dashboard/blob/master/src/components/Utility/InputCopy.vue#L74).

I tried to avoid using `position: absolute` and instead making an inline invisible input but `inputRef.current.select()` doesn't work properly with that, and just selects an empty string. 

Really excited about this joinmarket webui project and hope to contribute more in the future!

### DIFF
--- a/src/components/Receive.jsx
+++ b/src/components/Receive.jsx
@@ -1,5 +1,5 @@
 import React from 'react'
-import { useState, useEffect } from 'react'
+import { useState, useEffect, useRef } from 'react'
 import { BitcoinQR } from './BitcoinQR'
 import { useLocation } from 'react-router-dom'
 import * as rb from 'react-bootstrap'
@@ -12,6 +12,7 @@ import Sprite from './Sprite'
 export default function Receive({ currentWallet }) {
   const location = useLocation()
   const settings = useSettings()
+  const inputRef = useRef()
   const [validated, setValidated] = useState(false)
   const [alert, setAlert] = useState(null)
   const [isLoading, setIsLoading] = useState(false)
@@ -89,15 +90,15 @@ export default function Receive({ currentWallet }) {
                   data-bs-toggle="tooltip"
                   data-bs-placement="left"
                   onClick={() => {
-                    // might not work on iOS.
-                    navigator.clipboard.writeText(address).then(
-                      () => {
-                        setAddressCopiedFlag(addressCopiedFlag + 1)
-                      },
-                      () => {
-                        setAlert({ variant: 'warning', message: 'Could not copy address.' })
-                      }
-                    )
+                    inputRef.current.select()
+                    const success = document.execCommand('copy')
+
+                    if (!success) {
+                      setAlert({ variant: 'warning', message: 'Could not copy address.' })
+                      return
+                    }
+
+                    setAddressCopiedFlag(addressCopiedFlag + 1)
                   }}
                 >
                   {showAddressCopiedConfirmation ? (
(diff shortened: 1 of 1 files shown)


src/components/Receive.jsx
```diff
@@ -109,6 +110,16 @@ export default function Receive({ currentWallet }) {
                     'Copy'
                   )}
                 </rb.Button>
+                <input
+                  aria-hidden
+                  ref={inputRef}
+                  defaultValue={address}
```
theborakompanioni, line 116:
```suggestion
                  value={address}
```
Just noticed, that when you click "Get new address", the value of the input changes, however, it will only ever copy the first address. Using `value` instead of `defaultValue` seems to fix it.

src/components/Receive.jsx
```diff
@@ -109,6 +110,16 @@ export default function Receive({ currentWallet }) {
...
+                  style={{
+                    position: 'absolute',
+                    left: '-9999px',
+                    top: '-9999px',
+                  }}
+                />
```
dnlggr, line 122:
```suggestion
                <input
                  readOnly
                  aria-hidden
                  ref={inputRef}
                  defaultValue={address}
                  style={{
                    position: 'absolute',
                    left: '-9999px',
                    top: '-9999px',
                  }}
                />
```

Stops the keyboard from popping up when selecting the text on mobile.

dnlggr (commented):
Works nicely! The deprecated API is a bit unfortunate but we probably just have to live with it. Better than it not working for a large group of users.

theborakompanioni:
Nice @neb-b ! 

What do you think about keeping the original behaviour but fallback to the deprecated version if necessary?

e.g. something like this:
```js
const copyToClipboard = (text, fallbackInputField) => {
  const copyToClipboardFallback = (inputField) => new Promise((resolve, reject) => {
    inputField.select()
    const success = document.execCommand && document.execCommand('copy')
    inputField.blur()
    success ? resolve(success) : reject(new Error('Could not copy address.'))
  })

  // `navigator.clipboard` might not be available, e.g. on sites served over plain `http`.
  if (!navigator.clipboard) {
    return copyToClipboardFallback(fallbackInputField);
  }

  // might not work on iOS.
  return navigator.clipboard.writeText(text)
    .catch(() => copyToClipboardFallback(fallbackInputField))
}
```
and invoked via:
```jsx
<rb.Button
  variant="outline-dark"
  data-bs-toggle="tooltip"
  data-bs-placement="left"
  onClick={() => {
    copyToClipboard(address, inputRef.current).then(
      () => {
        setAddressCopiedFlag(addressCopiedFlag + 1)
      },
      (e) => {
        setAlert({ variant: 'warning', message: e.message })
      }
    )
  }}
>[..]</rb.Button>
```

Any thoughts? Do you think it is possible to write a test for this?

src/components/Receive.jsx
```diff
@@ -12,6 +12,7 @@ import Sprite from './Sprite'
 export default function Receive({ currentWallet }) {
   const location = useLocation()
   const settings = useSettings()
+  const inputRef = useRef()
```
theborakompanioni, line 15: In order for others not to guess or scroll around too much, maybe we can come up with a more descriptive name. Instead of `inputRef` maybe `addressCopyFallbackInputRef`?

neb-b:
Thanks for the quick review. @theborakompanioni I can write a test for this, although it will just be testing that the copy functions will have been called, not that the actual copy to clipboard works.

If you would like that is there a way to quickly get to the receive screen? Or will the test also have to type in the password/unlock the wallet/navigate to the receive tab?

dnlggr (approved):
LGTM ✅

PR #166, opened 2022-03-09
ui: display message if utxos are empty

theborakompanioni:
Felt awkward that only the button text changes from "Show UTXOs" to "Hide UTXOs" if the current wallet has no unspent outputs yet. These changes will display a message instead of showing nothing.

Edit: Too small of a change.. will add thorough tests for both wallet components (advanced and magic) later if that's okay.


src/components/CurrentWalletAdvanced.jsx
```diff
@@ -68,17 +68,24 @@ export default function CurrentWalletAdvanced() {
...
+              setShowUTXO(!showUTXO)
+            }}
+            className="mb-3"
+          >
+            {showUTXO ? 'Hide UTXOs' : 'Show UTXOs'}
+          </rb.Button>
+          <rb.Fade in={showUTXO} mountOnEnter={true} unmountOnExit={true}>
+            <div>
+              {utxos.length === 0 && <rb.Alert variant="info">No UTXOs available.</rb.Alert>}
+              <DisplayAccountUTXOs utxos={utxos} className="mt-3" />
```
dnlggr, line 84: Does the following make sense?

```suggestion
              {utxos.length === 0 ? <rb.Alert variant="info">No UTXOs available.</rb.Alert> :
              <DisplayAccountUTXOs utxos={utxos} className="mt-3" />}
```

Might need to run the formatter over this though. 😉

dnlggr (approved):
✅ 👏

PR #167, opened 2022-03-09
fix: change misdepth order

dnlggr:
As discussed in yesterday's call, un-reverses the mixdepth ordering.


theborakompanioni (approved):
lgtm :heavy_check_mark:

PR #168, opened 2022-03-09
fix: use resolvedLanguage in language picker

theborakompanioni:
Ahh, the ghost in the machine revealed itself.

This uses `resolvedLanguage` (e.g, 'en') instead of `language` (e.g. 'en-US').
From the docs of `resolvedLanguage`:
> Is set to the current resolved language.
   It can be used as primary used language, for example in a language switcher.

It might still be unset, which is the case in unit tests, however it has not happened in a real browser. I have to read up on this, but as a quickfix it might be enough.

### DIFF
--- a/src/components/Settings.jsx
+++ b/src/components/Settings.jsx
@@ -93,7 +93,7 @@ export default function Settings({ currentWallet }) {
         <rb.Dropdown>
           <rb.Dropdown.Toggle variant="outline-dark" className="border-0 mb-2 d-inline-flex align-items-center">
             <Sprite symbol="globe" width="24" height="24" className="me-2" />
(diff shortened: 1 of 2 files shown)


dnlggr (approved):

dnlggr:
Thanks!

dnlggr:
I wonder why it even worked on my machine. Do you happen to know @theborakompanioni? 🤔

theborakompanioni:
> I wonder why it even worked on my machine. Do you happen to know what it could have been @theborakompanioni? thinking

I'm not entirely sure either, as you can see on the proposed quick-fix. I think it depends on your current language settings, as well as that you already played around with the settings. I know too little about the inner workings of i18next to say for sure. But there must be a better solution to this.

PR #171, opened 2022-03-10
feat: satscomma formatting for bitcoin balances

dnlggr:
Uses satscomma-style [^1][^2][^3] formatting for the fractional part of a bitcoin balance. As mentioned in the last community call, this will make it way easier to parse the amount of sats in a balance displayed in bitcoin. Decided to go with a half-space as the digit group separator. Looks the cleanest imo. Also did some minor refactoring of the `Balance` component itself.

## 📸 

Screenshot 2022-03-10 at 17 11 05

[^1]: https://bitcoin.design/guide/payments/units-and-symbols/#satcomma
[^2]: https://medium.com/@mark.nugent.iv/grouping-bitcoins-fractional-digits-an-idea-whose-time-has-come-22d9dad8ac51
[^3]: https://medium.com/coinmonks/the-satcomma-standard-89f1e7c2aede

### DIFF
--- a/src/components/Balance.test.jsx
+++ b/src/components/Balance.test.jsx
@@ -42,6 +42,26 @@ describe('<Balance />', () => {
     expect(screen.getByText(`0`)).toBeInTheDocument()
   })
 
+  it('should render a large string BTC value correctly as BTC', () => {
+    render(<Balance valueString={'20999999.97690000'} convertToUnit={BTC} showBalance={true} />)
+    expect(screen.getByText(`20,999,999.97 690 000`)).toBeInTheDocument()
+  })
+
+  it('should render a large string BTC value correctly as SATS', () => {
+    render(<Balance valueString={'20999999.97690000'} convertToUnit={SATS} showBalance={true} />)
+    expect(screen.getByText(`2,099,999,997,690,000`)).toBeInTheDocument()
+  })
+
+  it('should render a max string BTC value correctly as BTC', () => {
+    render(<Balance valueString={'21000000.00000000'} convertToUnit={BTC} showBalance={true} />)
+    expect(screen.getByText(`21,000,000.00 000 000`)).toBeInTheDocument()
+  })
+
+  it('should render a max string BTC value correctly as SATS', () => {
+    render(<Balance valueString={'21000000.00000000'} convertToUnit={SATS} showBalance={true} />)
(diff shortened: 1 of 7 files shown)


src/components/Balance.test.jsx
```diff
@@ -0,0 +1,76 @@
+import React from 'react'
+import { render, screen } from '../testUtils'
+import { BTC, SATS } from '../utils'
+
+import Balance from './Balance'
+
+describe('<Balance />', () => {
```
theborakompanioni, line 7: Do you think it would be good to add tests for `0`, `1`, [`2099999997690000`](https://en.bitcoin.it/wiki/Controlled_supply#Projected_Bitcoins_Long_Term), `2100000000000000` sats too?
Thinking of potential floating point issues - JM needs some amounts denominated in `BTC`, we should be confident to handle these well -

src/components/Balance.test.jsx
```diff
@@ -0,0 +1,76 @@
...
+
+  it('should hide balance for SATS by default', () => {
+    render(<Balance value={'123'} unit={SATS} />)
+    expect(screen.getByText(`*****`)).toBeInTheDocument()
+    expect(screen.queryByText(`123`)).not.toBeInTheDocument()
+  })
+
+  it('should render a string BTC value correctly as BTC', () => {
+    render(<Balance value={'123.03224961'} unit={BTC} showBalance={true} />)
+    expect(screen.getByText(`123.03 224 961`)).toBeInTheDocument()
```
theborakompanioni, line 22:
```suggestion
    expect(screen.getByText(`123.03 224 961`)).toBeInTheDocument()
    expect(screen.get(`₿`)).toBeInTheDocument()
```
Should we test the currency sign as well?

theborakompanioni (approved):
Nice addition! It's so pleasing to know that some people care about, and keep track of following community best practices. :muscle:
And most of all: Thanks for writing tests.

src/components/Balance.test.jsx
dnlggr: Left it out since that would only work for BTC but not for sats (svg).

src/components/Balance.test.jsx
dnlggr: I updated the value parsing and tests a bit, please have another look and let me know if it is at least a bit clearer now. Still not perfect but I hope that what is happening is at least a bit more transparent now.

theborakompanioni (approved):

PR #176, opened 2022-03-15
Add typescript support

theborakompanioni:
This commit adds typescript support (`.ts`/`.tsx`).

- called `npm install --save-dev typescript @types/node @types/react @types/react-dom @types/jest`
- created `tsconfig.json` with `npx tsc --init`
- ~~added missing `include` option to `tsconfig.json`~~
- added `jsx: "react-jsx"` ~~and `noImplicitAny: false`~~ to `compilerOptions` in `tsconfig.json`

##### Examples
- ported `libs/JmWalletApi.js` to `libs/JmWalletApi.ts` and added exemplary types
- ported helper `session.js` to `session.ts`
- ported component `CurrentWalletAdvanced.jsx` to `CurrentWalletAdvanced.tsx`

This was just to easy... everything seems to work quite smoothly - I might be missing something (?)..


theborakompanioni:
> Maybe remove the commented out stuff from the config file?

:heavy_check_mark: 

Still have some issues with destructing object params, but I guess nothing that cannot be fixed.
Will ping you once progress is made and the PR is ready.

theborakompanioni:
> Awesome, sounds good!


This is now ready to the point where it has a couple of examples and a base that can be built upon. Is it okay for you if we implement this step by step? I did not want to start a full refactoring. It would be very helpful already in PR #175 .

Some shortcuts have been taken, as I just wanted to showcase at least porting one utils, one component and one test file.
e.g. `type` definitions of `Utxos` and `Alert` in `CurrentWalletAdvanced` are misplaced and must be adapted and move in the future.

src/components/CurrentWalletAdvanced.test.tsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react'
+// @ts-ignore
```
dnlggr, line 2: What does that do / why is it needed? 🤔

src/libs/JmWalletApi.ts
```diff
@@ -0,0 +1,295 @@
...
+const getWalletLock = async ({ token, signal, walletName }: WalletRequestContext) => {
+  return await fetch(`${BASE_PATH}/v1/wallet/${walletName}/lock`, {
+    headers: { ...Authorization(token) },
+    signal,
+  })
+}
+
+interface WalletUnlockRequest {
+  password: string
+}
```
dnlggr, line 119: Should we move all interface/type declarations to the top of the file for better discoverability?

src/components/CurrentWalletAdvanced.tsx
```diff
@@ -1,49 +1,69 @@
...
     setAlert(null)
     setIsLoading(true)
 
-    const loadingWallet = Api.getWalletDisplay({ walletName: name, token, signal: abortCtrl.signal })
+    const loadingWallet = Api.getWalletDisplay({ walletName, token, signal: abortCtrl.signal })
       .then((res) =>
-        res.ok ? res.json() : Promise.reject(new Error(res.message || t('current_wallet.error_loading_failed')))
+        res.ok
+          ? res.json()
+          : Api.Helper.toError(res, t('current_wallet.error_loading_failed')).then((e) => {
```
dnlggr, line 50: Would it work to throw the error inside `toError` (which would then need to be renamed to `throwError` or something like that)? Then we could get rid of the nested `.then()` call which is a bit cumbersome to read/write imo.

dnlggr (commented):
Cool stuff! I don't think we need  to rewrite everything, we can just gradually write new stuff in Typescript if we want to and keep the old code as is as long as it's not in the way.

src/components/CurrentWalletAdvanced.test.tsx
theborakompanioni: It is needed in the transition while we support js and ts.
If the line is not ignored by ts, the following warning is emitted:
> `Could not find a declaration file for module '../testUtils'. 'joinmarket-webui/src/testUtils.jsx' implicitly has an 'any' type.ts(7016)`

Either ignore the line with `// @ts-ignore` temporarily or disable `noImplicitAny` compiler options.
I wanted to go with the former, as removing the option seems to induce more work later on.

So it is just needed when importing code via object destructuring from js in a ts file. 
Do you think we should leave it that way or is there maybe an other, better solution to this?

dnlggr: Ok I see. Let's leave it like this! ✅

dnlggr (approved):
Really nice! ✅ 👏

PR #178, opened 2022-03-16
Update regtest env

theborakompanioni:
Small cleanups and improvements for the regtest environment:
- remove custom `gaplimit` in jm containers (as it is not really needed and we should aim at using defaults wherever it is possible)
- remove unused tor config in primary jm container (left unchanged for secondary container)
- reorder container definitions according to their "importance" (e.g. jm containers first)
- add npm scripts to interact with regtest env

e.g. when you want to use the latest update for the regtest environment, instead of
```sh
docker-compose -f ./docker/regtest/docker-compose.yml build --pull --no-cache
```
you can now just execute:
```sh
npm run regtest:rebuild
```

### DIFF
--- a/docker/regtest/docker-compose.yml
+++ b/docker/regtest/docker-compose.yml
@@ -109,7 +109,6 @@ services:
     environment:
       READY_FILE: /root/.nbxplorer/btc_fully_synched
       ENSURE_WALLET: 1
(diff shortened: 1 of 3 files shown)


dnlggr (approved):
Good idea using npm scripts for this. 👍

PR #184, opened 2022-03-18
feat: sweep mixdepths

### LINKED ISSUE #133: Sweep (send all) functionality in on send page
As [mentioned by Adam in Telegram](https://t.me/JoinMarketWebUI/1191). This is a central feature on Joinmarket since it allows an arbitrary-amount transaction without any toxic change.

I'll put this in v0.1.0. Feel free to move around @dergigi! 🙏

dnlggr:
Resolves #133.

⚠️ I think eventually we should move sending/sweeping functionality towards the main page and make sweeping accounts/mixdepths/pockets more prominent since it is one of distinguishing features of Joinmarket.

For now this adds sweeping functionality to the current send page.

The most tricky part was how to communicate locked or frozen UTXOs (which will not be swept). When sweeping we now show the user the total balance of all UTXOs in a mixdepth (including frozen and locked ones) and in addition to that alert him that _at least_ the sum of all frozen or locked UTXOs will not be sent.

I decided against doing the calculation `total_balance - frozen_balance - locked_balance` and show just that amount since it didn't feel right to replicate this business logic on the client (what if the backend behavior changes?). Therefore, I'm only showing which part of the total balance definitely won't be sent. This is a bit more conservative and cumbersome for the user but also less error prone potentially.

Let me know what you think about this and if I've been to careful here.

## 📸 

Screenshot 2022-03-18 at 11 31 07

Screenshot 2022-03-18 at 11 31 09


MaxHillebrand:
Nice addition, thanks!

If we know that some coins for sure are locked or frozen, then I think their value should already be subtracted from the amount textbox.

Underneath can be a warning with `maybe less sats will be sent, some coins might be locked or frozen.`

[maybe I misunderstood the problem]

If senders will announce to the merchant the amount of the payment, then they would say what is shown in the textbox. So making the calculation accurate would be the preferred solution.

theborakompanioni:
> Nice addition, thanks!

:orange_heart: :rocket: 

> If we know that some coins for sure are locked or frozen, then I think their value should already be subtracted from the amount textbox.

That'd be the preferred way of course, however I guess @dnlggr wanted to avoid exactly that for reasons stated in the description. Linking this issue here: [Spendable balance via api joinmarket-clientserver#1167](https://github.com/JoinMarket-Org/joinmarket-clientserver/issues/1167
)

dnlggr:
> [maybe I misunderstood the problem]
>
> If senders will announce to the merchant the amount of the payment, then they would say what is shown in the textbox. So making the calculation accurate would be the preferred solution.

I was probably a bit imprecise in my description, my that's my fault! All of this discussion refers to UI only. Independendtly from what we show in the amount field in case of a sweep, what we actually send to the Joinmarket backend is `amount: 0`. This then makes Joinmarket gather all UTXOs that can be sent (not frozen and not locked) and from those calculate the actual amount to send. Does that explanation make more sense?

I wasn't sure if I should replicate that calculation on the client side but reading the issue @theborakompanioni linked it seems like it should be fairly safe to do so.

MaxHillebrand:
Thank you.

Could we send this call to joinmarket when the user presses the sweep button, but in a way that the actual signing is only initiated when the user confirms the payment?

dnlggr:
> Could we send this call to joinmarket when the user presses the sweep button, but in a way that the actual signing is only initiated when the user confirms the payment?

I don't think that's possible at the moment. But I see where you're going at I think. It's weird showing the user some amount that we calculate locally when the backend later does it's own computation before sending (even though when the outcome should _theoretically_ be the same).

That's basically the reason for me not doing the calculation of the actually spendable amount on the UI side but keeping it more vague and just telling the user "You have X balance in this account from which Y are definitely no spendable". Again, not sure if this is the perfect way of doing it. Might be too cautious—but maybe it's better to be cautious for these things. 🤷‍♂️

theborakompanioni:
>> Could we send this call to joinmarket when the user presses the sweep button, but in a way that the actual signing is only initiated when the user confirms the payment?

That would probably be the best of all cases. :+1:

> That's basically the reason for me not doing the calculation of the actually spendable amount on the UI side but keeping it more vague and just telling the user "You have X balance in this account from which Y are definitely no spendable". Again, not sure if this is the perfect way of doing it. Might be too cautious—but maybe it's better to be cautious for these things. :man_shrugging:

Do you think the current approach is acceptable as a first version @MaxHillebrand ?

MaxHillebrand:
Current version is good enough for what we can do now. 

I would still change the textbox amount to exclude the known frozen coins, but keep the notification as is, so to show how that value is calculated and what total mix depth value is.

theborakompanioni (approved):
Looks great!

Some thoughts:
- The form feedback (`t('send.feedback_invalid_amount')` is gone. If it is readded, the sweep button will be misaligned - should we try to bring it back in as tooltip?
i.e
```
<rb.Form.Control.Feedback form="send-form" type="invalid" tooltip={true}>
  {t('send.feedback_invalid_amount')}
</rb.Form.Control.Feedback>
```
I am unsure, as it is favorable to use one approach throughout the app. Just wondering if we should show form validation feedack as tooltip in general.

- With typescript support and more and more tests written, development will be much more pleasant. Tests for the Send component would be cool, but no blocker. We should try to add them for the most important functionality before v0.1.0.

- Method `balanceBreakdown` might return `null`, but it is used as if it always returns a value in the view (`valueString={balanceBreakdown(account).totalBalance.toString()}`) - just wanted to say that I am looking forward to use Typescript for all of the components soon :muscle:

dnlggr:
> I would still change the textbox amount to exclude the known frozen coins, but keep the notification as is, so to show how that value is calculated and what total mix depth value is.

Screenshot 2022-03-21 at 14 23 33

Thoughts on a calculation breakdown like this? ☝️ (Exact text TBD).

MaxHillebrand:
That looks good - but maybe a lot of clutter... how about you make it an optional drop down, or tooltip or something?

src/components/Send.jsx
```diff
@@ -327,6 +343,133 @@ export default function Send({ makerRunning, coinjoinInProcess }) {
...
+      <div className="sweep-breakdown mt-1">
+        <rb.Accordion flush>
+          <rb.Accordion.Item eventKey="0">
+            <rb.Accordion.Header>
+              <div className="d-flex align-items-center justify-content-end w-100">
+                {t('send.button_sweep_amount_breakdown')}
+              </div>
+            </rb.Accordion.Header>
+            <rb.Accordion.Body className="px-0">
+              <table class="table table-sm">
```
theborakompanioni, line 403:
```suggestion
              <table className="table table-sm">
```

theborakompanioni: Also, the text inside `table` is invisible in dark mode. (`color: #212529;` in _tables.scss)

dnlggr: Yep still have to test dark mode!

dnlggr:
> Good question. Reloading seems a bit excessive but on the other hand, if someone does multiple payments while staying on the send page without reloading, data would be wrong. So we probably have to do it 🤔

I played around for a bit but I couldn't really find a good solution. An instant reload after send is super confusing since the page disapears briefly. Feels like something with the send went wrong. Ok if I track this as an improvement and tackle separately before the next release ideally?

dnlggr:
Tracked https://github.com/joinmarket-webui/joinmarket-webui/issues/187. Hope this is ok @theborakompanioni. Let me know.

dnlggr:
Also tacked https://github.com/joinmarket-webui/joinmarket-webui/issues/188 for tests.

GBKS:
The Sweet button with the black background is taking over the visual hierarchy, it's the first thing my eyes are drawn to - even though it's not the most important element on the screen. May want to remove the background and use have the icon and label (or a white background).

theborakompanioni:
> The "Sweep" button with the black background is taking over the visual hierarchy, it's the first thing my eyes are drawn to - even though it's not the most important element on the screen. May want to remove the background and use have the icon and label (or a white background).

Do you think it is acceptable to use the existing buttons (changes background color on hover) for this? Or is it still too distracting? It is the only button rendered inside an input element at the moment - it would also make sense to define a custom style (e.g. soften the border color to match the input element in "light" mode).

## 📸 
Screenshot 2022-03-23 at 11 31 07 Screenshot 2022-03-23 at 11 31 08

GBKS:
@theborakompanioni that works.

The [JoinMarket Figma file here](https://www.figma.com/file/kfejZJFlwBywvLEnPEmJo1/JoinMarket-UI?node-id=96%3A2639) has a "free" style for buttons that can be easily applied (in addition to filled and outline). Maybe this could become a general style? Might come in handy in other places as well.

theborakompanioni:
> The [JoinMarket Figma file here](https://www.figma.com/file/kfejZJFlwBywvLEnPEmJo1/JoinMarket-UI?node-id=96%3A2639) has a "free" style for buttons that can be easily applied (in addition to filled and outline). Maybe this could become a general style? Might come in handy in other places as well.

Nice, yeah. Integrating a "free" link style button without underlined text looks good. Currently, a slightly adapted stylesheet from `react-bootstrap` ([see the buttons page](https://react-bootstrap.github.io/components/buttons/)) is used, which is a little buggy here and there.

I will merge this for now as it can be revisited in #178 and #188

editwentyone:
is this helping? 
Bildschirmfoto 2022-04-26 um 15 35 55

PR #195, opened 2022-03-29
chore(release): prepare v0.0.5

dergigi:
Prepares the v0.0.5 release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,20 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+### [0.0.5](https://github.com/joinmarket-webui/joinmarket-webui/compare/v0.0.4...v0.0.5) (2022-03-29)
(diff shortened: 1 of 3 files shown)


theborakompanioni (approved):
Maybe we should pin the npm version (v8.3.1 here), as your changes to package-lock.json are reverted locally on my end when executing `npm install`. Otherwise: lgtm!

dergigi:
Ah, damn. I updated npm to latest using `npm update -g` and ran `npm i` again. Is there an easy way to force a specific version via what is in git? I know that ruby has [version files](https://gist.github.com/fnichol/1912050), for example. 

Happy to pin whatever version you think is best.

theborakompanioni:
> Is there an easy way to force a specific version via what is in git?

Guess it is possible with a combination of ["engines" in package.json](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#engines) and a [`.npmrc` file](https://docs.npmjs.com/cli/v8/configuring-npm/npmrc) with [`engines-strict := true`](engine-strict). I'll give it a try and create a PR to discuss this further. This should not delay the release.

> Happy to pin whatever version you think is best.

I have no suggestions on which version to pin, as I do not keep track of npm versions and update on demand most of the time..

Issue #149, opened 2022-03-03
Error: No connection to backend: Gateway Time-out.


bitpaint:
Hey,
Im on a Pie4 with Umbrel using Tor Browser with minimal security settings.
When trying to create a wallet, this happens:

![image](https://user-images.githubusercontent.com/67663265/156582022-ddfb7c8c-53c8-4e1d-9e55-062de6fda951.png)
![image](https://user-images.githubusercontent.com/67663265/156582084-b2b2694f-f525-40ff-9407-7f33e618e56c.png)
![image](https://user-images.githubusercontent.com/67663265/156582368-9ad589d1-cdc4-440c-bbbd-8caef485a758.png)

**Umbrel  logs :**
`jam

jam_1  | Zombie processes will not be re-parented to Tini, so zombie reaping won't work.
jam_1  | To fix the problem, use the -s option or set the environment variable TINI_SUBREAPER to register Tini as a child subreaper, or run Tini as PID 1.
jam_1  | ~/.joinmarket/ssl /src/scripts
jam_1  | Generating a RSA private key
jam_1  | ............................................................................................................................................................++++
jam_1  | .........................++++
jam_1  | writing new private key to 'key.pem'
jam_1  | -----
jam_1  | /src/scripts
jam_1  | Auto start: nginx
jam_1  | Auto start: ob-watcher
jam_1  | Auto start: jmwalletd
jam_1  | Creating wallet jam_default if missing...
jam_1  | Loading wallet jam_default...
jam_1  | /usr/lib/python3/dist-packages/supervisor/options.py:474: UserWarning: Supervisord is running as root and it is searching for its configuration file in default locations (including its current working directory); you probably want to specify a "-c" argument specifying an absolute path to a configuration file for improved security.
jam_1  |   self.warnings.warn(
jam_1  | 2022-03-03 14:05:03,507 CRIT Supervisor is running as root.  Privileges were not dropped because no user is specified in the config file.  If you intend to run as root, you can set user=root in the config file to avoid this message.
jam_1  | 2022-03-03 14:05:03,507 INFO Included extra file "/etc/supervisor/conf.d/jmwalletd.conf" during parsing
jam_1  | 2022-03-03 14:05:03,507 INFO Included extra file "/etc/supervisor/conf.d/nginx.conf" during parsing
jam_1  | 2022-03-03 14:05:03,508 INFO Included extra file "/etc/supervisor/conf.d/ob-watcher.conf" during parsing
[...]

PR #198, opened 2022-03-29
Remove skip button quiz screen

### LINKED ISSUE #196: Remove "Skip" button when on mainnet
If the seed quiz can be skipped, it will be skipped. I would hide the button when Jam is running on mainnet. Skipping the quiz on any testnet is fine, of course.

I already got one message from a user who played around with Jam on his Umbrel node, didn't write down the seed, and wiped his Umbrel node soon after - forgetting that he had funds in JoinMarket.

This is bad, and I would prefer to err on the side of not burning sats, even though such a sacrifice could be seen as a donation to all 😅

theborakompanioni:
Closes #196 

- [x] Show "Skip" button only in dev mode
- [x] Prevent users from leaving the page by mistake, e.g. refreshing the page or closing the tab.
  Unfortunately, for now it wont protect from clicking a link inside the app, ie. the Jam logo
- [x] Minimal tests that verifies "skip" button is not shown when `env := 'test'`

One problem still needs to be addressed: In case the wallet was successfully created, but the backend goes down (for whatever reason), the whole screen is replaced by a "No connection" error message. This should be prevented when `CreateWallet` component is active. It is partly tracked in #149 and covered in #199

### DIFF
--- a/src/components/CreateWallet.jsx
+++ b/src/components/CreateWallet.jsx
@@ -103,9 +103,11 @@ const SeedWordInput = ({ number, targetWord, isValid, setIsValid }) => {
 }
 
 const BackupConfirmation = ({ createdWallet, walletConfirmed, parentStepSetter }) => {
+  const seedphrase = createdWallet.seedphrase.split(' ')
+  const showSkipButton = process.env.NODE_ENV === 'development'
+
   const { t } = useTranslation()
   const [seedBackup, setSeedBackup] = useState(false)
-  let seedphrase = createdWallet.seedphrase.split(' ')
   const [seedWordConfirmations, setSeedWordConfirmations] = useState(new Array(seedphrase.length).fill(false))
 
   useEffect(() => {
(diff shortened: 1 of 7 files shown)


secondl1ght:
I was able to verify point 2, I refreshed the page and tried closing the tab and got a warning message. How do I test point 1? I didn't see a .env file to change the environment variable to `test`, do I have to create one?

theborakompanioni:
> I was able to verify point 2, I refreshed the page and tried closing the tab and got a warning message. 

:muscle: 

> How do I test point 1? I didn't see a .env file to change the environment variable to `test`, do I have to create one?

`NODE_ENV` cannot be overwritten.. I guess the only thing you can do is change the assignment of `devMode` in `App` component to `false` temporarily. There is also a test - when you run the tests (e.g. with `CI=true npm test`) the `NODE_ENV` will be `test`.

secondl1ght:
Sounds good - OK I tested `devMode` being false and it works as expected, I also ran the test suite and all of the tests passed. Please note I did not do a code review on this PR but I did test the UI.

secondl1ght (approved):
Tested the UI and it works as described.

src/components/CreateWallet.jsx
```diff
@@ -8,6 +8,35 @@ import ToggleSwitch from './ToggleSwitch'
 import { serialize, walletDisplayName } from '../utils'
 import { useServiceInfo } from '../context/ServiceInfoContext'
 import * as Api from '../libs/JmWalletApi'
+import './CreateWallet.css'
```
dnlggr, line 11: This is a bit of an anti-pattern I believe. Using separate files to scope CSS is fine but it should be [CSS Modules](https://create-react-app.dev/docs/adding-a-css-modules-stylesheet/) then. CSS Modules let you use the same CSS class name in different files without worrying about naming clashes. Having to worry about unique naming across several separate files can become cumbersome quickly.

theborakompanioni: Ah.. thank you. Really thought this is a good idea when only styles specific to the current screens are placed inside own files (and tightly scoped with `.<screen_name> <class>`). I will stop doing this and revert these changes in upcoming commits. Thanks for your feedback. :orange_heart:

dnlggr: We can definitely start splitting CSS up into separate files -- your point about styles which are specific to single components is totally valid. My only complaint is that if we do it, it should probably be done using CSS Modules. :)

Issue #169, opened 2022-03-09
Turn display seed phrase on settings page into modal


secondl1ght:
There was discussion in PR #160 about turning this into a popup modal, I am creating this issue to track further discussion on this. If it is decided that we want this functionality I can look into implementing it.

editwentyone:
here is a preview for a modal 

https://www.figma.com/proto/kfejZJFlwBywvLEnPEmJo1/JoinMarket-UI?page-id=2850%3A67638&node-id=3283%3A85955&viewport=367%2C48%2C0.57&scaling=min-zoom&starting-point-node-id=3283%3A85955&show-proto-sidebar=1

PR #201, opened 2022-03-31
ui: small figma adaptations

theborakompanioni:
WIP.

- [x] ui: adapt info alert color to match figma more closely
- [ ] Connection indicator symbol
- [ ] Levels screen
- [ ] Seed Modal #169 
- [ ] much more

Feel free to participate!

### DIFF
--- a/src/components/Earn.jsx
+++ b/src/components/Earn.jsx
@@ -256,9 +256,9 @@ export default function Earn() {
           <PageTitle title={t('earn.title')} subtitle={t('earn.subtitle')} />
 
           <rb.Fade in={serviceInfo?.coinjoinInProgress} mountOnEnter={true} unmountOnExit={true}>
-            <div className="mb-4 p-3 border border-1 rounded">
-              <small className="text-secondary">{t('earn.alert_coinjoin_in_progress')}</small>
-            </div>
+            <rb.Alert variant="info" className="mb-4">
+              {t('earn.alert_coinjoin_in_progress')}
+            </rb.Alert>
           </rb.Fade>
 
           {alert && <rb.Alert variant={alert.variant}>{alert.message}</rb.Alert>}
--- a/src/components/Send.jsx
+++ b/src/components/Send.jsx
@@ -490,12 +490,10 @@ export default function Send() {
           <PageTitle title={t('send.title')} subtitle={t('send.subtitle')} />
 
           <rb.Fade in={!isCoinjoinOptionEnabled} mountOnEnter={true} unmountOnExit={true}>
-            <div className="mb-4 p-3 border border-1 rounded">
-              <small className="text-secondary">
-                {serviceInfo?.makerRunning && t('send.text_maker_running')}
-                {serviceInfo?.coinjoinInProgress && t('send.text_coinjoin_already_running')}
-              </small>
-            </div>
+            <rb.Alert variant="info" className="mb-4">
+              {serviceInfo?.makerRunning && t('send.text_maker_running')}
+              {serviceInfo?.coinjoinInProgress && t('send.text_coinjoin_already_running')}
+            </rb.Alert>
           </rb.Fade>
 
           {alert && (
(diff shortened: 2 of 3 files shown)


theborakompanioni:
Button texts in `CurrentWalletMagic` is currently "Deposit" and "Withdraw". In Figma it is "Send" and "Receive".
Personally I lean more towards former, but "Send" and "Receive" is also used in the `Navbar`.

Any suggestions on what is best?
Has this already been discussed and settled with "Send" and "Receive"?
No matter what, it should be consistent.

editwentyone:
Send/Receive is not settled, I just wanted to have it consistent and also added icons. 

Deposit/Withdraw (sounds more formal, like on exchanges) is also valid, but should be used everywhere. Send/Receive (sounds more in everyday language) thus I chose that

dnlggr:
Let's go with the more "lighter" Send/Receive. Like @editwentyone said, "Deposit/Widthdraw" sounds more like an exchange, less than a wallet/tool.

dnlggr (commented):
Not sure what it is, but something is weird when I click on the "Connected" icon (in Firefox). Can you reproduce that? 🤔


https://user-images.githubusercontent.com/10026790/161948137-430f0132-8608-4311-89de-2b0c1f91077f.mov

theborakompanioni:
> Let's go with the more "lighter" Send/Receive. Like @editwentyone said, "Deposit/Widthdraw" sounds more like an exchange, less than a wallet/tool.

Just to explain my thinking: I like the "Deposit"/Widthdraw" terminology because it kind of encourages you to stay in the ecosystem. Will change it to "Send"/"Receive" to stay consistent! Thanks for your inputs!

> Not sure what it is, but some weird blue things show up when I click on the "Connected" icon (in Firefox). Can you reproduce that?

What the.. :rofl: No, just tried that. Cannot reproduce. Must be some weird OS stuff, then? Can you see if this is actually in the DOM?

dnlggr (commented):
Screenshot 2022-04-06 at 14 49 56

After reordering the input fields on the Send screen, the spacing below the amount field when sweeping is a bit too much. This is probably because it used to be the last input field and therefore had more spacing at the bottom to visually separate it a bit from the toggle section. That spacing is not needed anymore I'd say. I'll make the update myself if that's ok with you.

dnlggr:
Hm ok looking at it again, it just looks like it has more spacing because the balance breakdown has a lot of empty space on the left.

Visually I think it looked better with this field as the last input because. Is there a specific UX reason behind reordering the inputs the way they are now? If there is we'll of course need to change the visuals to better fit the new input order but if not I'd just move the amount field down again.

theborakompanioni:
> Visually I think it looked better with this field as the last input because. Is there a specific UX reason behind reordering the inputs the way they are now?

I think this was mentioned in the community call yesterday. If I recall correctly, it was because of some UI changes depending on what is selected in the dropdown. Other than that I make no UI changes on my own, and if I do, I will add a description and a reason why I deviated from it.

dnlggr:
> If I recall correctly, it was because of some UI changes depending on what is selected in the dropdown (in upcoming releases, not included in this PR).

Ok, what could that be? 🤔  I fixed the spacing nonetheless. Hope that is ok.

theborakompanioni:
> > If I recall correctly, it was because of some UI changes depending on what is selected in the dropdown (in upcoming releases, not included in this PR).
> 
> Ok, what could that be?

It was mentioned right at the beginning at ~2m30s: https://bitcointv.com/w/iuTzoVvgW3r317MHDSwy7R?start=2m30s

> I fixed the spacing nonetheless. Hope that is ok.

Of course! Nice! Thanks :rocket:

editwentyone:
hey, thanks for tagging. 

my thoughts for reorganizing the field order is that I think its easier to select a layer or the complete wallet (= tumbler) as the first decision, so afterwards everything loads accordingly.

dnlggr (approved):

Issue #66, opened 2022-02-08
List of little annoyances and other stuff


theborakompanioni:
A small list of things that stood out during testing and trial. 

Please expand it according to your own assessment and let's talk about which are important before the first beta version is released.
Feel free to give an estimated priority as well.

- [x] Inform user about coinjoin state immediately after coinjoin is initiated. Preferably via Websocket ~~Fetch `/session` immediately after starting taker~~ (otherwise user might wait up to 10 seconds to know what is going on) - https://github.com/joinmarket-webui/joinmarket-webui/issues/46
- [x] #83
- [x] #87
- [x] From @AdamISZ : When the maker is started without coins it says "maker cannot start, no coins", but it's important that it should say "no confirmed coins". See https://github.com/joinmarket-webui/joinmarket-webui/issues/34#issuecomment-1032558663 - PR opened https://github.com/JoinMarket-Org/joinmarket-clientserver/pull/1190
- [x] #157
- [x] Improve wording of error message when trying to send to an "invalid" address: "Invalid request format." is not meaningful. (400 InvalidRequestFormat on all assertion errors.. see [wallep_rpc.py#L548](https://github.com/JoinMarket-Org/joinmarket-clientserver/blob/fd5871bb2d89e96efe344040b4576627e84fc644/jmclient/jmclient/wallet_rpc.py#L548))
- [ ] Prevent user from starting taker with insufficient funds (?):  #155 
  - Might be that a taker operation is started and then aborted - either unconfirmed or too little funds available (takes 60s to be resolved)
- [ ] Balance shown on Send page might not be spendable (e.g. utxo is frozen) https://github.com/joinmarket-webui/joinmarket-webui/issues/78

(This list should not contain missing features (e.g. "display spendable funds" or "ability to sweep"), but rather just document odd behaviour or possible improvements for existing ones.)

theborakompanioni:
> > Fetch `/session` immediately after starting taker (otherwise user might wait up to 10 seconds to know what is going on)
> 
> Comment from @dennisreimann: Might be a good idea to bring back the WebSocket connection, it would contain this information.

Yeah, but I also stated it wrong - should be neutral, I did not want to propose a solution already.. will change to "Inform user about coinjoin state immediately after coinjoin is initiated.". Websocket would be great of course! :+1:

dergigi:
[...]

PR #202, opened 2022-04-01
ui: enhance error message on bad request errors

theborakompanioni:
Tiny, tiny PR to get item "Improve wording of error message when trying to send to an 'invalid' address" off the list #66

When sending non-collaborative transactions with an invalid recipient, the api responds with `400 Bad Request` and an error message of `Invalid request format`. 
`InvalidRequestFormat` error is thrown on all assertion errors.. see [wallep_rpc.py#L548](https://github.com/JoinMarket-Org/joinmarket-clientserver/blob/fd5871bb2d89e96efe344040b4576627e84fc644/jmclient/jmclient/wallet_rpc.py#L548), but  `Invalid request format` is not really useful for the user and it might not be clear what to do.
As this is an ubiquitous error message, and address validation is avoided on the client side, to give at least a hint about what might be wrong, the message is appended with "Please verify your inputs carefully." so as to know that the error might be related to the given inputs.

The only data sent when initiating a non-collaborative transactions is `mixdepth`, `destination` and `amount_sats` so this error is most definitely raised on invalid addresses. However, the appended message tries to be neutral and advice the user recheck all inputs.
(If there is a problem with the amount, `409 Conflict` with `Transaction failed` is raised)

### DIFF
--- a/src/components/Send.jsx
+++ b/src/components/Send.jsx
@@ -121,6 +121,15 @@ const CollaboratorsSelector = ({ numCollaborators, setNumCollaborators, minNumCo
   )
 }
 
+const enhanceDirectPaymentErrorMessageIfNecessary = async (httpStatus, errorMessage, onBadRequest) => {
+  const tryEnhanceMessage = httpStatus === 400
+  if (tryEnhanceMessage) {
+    return onBadRequest(errorMessage)
+  }
+
+  return errorMessage
+}
+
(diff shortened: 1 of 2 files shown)


secondl1ght (approved):
Was able to produce the updated error message when inputting a non-valid bitcoin address as the recipient.

image

PR #203, opened 2022-04-02
ui: Add content placeholders

Djo1e:
This PR replaces loaders with content placeholders. By doing this we remove jitter and give the user a visual hint. This is especially useful in the `Send` form, where the user can see what information needs to be entered.

I've opted to use the `Placeholder` component from `react-bootstrap` instead of introducing an additional lib.

**Before:**
https://www.loom.com/share/fc39755d2a1f4f288cb7419470cf08ba

**After:**
https://www.loom.com/share/ac93b1ac92994390a51ef49a5f490b41

**How to test:**
Run locally, make sure all buttons are disabled and that no action can be taken before loading is complete. 

I'm fairly new to the codebase/app so please test carefully. All suggestions are welcome!


src/components/Send.jsx
```diff
@@ -480,33 +480,36 @@ export default function Send() {
...
               </small>
             </div>
-          </rb.Fade>
-
-          {alert && (
-            <rb.Alert className="slashed-zeroes" variant={alert.variant}>
-              {alert.message}
-            </rb.Alert>
+          ) : (
+            <></>
```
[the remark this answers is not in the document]
theborakompanioni: Feels that this could be removed by someone not knowing the impacts in the future.  Do you think the whole content should be wrapped in `<></>` then?

theborakompanioni:
This is a really nice visual improvement. The current loading behaviour always felt clunky. The screens are looking great already. Cannot wait till this is ready for review!

dnlggr:
Awesome thanks for this PR @Djo1e! 🙏 🙌 

Can I mark it as ready for review?

src/components/Send.jsx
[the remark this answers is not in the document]
Djo1e: Good point. Changed!

src/components/CurrentWalletAdvanced.tsx
```diff
@@ -68,45 +69,35 @@ export default function CurrentWalletAdvanced() {
...
-        <rb.Row className="justify-content-center">
-          <rb.Col className="flex-grow-0">
-            <div className="d-flex mb-3">
-              <rb.Spinner as="span" animation="border" size="sm" role="status" aria-hidden="true" className="me-2" />
-              {t('current_wallet.text_loading')}
-            </div>
-          </rb.Col>
-        </rb.Row>
-      )}
+      {isLoading && <LoadingAccounts amount={5} />}
```
dnlggr, line 72: Are you planning on reusing `LoadingAccounts` anywhere else? If not we could think about inlining it here similar to the placeholders in the other components.

src/components/LoadingAccounts.css
```diff
@@ -0,0 +1,4 @@
+.placeholder-loading-account {
+  height: 3.5rem;
+  margin-bottom: 0.375rem;
+}
```
dnlggr, line 4: Let's use [CSS Modules](https://create-react-app.dev/docs/adding-a-css-modules-stylesheet/) when putting component CSS into separate files.

src/components/LoadingAccounts.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+interface LoadingAccountsProps {
+  amount?: number
+}
+
+export function LoadingAccounts({ amount = 1 }: LoadingAccountsProps) {
+  return (
+    <div>
+      {[...new Array(amount)].map((a, index) => (
+        <rb.Placeholder key={index} as="div" animation="glow" style={{ marginBottom: 6 }}>
```
dnlggr, line 11:
```suggestion
        <rb.Placeholder key={index} as="div" animation="wave" style={{ marginBottom: 6 }}>
```

Let's be consistent in the animation we use. I don't have a strong preference over "wave" or "glow" but whatever we pick, let's make it consistent across the app.

dnlggr, line 11:
```suggestion
        <rb.Placeholder key={index} as="div" animation="glow">
```

Let's use the CSS from the CSS module here.

src/components/Send.css
```diff
@@ -122,3 +122,15 @@ input::-webkit-inner-spin-button {
...
   border-color: var(--bs-gray-dark) !important;
 }
+
+.input-loader {
+  height: 3.5rem;
+}
+
+.text-loader {
+  height: 1.75rem;
+}
```
dnlggr, line 132:
```suggestion
```

This is not used anywhere I believe.

src/components/Send.css
```diff
@@ -122,3 +122,15 @@ input::-webkit-inner-spin-button {
   background-color: var(--bs-gray-dark);
   border-color: var(--bs-gray-dark) !important;
 }
+
+.input-loader {
+  height: 3.5rem;
+}
```
dnlggr, line 128:
```suggestion
.send .input-loader {
  height: 3.5rem;
}
```

Since this file is not a CSS Module, we have to be careful about naming clashes.

dnlggr (commented):
Thanks for this PR @Djo1e! Great job, it's looking really good already. Great visual improvement compared to before. 🎉 👏 

I left some comments inline. Let me know what you think or if you need further clarification on something.

On the receive screen,  let's also disable the "Settings" button while the content is loading:

Screenshot 2022-04-06 at 11 16 28

---

Final thing: The `CurrentWalletMagic` screen could use some placeholders as well. This is the first screen users will see first. Would be great if you could you take a look at that as well. 🙏

Djo1e:
@dnlggr Thanks for the review! I have already addressed some of them, but will need to clean it up a little bit. I'll probably finish this over the weekend when I have more time. I added the content placeholder for the `CurrrentWalletMagic` screen, but since we won't know how many wallets are there until loading is finished, I'm not sure how many placeholders to add. I tried something, but it does't look that good if there are less wallets than the loaders. See https://www.loom.com/share/4eb1f4573e4b476491a4820561080dd2

dnlggr:
Thanks for the update @Djo1e! No worries take your time.

For the `CurrentWalletMagic` I actually meant this screen:

Screenshot 2022-04-07 at 12 31 22

You can toggle it to be the default wallet screen in settings:

Screenshot 2022-04-07 at 12 31 30

editwentyone:
I like it so far :) is there a chance to add subtle animated dots like "…" per field? 
maybe its a bit much, but I would show the user more clear that something is "on the way"

dnlggr:
> I like it so far :) is there a chance to add subtle animated dots like "…" per field?
maybe its a bit much, but I would show the user more clear that something is "on the way"

The placeholders have a built in loading animation. It's just not very well visible with the current color choice. Maybe we can find a better color for the placeholders @Djo1e?

Example with a (too) dark color just to show the animation:

https://user-images.githubusercontent.com/10026790/162457038-f6b94540-089d-4e02-aae4-0ecd5338380f.mov

editwentyone:
keep it as it is, I would love to see it in action on my screen and then we could tweak it further if needed. also lets see what other feedback will come back

src/components/CurrentWalletMagic.jsx
```diff
@@ -8,46 +8,86 @@ import Balance from './Balance'
...
+    </div>
+  )
+}
+
+const PrivacyLevel = ({ numAccounts, level, balance, loading }) => {
   const settings = useSettings()
 
   const filledShields = Array(level + 1)
-    .fill()
+    .fill('')
```
Djo1e, line 81: Typescript was complaining

src/components/ExtendedLink.tsx
```diff
@@ -0,0 +1,19 @@
+import { Link, LinkProps } from 'react-router-dom'
+
+interface Props extends LinkProps {
+  disabled?: boolean
+}
+
+export function ExtendedLink({ disabled, ...props }: Props) {
```
Djo1e, line 7: I've extended the `Link` component since it doesn't have a disabled state internally

src/components/CurrentWalletAdvanced.tsx
Djo1e: Changed

src/components/LoadingAccounts.tsx
Djo1e: Good catch, changed.

src/components/Send.css
Djo1e: Removed

src/components/Send.css
Djo1e: Done. I've moved all css that isn't global into an appropriate module.

dnlggr (dismissed):
Thanks for the update @Djo1e! 🎉 

Visually looks great now! Please fix the `undefined` in the Balance component, though. After that we should be good to merge. 👍

src/components/Balance.jsx
```diff
@@ -66,14 +76,14 @@ const BalanceComponent = ({ symbol, value, symbolIsPrefix }) => {
  * @param {showBalance}: A flag indicating whether to render or hide the balance.
  * Hidden balances are masked with `*****`.
  */
-export default function Balance({ valueString, convertToUnit, showBalance = false }) {
+export default function Balance({ valueString, convertToUnit, showBalance = false, loading }) {
```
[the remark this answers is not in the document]
Djo1e: Done!

src/components/Balance.jsx
```diff
@@ -83,13 +93,14 @@ export default function Balance({ valueString, convertToUnit, showBalance = fals
...
+        loading={false}
       />
     )
   }
 
   if (typeof valueString !== 'string') {
-    console.warn('<Balance /> component expects string input')
-    return <BalanceComponent symbol={''} value={valueString} symbolIsPrefix={false} />
+    if (!loading) console.warn('<Balance /> component expects string input')
+    return <BalanceComponent symbol={undefined} value={undefined} symbolIsPrefix={false} loading={loading} />
```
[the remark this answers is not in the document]
Djo1e: Fixed

src/components/CurrentWalletMagic.jsx
```diff
@@ -8,46 +8,86 @@ import Balance from './Balance'
...
+            />
+          ))}
+      </div>
+    </div>
+  )
+}
+
+const LoadingPrivacyLevel = ({ level }) => {
+  const loadingShields = Array(level)
+    .fill()
```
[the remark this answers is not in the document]
Djo1e: It should - added

src/components/ExtendedLink.tsx
```diff
@@ -0,0 +1,19 @@
...
+  /* eslint-disable @typescript-eslint/no-unused-vars */
+  const { reloadDocument, replace, state, to } = props
```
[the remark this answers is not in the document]
Djo1e: removed

theborakompanioni:
Two things that were noticed:
- Sprites "node", "receive", "send" were not adopted
- flickering when loading a new address on "Receive" screen

I guess these are rather minor changes, so we could also fix them later.
Waiting for an answer as you might want to fix them yourself.
@dnlggr What do you think, are we ready to merge?

Thank you @Djo1e :pray:

Djo1e:
@theborakompanioni I've added the missing sprites and fixed the flickering. Thanks for taking the time to review (@dnlggr as well!)

theborakompanioni (approved):
I guess some really tiny stuff still needs to be adjusted, e.g. "send" and "receive" sprites in the buttons in `CurrentWalletMagic` screen are missing - but we can address that later.

lgtm!

dnlggr:
Thanks @theborakompanioni and @Djo1e. Yeah I agree, we can address little things separately. 

Thanks for the contribution @Djo1e and thanks for taking over the review @theborakompanioni! 🎉

PR #206, opened 2022-04-04
ui: move seed phrase on settings page to modal

### LINKED ISSUE #169: Turn display seed phrase on settings page into modal

theborakompanioni:
Closes #169. Show the seed phrase on settings page in a modal [based on screens](https://www.figma.com/proto/kfejZJFlwBywvLEnPEmJo1/JoinMarket-UI?page-id=2850%3A67638&node-id=3283%3A85955&viewport=367%2C48%2C0.57&scaling=min-zoom&starting-point-node-id=3283%3A85955&show-proto-sidebar=1) by @editwentyone.

### DIFF
--- a/src/components/Seedphrase.jsx
+++ b/src/components/Seedphrase.jsx
@@ -1,9 +1,13 @@
 import React from 'react'
 import './Seedphrase.css'
 
-export default function Seedphrase({ seedphrase, isBlurred = true }) {
+export default function Seedphrase({ seedphrase, isBlurred = true, centered = false }) {
   return (
-    <div className="seedphrase slashed-zeroes d-flex flex-wrap">
+    <div
+      className={`seedphrase slashed-zeroes d-flex flex-wrap ${
+        centered ? 'justify-content-center align-items-center' : ''
+      }`}
+    >
       {seedphrase.split(' ').map((seedWord, index) => (
         <div key={index} className="d-flex py-2 ps-2 pe-3">
           <span className="seedword-index text-secondary text-end">{index + 1}</span>
--- a/src/components/Settings.jsx
+++ b/src/components/Settings.jsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react'
+import React, { useEffect, useState } from 'react'
 import * as rb from 'react-bootstrap'
 import { useTranslation } from 'react-i18next'
 import Sprite from './Sprite'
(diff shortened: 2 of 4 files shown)


src/components/Seedphrase.jsx
```diff
@@ -1,9 +1,13 @@
 import React from 'react'
 import './Seedphrase.css'
 
-export default function Seedphrase({ seedphrase, isBlurred = true }) {
+export default function Seedphrase({ seedphrase, isBlurred = true, centered = false }) {
   return (
-    <div className="seedphrase slashed-zeroes d-flex flex-wrap">
+    <div
+      className={`seedphrase slashed-zeroes d-flex flex-wrap ${
```
Djo1e, line 7: We should consider using [classnames](https://www.npmjs.com/package/classnames) for this. It's very lightweight and makes conditionally joining classNames a bit easier.

Eg, this could be written as
```
<div
  className={classNames('seedphrase slashed-zeroes d-flex flex-wrap', {
    'justify-content-center align-items-center': centered,
  })}
>
```

src/components/Settings.jsx
```diff
@@ -129,48 +192,15 @@ export default function Settings() {
         <rb.Button
           variant="outline-dark"
           className="border-0 mb-2 d-inline-flex align-items-center"
-          onClick={async (e) => {
+          onClick={(e) => {
             e.preventDefault()
```
Djo1e, line 196: NIT: I think `e.preventDefault()` can be removed. It doesn't really do anything in this case. I've tested it locally.

Djo1e (approved):
Looks and works great! 👍

secondl1ght:
Tested the functionality on this - looks and works great! Nice work, one minor thing is that I don't know if having the wallet name wrapped in arrows <> looks good. Maybe its because I am a developer so it looks like code to me? A regular user might think it looks fine? Just a comment, but yeah really a nice improvement from the old version. 🙌🏼

secondl1ght (approved):
Nice!

theborakompanioni:
> [...] one minor thing is that I don't know if having the wallet name wrapped in arrows <> looks good. 

Thanks for reviewing :pray:.
You are right. This is coming from the Figma screens and to be honest, I am not sure myself if this should be included or if it just indicates a placeholder..

Ping @editwentyone

dnlggr:
Let's remove the angle brackets!

src/components/Settings.jsx
```diff
@@ -7,17 +7,80 @@ import Seedphrase from './Seedphrase'
...
+
+    if (show) {
+      loadSeed()
+    }
+  }, [show, currentWallet])
+
+  return (
+    <rb.Modal size="lg" show={show} onHide={onHide} keyboard={false} centered={true} animation={true}>
+      <rb.Modal.Header closeButton>
+        <rb.Modal.Title>&lt;{walletDisplayName(currentWallet.name)}&gt;</rb.Modal.Title>
```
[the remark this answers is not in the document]
dnlggr: Had this written yesterday and didn't submit the review. 😭 🤦

dnlggr (commented):
Looks good!

One thing I'm not so certain about is the `centered` prop. Ideally, the seedphrase would be an edge-to-edge component without any internal margins and the parent component would be responsible for placing it in the center or on the left or where ever it wants it to be. Do you know what I mean? It's a nit so you don't have to fix it in this PR, just noting it down.

~On mobile the centering doesn't really seem to work which is another nit though:~ Edit: Ignore, that's not true.

theborakompanioni:
> Do you know what I mean? It's a nit so you don't have to fix it in this PR, just noting it down.

Yes, I know what you mean. I am unsure how to do that properly. Can you guide me in the right direction?

dnlggr:
> Yes, I know what you mean. I am unsure how to do that properly. Can you guide me in the right direction?

Not sure as well, I'll have a look. Not a change request for this PR though. We can merge it as is.

theborakompanioni:
> Not sure as well, I'll have a look. Not a change request for this PR though. We can merge it as is.

I'll wait for your inputs. It does not hurry.

dnlggr:
> I'll wait for your inputs. It does not hurry.

Couldn't really come up with a good solution straight away. Let's keep it like it is for now. :)

PR #207, opened 2022-04-04
refactor: Replace react-bootstrap component imports

Djo1e:
Read issue https://github.com/joinmarket-webui/joinmarket-webui/issues/205 for more information.

This PR changes all the components from `react-bootstrap` to be imported as recommended by the [react-bootstrap docs](https://react-bootstrap.github.io/getting-started/introduction#importing-components)

`import * as rb from 'react-bootstrap` => `import <component> from 'react-bootstrap/<component>`

### DIFF
--- a/src/components/Earn.jsx
+++ b/src/components/Earn.jsx
@@ -15,7 +15,7 @@ import Col from 'react-bootstrap/Col'
 import Spinner from 'react-bootstrap/Spinner'
 import Fade from 'react-bootstrap/Fade'
 import Button from 'react-bootstrap/Button'
-import Alert from './Alert'
+import Alert from 'react-bootstrap/Alert'
 
 const OFFERTYPE_REL = 'sw0reloffer'
 const OFFERTYPE_ABS = 'sw0absoffer'


src/components/Alert.jsx
```diff
@@ -1,5 +1,5 @@
 import React, { useState } from 'react'
-import { Alert as BsAlert } from 'react-bootstrap'
+import BsAlert from 'react-bootstrap/Alert'
```
theborakompanioni, line 2: This does not seem quite right. `Alert` is imported as `BsAlert` as our own component is named `Alert`.

src/components/Earn.jsx
```diff
@@ -31,10 +38,10 @@ const YieldgenReport = ({ lines, maxAmountOfRows = 25 }) => {
 
   return (
     <div className="mt-2 mb-3">
-      {empty && <rb.Alert variant="info">{t('earn.alert_empty_report')}</rb.Alert>}
+      {empty && <Alert variant="info">{t('earn.alert_empty_report')}</Alert>}
```
theborakompanioni, line 41: As far as I can see our own `Alert` component does not support `children`, but a `message` property. Does this work on your end?

Djo1e: Nice catch! This is supposed to be a react-bootstrap component, I accidentally imported the local one. Fixed!

src/components/Alert.jsx
Djo1e: Should I name it differently? Maybe `RBAlert`?

Djo1e: No, not at all. I'm fine with that. It's also fine if we don't merge this at all :)

dnlggr: I see you point @theborakompanioni, `rb.*` definitely increases readability. But on the other hand if this is the recommended approach from the library authors, I'd trust that and go with the explicit component imports. We just need to be careful not to confuse our own components with the ones from Bootstrap. No super strong feelings towards either approach though, just a feeling.

theborakompanioni: Having thought it through again, I am in favor of not merging. And after the thorough analysis of @Djo1e, I see even less reasons to do so. Also, no super strong feelings. Just to let you know.

dnlggr: Ok then let's not merge. Ok with you @Djo1e?

Djo1e: Yes, closing this PR

PR #210, opened 2022-04-05
feat: Confirm password on Create Wallet screen

### LINKED ISSUE #208: Wallet Password Confirmation / Change ?
Just noticed when creating the wallet there is not password confirmation, its shown on the next screen but that is too late.  Also there is no way to change a wallet password.

Umbrel version 0.4.16

theborakompanioni:
Closes #208. This PR adds password confirmation functionality to the Create Wallet screen.

The "pattern" property of the input has been used to circumvent some strange validation behaviour of react-bootstrap.
If there is any chance someone comes up with a good alternative solution, while still using the inbuilt validation feedback elements, please help (but please try your approach before suggesting it; many attempts that come to mind have been tried and failed)! Downside of this approach is: 
- Potentially error prone regex escaping is used
- Password is "visible" in the html (think this is a non-issue, although not pretty, passwords can be gathered anyways with access to the DOM)


MaxHillebrand (approved):
cACK

dnlggr:
> circumvent some strange validation behaviour of react-bootstrap

What was the strange validation behavior? 🤔

dnlggr:
> Edit: To be fair, react-bootstrap [hints at using an external library](https://react-bootstrap.github.io/forms/validation/#form-libraries-and-server-rendered-styles) for form validation. Refrained from doing that for this PR.

Yeah I think that's what we should eventually do. I also had to do some hacky stuff on the send page because form validation was weird. Would probably be better to eventually use a library if it suits our needs. Maybe let's try that out next time we need to do form validation and bootstrap behaving weird.

theborakompanioni:
> Maybe let's try that out next time we need to do form validation and bootstrap behaving weird.

What about including it with this PR? I have some troubles with the current approach myself..
Should we go with [Formik](https://github.com/jaredpalmer/formik) as suggested?

dnlggr:
> What about including it with this PR? I have some troubles with the current approach myself..

If you're motivated to do it now, sure!

Chose whatever you like, Formik is recommended by Boostrap and has 30k stars so it's probably trustworthy and stable. 👍

theborakompanioni:
Formik integrated. Unfortunately, there seems to be a bug https://github.com/jaredpalmer/formik/issues/1543 that makes the test look a bit strange now. But the actual implementation is much cleaner. Needs a careful review!

dnlggr (approved):
Nice! Tested and works as expected. Form lib looks clean. Looking forward to using it elsewhere in the app (e.g. my ugly `Send` page 😄)

PR #212, opened 2022-04-07
prepare scheduled transactions on regtest

theborakompanioni:


theborakompanioni:
Wow.. that was fast.. @MaxHillebrand is clearly a bot : D
Please do not expect too much or anything at all. Just testing and figuring stuff out in the open.

MaxHillebrand:
I must confess, I have a serious github notification addiction...

PR #215, opened 2022-04-09
refactor: Add route constants

Djo1e:
This PR adds route constants, by doing this we will:
- Prevent typos when specifying a route
- Make it easier for new devs to get familiarised with all current routes/pages


theborakompanioni (approved):
Nice. Thank you :orange_heart:

PR #218, opened 2022-04-14
fix: prevent operations when maker/taker service is running

theborakompanioni:
This PR _tries_ to prevent  operations that could negatively  impact running maker or taker services.
See comment https://github.com/JoinMarket-Org/joinmarket-clientserver/pull/1242#issuecomment-1099423159 by @AdamISZ which implies that it can have unforeseen side-effects.

Currently the following actions are prevented when a maker or a taker service is running:
- Sending a transaction (till now "non-collaborative" transaction were possible even when services were running)
- freeze/unfreeze actions

Additionally (has nothing to do with running maker/taker):
- prevent unfreezing of fidelity bonds before locktime is over

This is done by disabling some forms or buttons - as you know, this does not prevent anybody from sending API requests directly. It should just mitigate the problem for everyday users.


AdamISZ:
I presume you'd agree we should also disable it in the backend?

src/components/DisplayUTXOs.jsx
```diff
@@ -71,28 +80,30 @@ const Utxo = ({ utxo, ...props }) => {
...
-                    ? t('current_wallet_advanced.button_unfreeze')
-                    : t('current_wallet_advanced.button_freeze')}
-                </rb.Button>
-              </div>
+              {showFreezeActionButton && (
+                <div>
+                  <rb.Button
+                    size="sm"
+                    variant={utxo.frozen ? 'outline-warning' : 'outline-info'}
+                    disabled={isSending || !isOperationEnabled()}
```
dnlggr, line 88: The disabled state for this button doesn't really look like a disabled button imo:

### Enabled:
Screenshot 2022-04-15 at 16 17 16

### Disabled:
Screenshot 2022-04-15 at 16 17 24

Maybe we just hide the button? Since there's no explanation on why freezing is disabled, we could just remove the buttons altogether as long as they cannot be clicked. What d you think?

dnlggr (commented):
Nice improvement! 👏  Let me know what you think of my comments. 🙏 

---

I feel like it would be useful if this info text would be a button that leads to the "Earn" page (or if there'd be a link in the text). Some way for the user to instantly go to the place where he needs to take action to make sending possible again.

Screenshot 2022-04-15 at 16 14 02

src/components/DisplayUTXOs.jsx
```diff
@@ -1,24 +1,29 @@
...
 const Utxo = ({ utxo, ...props }) => {
   const { t } = useTranslation()
   const settings = useSettings()
   const currentWallet = useCurrentWallet()
+  const serviceInfo = useServiceInfo()
 
   const [alert, setAlert] = useState(null)
   const [isSending, setIsSending] = useState(false)
 
+  const isOperationEnabled = useCallback(() => {
```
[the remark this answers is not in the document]
theborakompanioni: Personally, I try to always use a "positive" wording for boolean flags to be consistent, but this is just a personal style and should not block if you think it'd increase readability.

src/components/Send.jsx
```diff
@@ -61,7 +62,7 @@ const CollaboratorsSelector = ({ numCollaborators, setNumCollaborators, minNumCo
   }
 
   return (
-    <rb.Form noValidate className="collaborators-selector">
+    <rb.Form noValidate className="collaborators-selector" disabled={!isEnabled}>
```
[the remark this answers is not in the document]
theborakompanioni: In this case, I think "disabled" would actually make it better, as it is also a property of some `HTMLElement`s. Will be done! :muscle:

src/components/DisplayUTXOs.jsx
theborakompanioni: Yes, very valid. Will do!

theborakompanioni:
> I presume you'd agree we should also disable it in the backend?

Yes, that feels perfectly reasonable.

src/components/DisplayUTXOs.jsx
theborakompanioni: Done!

src/components/Send.jsx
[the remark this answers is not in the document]
theborakompanioni: Done.

theborakompanioni:
> I feel like it would be useful if this info text would be a button that leads to the "Earn" page (or if there'd be a link in the text). 

:+1: Done! "Earn" is now a link to the Earn screen. I feel like we could also make a "Stop" button here as this is what the user might do on the Earn screen anyway.. what about that?

src/components/DisplayUTXOs.jsx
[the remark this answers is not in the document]
theborakompanioni: Please let me know if this is okay for you @dnlggr . 
It's is totally fine to change it, but I think it is the second time you mentioned it.. and I'd need to do a better job at adjusting my mental model and be careful not to let this creep into future changes as well.

theborakompanioni:
Blocked by #222 .

editwentyone:
I added the Earn Link and also a Chevron to the right: https://www.figma.com/file/kfejZJFlwBywvLEnPEmJo1/?node-id=2850%3A67638 both will link to the earn page. 

I wouldn't add another functionality like direct stop, the information should be more than enough at the beginning for the user.

MaxHillebrand:
Is there a way to have a "graceful fallback", so when it's not possible to make a payment, instead of failing, stop maker, make payment, start maker.

theborakompanioni:
> Is there a way to have a "graceful fallback", so when it's not possible to make a payment, instead of failing, stop maker, make payment, start maker.

I think I understand what you are getting at. Ideally a user has the maker running at all times without too much manual intervention.. would you open a ticket so it can be addressed thoroughly? Once feature parity has been reached, these are ideas that can be focused on. :muscle:

src/components/DisplayUTXOs.jsx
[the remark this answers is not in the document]
theborakompanioni: I see your point now - it is definitely much clearer and more obvious when using `disabled`. Will adapt - code and behaviour.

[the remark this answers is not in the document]
theborakompanioni: Done.

theborakompanioni:
Updated! Latest changes include reloading the session information, as it should be decided as soon as possible whether the operation is enabled at all. This request does not have much overhead and the response should arrive quite fast.

src/components/Send.jsx
```diff
@@ -491,20 +499,31 @@ export default function Send() {
...
-                  {serviceInfo?.coinjoinInProgress && t('send.text_coinjoin_already_running')}
-                </small>
-              </div>
-            ) : null}
-          </>
+        <rb.Fade in={isOperationDisabled()} mountOnEnter={true} unmountOnExit={true}>
+          <rb.Alert variant="info" className="mb-4">
+            <>
+              {serviceInfo?.makerRunning && (
+                <rb.Row>
```
dnlggr, line 512:
```suggestion
                <rb.Row className="align-items-center">
```

To center the caret vertically:

Screenshot 2022-04-20 at 16 04 37

src/components/Send.jsx
```diff
@@ -491,20 +499,31 @@ export default function Send() {
...
+                  <rb.Col>
+                    <Trans i18nKey="send.text_maker_running">
+                      <Link to={routes.earn}>Earn</Link> is active. Stop the service in order to send collaborative
+                      transactions.
+                    </Trans>
+                  </rb.Col>
+                  <rb.Col xs="auto">
+                    <Sprite symbol="caret-right" width="24px" height="24px" />
```
dnlggr, line 520: I feel like with the caret, the whole thing should be clickable not just "Earn" in the text. What do you think?

src/components/Send.css
```diff
@@ -25,6 +25,21 @@ input::-webkit-inner-spin-button {
...
 
+.send.maker-running form,
+.send.maker-running .collaborators-selector,
+.send.maker-running button[type='submit'] {
+  filter: blur(2px);
+}
+
+/* figma design would like to have the values shown when a taker operation
+   is running - since we do not support this by now, lets blur the
+   forms in the meantime, just like when a maker is running. */
```
dnlggr, line 36:
```suggestion
```

I'd leave that out. Not really relevant to document here imo.

dnlggr (commented):
Looks good!

A couple of comments inline, nothing tragic.

What's up with the Sprite? 😅 

Screenshot 2022-04-20 at 16 08 08

src/components/Send.css
theborakompanioni: Removed.

src/components/Send.jsx
theborakompanioni: Done!

src/components/Send.jsx
theborakompanioni: Definitely!

Hmm.. trickier than expected.. I could not prevent browsers from underlining the text inside the alert with `a .alert { text-decoration: none }`.. That's why it's now done with an `.unstyled` class as workaround. : /

Maybe you have a cleaner solution at hand - that'd be great.

theborakompanioni:
> Looks good!
> A couple of comments inline, nothing tragic.

Thanks for your feedback @dnlggr . :pray: 

> What's up with the Sprite?

Was wondering that too!
Resisted to investigate successfully : D

dnlggr (approved):

src/components/Send.jsx
dnlggr: Hm yeah I also couldn't get it to work right away. `unstyled` is fine imo.

PR #220, opened 2022-04-15
fix: remove hardcoded account array

theborakompanioni:
This PR removes the hardcoded account array that has been used in the Receive screen.

Instead of the hardcoded array, the information stored in `walletInfo` is used.
The data is deliberately not reloaded for now, as the info is fetched at least once outside the component and account creation is a rather rare event. Also, the main function of the component - fetching and displaying an address - can be used without this information as well. Additionally, it is only used in a dropdown in "advanced mode", since Jam encourages to deposit to account `0` only.


src/components/Receive.jsx
```diff
@@ -16,41 +15,46 @@ export default function Receive() {
...
-        .then((data) => setAddress(data.address))
-        .catch((err) => {
-          !abortCtrl.signal.aborted && setAlert({ variant: 'danger', message: err.message })
-        })
-        .finally(() => !abortCtrl.signal.aborted && setIsLoading(false))
-    }
-
-    if (ACCOUNTS.includes(account)) {
-      fetchAddress(account)
-    }
```
dnlggr, line 49: It's ok to get rid of that check since `accounts` can't be set manually anyway (it can just be selected from a pre-filled dropdown), right?

theborakompanioni: Yes. And even if it could be set manually, it would only lead to an error being displayed, which feels better than just ignoring the request.

dnlggr: Sounds good to me! ✅

dnlggr (commented):
LGTM!

dnlggr (approved):

Issue #31, opened 2022-01-29
Fidelity Bonds


dergigi:
Add support to lock up funds in a [fidelity bond](https://github.com/JoinMarket-Org/joinmarket-clientserver/blob/master/docs/fidelity-bonds.md).

This is somewhat of a placeholder issue for now, since we need to update the API first afaik.

AdamISZ:
Was just looking at the Issue list here and noticed this. The API exposes an endpoint to get a timelocked destination, and then you can send funds to it. Maker automatically uses the largest FB utxo that exists in the wallet. Clearly it would need some UI but, given this, I don't think API blocks here, unless I forgot something.

dergigi:
Wonderful! Thanks for the clarification, Adam!

MaxHillebrand:
> Update the wallet page to show FB balance (with locked/unlocked status)

add when the FB unlocks/expires.

also, how about having a "auto renew" function, so that when the FB expires, the wallet automatically opens a new one with the same parameters.

theborakompanioni:
> > Update the wallet page to show FB balance (with locked/unlocked status)
> 
> add when the FB unlocks/expires.

:+1:
> also, how about having a "auto renew" function, so that when the FB expires, the wallet automatically opens a new one with the same parameters.

Guess making this not only possible but easier is a great improvement. Little skeptical of doing something like locking coins automatically.
As a first step, what about presenting a big button to create a Fidelity Bond (if none exists) when the user wants to start a maker, which must be manually confirmed?

dnlggr:
There are some design explorations for fidelity bonds [in Figma](https://www.figma.com/file/kfejZJFlwBywvLEnPEmJo1/JoinMarket-UI?node-id=2771%3A66487) already. Posting them here for reference.

dergigi:
Moved to a later version. Not saying we shouldn't work on this, but as @openoms correctly pointed out in Call 11, we should prioritize #179 so users can create fidelity bonds in a private manner. We don't want users to create a bond with KYC'd UTXOs, announcing to the world what they are doing.

theborakompanioni:
> I'm interested in adding this feature since I need it. 

Hey @abhiShandy , have you already started working on this?
I have given it some thought and started a small prototype.

> Do you have a workflow in mind? I can think of the following checklist:
> 
>     * [ ]  show an optional FB option on Earn page
> 
>     * [ ]  generate a QR code with a desired year and month
> 
[...]

PR #221, opened 2022-04-15
feat: Fidelity bonds in developer mode

theborakompanioni:
Part of #31.

This PR contains a minimalist screen to create Fidelity Bonds **in advanced mode**.
It is behind a feature flag that is only activated **for local development**.

Fidelity Bonds for **normal mode** is a bit comprehensive and is done in follow-up PR.

Currently, a link from the `CurrentWalletAdvanced` view leads to a simple form that let's users choose the locktime and will subsequently display the corresponding address. As this is in **advanced mode**, a user can then **manually** fund this address by himself.

In the **normal mode**, the fidelity bond should be created **automatically**.

Early state. This is just a proof of concept. All feedback very welcome.
Keep in mind that it is still hidden behind a feature flag!

https://user-images.githubusercontent.com/3358649/171407524-d33b30e5-ffe5-4c4d-856c-ede69d2f6b5b.mp4

It would be great to have this merged for local development, even if it is not fully polished yet (e.g. some translation strings are missing).


theborakompanioni:
Updated:
- fixed dark mode
- remove "advanced view": Simple (not implemented) and DevOnly view a two distinct components

Review required. Everything said before is still true:
- Not polished but would be good to have it in yet
- It is hidden behind a feature flag only available in local development
- Missing translation strings and other stuff will be done in follow-up PRs (e.g. #307)

Needs a review.

src/components/FidelityBond.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+          <rb.Alert variant="danger" className="mb-4">
+            <Trans i18nKey="fidelity_bond.alert_warning_dev_mode">
+              Fidelity Bonds are currently only available in developer mode.
+              {featureAdvancedEnabled && (
+                <>
+                  {' '}
+                  <Link className="unstyled" to={routes.fidelityBondsDevOnly}>
+                    Switch to developer view.
+                  </Link>
```
dnlggr, line 55: Honestly I don't think we need that extra step. It's already only in "advanced" mode (which is called developer mode in the app already). Once the user is here I'd say we can let them do their stuff. No need for the extra hurdle imo. What do you think about it?

src/components/FidelityBondDevOnly.tsx
```diff
@@ -0,0 +1,350 @@
...
+      return 1
+    }
+
+    return currentMonth + 1 // can be '13' - which means it never is valid and user must adapt 'year'.
+  }
+
+  return (
+    <rb.Row>
+      <rb.Col xs={6}>
+        <rb.Form.Group className="mb-4" controlId="locktimeYear">
```
dnlggr, line 72: It's a shame that Firefox still [doesn't support](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/month#browser_compatibility) the "month" date picker. It's been tracked for over 9 years... Same for Safari.

Anyway what do you think of a dropdown here? At least for the months a 12 value dropdown is something users will know from most websites. The same applies to the year as well I guess. Most users won't pick a date several decades ahead (I guess) so the year will be accessible quickly in the dropdown.

Longer term, we should go with a nice date picker of course but this can be an improvement. Shame on you Firefox and Safari.

src/components/FidelityBondDevOnly.tsx
```diff
@@ -0,0 +1,350 @@
...
+import styles from './FidelityBond.module.css'
+
+type AlertWithMessage = rb.AlertProps & { message: string }
+
+const dateToLockdate = (date: Date): Api.Lockdate =>
+  `${date.getUTCFullYear()}-${date.getUTCMonth() >= 9 ? '' : '0'}${1 + date.getUTCMonth()}` as Api.Lockdate
+
+const lockdateToTimestamp = (lockdate: Api.Lockdate): number => {
+  const split = lockdate.split('-')
+  return Date.UTC(parseInt(split[0], 10), parseInt(split[1], 10) - 1, 1, 0, 0, 0)
```
dnlggr, line 25:
```suggestion
  return Date.UTC(parseInt(split[0], 10), parseInt(split[1], 10) - 1, 1)
```

The time parameters are 0 [per default](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/UTC#syntax). the day is 1 per default as well but I'd say that's fine to leave in for purposes of documentation.

src/components/FidelityBondDevOnly.tsx
```diff
@@ -0,0 +1,350 @@
...
+import { routes } from '../constants/routes'
+import * as Api from '../libs/JmWalletApi'
+import styles from './FidelityBond.module.css'
+
+type AlertWithMessage = rb.AlertProps & { message: string }
+
+const dateToLockdate = (date: Date): Api.Lockdate =>
+  `${date.getUTCFullYear()}-${date.getUTCMonth() >= 9 ? '' : '0'}${1 + date.getUTCMonth()}` as Api.Lockdate
+
+const lockdateToTimestamp = (lockdate: Api.Lockdate): number => {
```
dnlggr, line 23: These two would be ideal for some unit tests.

src/components/FidelityBondDevOnly.tsx
```diff
@@ -0,0 +1,350 @@
...
+}
+const LockdateForm = ({ onChange, maxYears = DEFAULT_MAX_TIMELOCK_YEARS }: LockdateFormProps) => {
+  const { t } = useTranslation()
+
+  const now = new Date()
+  const currentYear = now.getUTCFullYear()
+  const currentMonth = 1 + now.getUTCMonth() // utc month ranges from [0, 11]
+
+  const [lockdateYear, setLockdateYear] = useState(currentYear + 1)
+  const [lockdateMonth, setLockdateMonth] = useState(currentMonth)
```
dnlggr, line 53: This duplicates the logic from `initialLockdate` above.

src/components/FidelityBondDevOnly.tsx
```diff
@@ -0,0 +1,350 @@
...
+
+  useEffect(() => {
+    const date = new Date(Date.UTC(lockdateYear, lockdateMonth - 1, 1, 0, 0, 0))
+    console.log(date.toLocaleDateString())
```
dnlggr, line 57:
```suggestion
```

src/components/FidelityBondDevOnly.tsx
```diff
@@ -0,0 +1,350 @@
...
+
+  const now = new Date()
+  const currentYear = now.getUTCFullYear()
+  const currentMonth = 1 + now.getUTCMonth() // utc month ranges from [0, 11]
+
+  const [lockdateYear, setLockdateYear] = useState(currentYear + 1)
+  const [lockdateMonth, setLockdateMonth] = useState(currentMonth)
+
+  useEffect(() => {
+    const date = new Date(Date.UTC(lockdateYear, lockdateMonth - 1, 1, 0, 0, 0))
```
dnlggr, line 56:
```suggestion
    const date = new Date(Date.UTC(lockdateYear, lockdateMonth - 1, 1))
```

dnlggr: This comes up quite often and not forgetting the `- 1` part is critical. Should we move this to a function?

src/components/FidelityBondDevOnly.tsx
editwentyone: can we implement something like this? 
image
Source: https://www.primefaces.org/primeng/calendar

editwentyone: found this, maybe it helps: https://raygun.com/blog/building-great-looking-useable-date-time-picker-without-bootstrap/

src/components/CurrentWalletAdvanced.tsx
```diff
@@ -76,21 +81,42 @@ export default function CurrentWalletAdvanced() {
...
+                  <Trans i18nKey="fidelity_bond.alert_no_fidelity_bonds" as="span">
+                    No Fidelity Bond present.
+                  </Trans>{' '}
+                  {featureFidelityBondsEnabled && (
+                    <Link to={routes.fidelityBonds}>
+                      <Trans i18nKey="current_wallet_advanced.link_fidelity_bonds_create_text" as="span">
+                        Create a Fidelity Bond.
+                      </Trans>
+                    </Link>
+                  )}
```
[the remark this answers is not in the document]
AdamISZ: It's important for users to understand that they shouldn't create more than one simultaneously (it'll be economically against their interests), and Joinmarket itself will only ever advertise one (the most valuable one) in the currently in-use wallet. See "Can my yield-generator use multiple timelocked addresses or UTXO?" [here](https://joinmarket-org.github.io/joinmarket-clientserver/fidelity-bonds.html).

theborakompanioni:
Hey guys, thank you very much for your feedback. I'll try to integrate it soon and will keep all things said here in mind while working on #307 today. Will ping you, once updates were made.

src/libs/JmWalletApi.ts
```diff
@@ -29,9 +29,9 @@ type WithMixdepth = {
 type Digit = '0' | '1' | '2' | '3' | '4' | '5' | '6' | '7' | '8' | '9'
 type YYYY = `2${Digit}${Digit}${Digit}`
```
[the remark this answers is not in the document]
theborakompanioni: This is only for increased developer experience.. there are no checks at runtime ; -)

src/components/CurrentWalletAdvanced.tsx
```diff
@@ -76,21 +81,42 @@ export default function CurrentWalletAdvanced() {
...
+        )}
+
+        {!isLoading && fidelityBonds && (
+          <>
+            {fidelityBonds.length === 0 ? (
+              <rb.Alert variant="info">
+                <>
+                  <Trans i18nKey="fidelity_bond.alert_no_fidelity_bonds" as="span">
+                    No Fidelity Bond present.
+                  </Trans>{' '}
```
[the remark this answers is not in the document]
theborakompanioni: If `featureFidelityBondsEnabled` is true, then yes, otherwise no. Moved it to be "rendered" only if feature is enabled. :+1:

src/components/FidelityBondDevOnly.tsx
theborakompanioni: Thought to have the actual day time related values as well for "purposes of documentation", but that is also fine with me. Removed!

src/components/CurrentWalletAdvanced.tsx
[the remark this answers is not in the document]
dnlggr: > It's important for users to understand that they shouldn't create more than one simultaneously (it'll be economically against their interests), and Joinmarket itself will only ever advertise one (the most valuable one) in the currently in-use wallet.

Thanks @AdamISZ for clarifying. That makes a lot of sense.

theborakompanioni:
Closing this in favor of #307 and try to address all feedback there.

PR #222, opened 2022-04-18
ui: small visual adaptions after placeholder PR

theborakompanioni:
Some small stuff that has been noticed after #203 has been merged.

- add border-radius to input field placeholder
- align loading spinner in submit buttons on `Send` and `Receive` screen
- re-add initial collaborative transaction toggle state on Send screen
- restore send form field order ("account to send from" is at the top)
- avoid flickering on receive page (still not sure if the placeholder for the qr code should be displayed, when an address is already present; there is a case for showing it just for the initial loading phase)
- re-add icons on send/receive buttons in `CurrentWalletMagic`

Findings: When the initial page load is the `Send` page, the "collaborative transaction" toggle is not enabled - this should be fixed after is #218 rebased. 

Edit: Uff.. these placeholders are insanely resource intensive.. - my fans spin up if they are displayed for a while (e.g. when token became invalid).

### DIFF
--- a/src/components/CurrentWalletMagic.jsx
+++ b/src/components/CurrentWalletMagic.jsx
@@ -15,18 +15,18 @@ const WalletHeader = ({ name, balance, unit, showBalance, loading }) => {
     <div className="d-flex flex-column align-items-center">
       {loading && (
         <rb.Placeholder as="div" animation="wave">
-          <rb.Placeholder className={styles['wallet-header-title']} />
+          <rb.Placeholder className={styles['wallet-header-title-placeholder']} />
         </rb.Placeholder>
       )}
       {!loading && <h1 className="text-secondary fs-6">{walletDisplayName(name)}</h1>}
       {loading && (
         <rb.Placeholder as="div" animation="wave">
-          <rb.Placeholder className={styles['wallet-header-subtitle']} />
+          <rb.Placeholder className={styles['wallet-header-subtitle-placeholder']} />
         </rb.Placeholder>
       )}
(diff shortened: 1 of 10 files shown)


src/components/CurrentWalletMagic.jsx
```diff
@@ -67,7 +70,7 @@ const LoadingPrivacyLevel = ({ level }) => {
     <div className="d-flex align-items-center">
       <div className="d-flex">{loadingShields}</div>
       <div className="ps-2">
```
Djo1e, line 72: Can you can add a width of `4.3rem` to this div, so it matches the non-loading size?

Djo1e (approved):
Looks good, thanks for the follow-up work @theborakompanioni! 🚀

src/components/CurrentWalletMagic.jsx
theborakompanioni: Would you tell me what you have in mind? I tried but failed to see how `4.3rem` can be applicable. Any hint would be appreciated!

Balance placeholders are tricky, because the balance can either be shown or hidden, nominated in btc or sats - we can never know what the exact width will be beforehand.

Djo1e: Sorry, `4.3rem` only covers one of these cases. I think it's pretty hard to get it correctly for all scenarios, especially because it's centered. Please disregard.

dnlggr (approved):
LGTM ✅ 

Thanks @theborakompanioni!

PR #223, opened 2022-04-19
feat: reset wallet when token became invalid

theborakompanioni:
This PR is needs https://github.com/JoinMarket-Org/joinmarket-clientserver/issues/1244 to be addressed, but can be already applied without impairing the functionality.

Before this PR, if a token became invalid, e.g. when unlocking a wallet from another tab, the UI would display an error message "Invalid credentials" on every authenticated - and subsequently failing - API request. This is because the periodically polled `/session` endpoint does not need authentication and Jam is not aware that anything is wrong.

After this PR, the `/session` is provided with the current token and will reset the wallet and token on `401 Unauthorized` errors. If this happens, users will be redirected to the initial wallet overview page.

The PR can even be applied before the server project addressing the issue, as the `/session` endpoint will ignore any token (valid or invalid) at the moment of writing.


theborakompanioni:
Even if it is safe to apply, let's wait for https://github.com/JoinMarket-Org/joinmarket-clientserver/pull/1251 to be merged.

theborakompanioni:
> Even if it is safe to apply, let's wait for [JoinMarket-Org/joinmarket-clientserver#1251](https://github.com/JoinMarket-Org/joinmarket-clientserver/pull/1251) to be merged.

Merged now. Testing can be done after rebuilding the local regtest setup.

src/context/ServiceInfoContext.tsx
```diff
@@ -74,15 +79,19 @@ const ServiceInfoProvider = ({ children }: React.PropsWithChildren<{}>) => {
...
+              resetWalletAndClearSession()
             }
           }
           return info
         })
         .catch((err) => {
           if (!signal.aborted) {
-            setConnectionError(err)
+            const shouldResetState = err instanceof Api.JmApiError && err.response.status === 401
+            if (shouldResetState) {
```
dnlggr, line 90: Nit I'd love to hear your opinion on:

I'd personally either inline the `err instanceof ...` check or, preferably, name the variable `isUnauthorized`.

My thinking is that it's more clear to have the if-statement read:

"If the request is unauthorized, reset the wallet and clear the session."

Compared to:

"If we should reset the state, reset the wallet and clear the session."

Since the former contains a more clear picture of what's happening compared to the latter. With the variable named `shouldResetState` we're basically moving the if-statement "logic" to the variable assignment and the if-statement itself is not convening much information anymore (at least when it comes to naming).

This is a super esoteric discussion of course and _not_ a change request. I'd still love to hear your opinion on it. I have a soft spot for obsucre naming discussions sometimes. 😄

dnlggr (approved):
LGTM ✅

src/context/ServiceInfoContext.tsx
theborakompanioni: Affirmative. Thanks for taking the time to lay down your thoughts. Totally reasonable and improves readability a lot. 
My opinion is, that I should be more thoughtful about initial variable naming. I sometimes think I can revisit it in the future and end up never doing it. Improving these habits is a skill worth practicing. You'll notice in the future if I could live up to it. Please always mention if you see an opportunity to increase readability.

PR #227, opened 2022-04-21
ui: move gh link to settings and remove other footer links

dnlggr:
The links in the footer don't really add much value imo and look a bit off now that we have the cheatsheet (which is great btw!):

Screenshot 2022-04-21 at 16 37 44

- _🔗 Docs_ was just linking to our [wiki](https://github.com/joinmarket-webui/joinmarket-webui/wiki) which right now is mostly only used for meeting notes and other development related resources. So not something the average user is concerned with much. At least not in a way that it justifies showing this link prominently on every screen. Once we have actual user docs, that might be different of course.
- _🔗 Features_ just linked to our GitHub readme. It's de-facto the same link as _🔗 GitHub_.
- _🔗 Twitter_ linked to the (not very active) [JoinMarket Twitter](https://twitter.com/joinmarket). Not sure the user needs that on every page.
- _🔗 GitHub_ is a useful link. It can also live in the _Settings_ page though imo:

Screenshot 2022-04-21 at 16 36 30

That leaves the _Cheatsheet_ be the only item in the center of the footer, shown on every page which is the prominent place it deserves imo:

Screenshot 2022-04-21 at 16 43 51

Let me know what you think!


editwentyone:
I would keep the following in the footer for the time being: Cheatsheet, Twitter, Github 

Docs should be available again as soon its more useful, Features is just a Category inside Docs isn't it?

theborakompanioni:
First of all: I think it is a valid change request and with the way you've described your thoughts, it is completely reasonable. :+1:

I am certainly missing the GitHub link in the footer, as it highlights that the project is open to all. Moving it to the settings page feels like this fact is downgraded to a second class citizen.. does it make sense to keep it at least as long as we have the "alpha warning"?

Another thing which probably should be addressed, is the translation keys.
There is probably no "one size fits all" solution here, but should be try:
- remove unused keys, in this case "footer.docs", "footer.features", etc.
- have a global namespace (e.g. "global") for frequently used phrases throughout the app (e.g. "loading")

dnlggr:
Ok Let's meet in the middle here :) 

I added social links back to the footer but more subtle just with icons. From my experience this is what most FOSS projects do. Once the alpha warning is gone we can re-arrange everything since we'll have more space. Adding Telegram is a good idea @editwentyone!

Screenshot 2022-04-22 at 11 38 29

As for Twitter, I really think it is next to useless for the user to have is so prominently in the footer. Basically all the [JoinMarket Twitter](https://twitter.com/JoinMarket) does is tweet out release notifications. Those releases don't even correlate with Jam releases so I don't thing there's any value for the user to have this visible in the footer at all times. The only reason why it is still there in the first place is because it's a left over from when we forked the old [jm-web-client](https://github.com/JoinMarket-Org/jm-web-client) project. In the original designs for Jam it wasn't included as well it was just never removed.

theborakompanioni:
> Ok Let's meet in the middle here :)

:pray: 

> I added social links back to the footer but more subtle just with icons. From my experience this is what most FOSS projects do. Once the alpha warning is gone we can re-arrange everything since we'll have more space. Adding Telegram is a good idea @editwentyone!

:+1:

> As for Twitter, I really think it is next to useless for the user to have is so prominently in the footer. Basically all the [JoinMarket Twitter](https://twitter.com/JoinMarket) does is tweet out release notifications. Those releases don't even correlate with Jam releases so I don't thing there's any value for the user to have this visible in the footer at all times. 

:+1:

> We could add it to the settings as some sort of attribution to the server project though?

I am certainly in favor of doing that in any shape or form. Let's see with what we come up with.


One last thing: What about the french translation files? Should they be edited in-place or via Transifex?

src/components/App.jsx
```diff
@@ -180,60 +180,46 @@ export default function App() {
...
+              >
+                <Sprite symbol="github" width="18px" height="18px" className="" />
+              </a>
+              <a
+                href="https://t.me/JoinMarketWebUI"
+                target="_blank"
+                rel="noopener noreferrer"
+                className="d-flex align-items-center text-secondary"
+              >
+                <Sprite symbol="telegram" width="18px" height="18px" className="" />
```
theborakompanioni, line 200:
```suggestion
                <Sprite symbol="telegram" width="18px" height="18px" />
```
?

src/components/App.jsx
```diff
@@ -180,60 +180,46 @@ export default function App() {
...
+          </div>
+          <div className="d-flex flex-1 order-2 justify-content-end align-items-center gap-1">
+            <div className="d-flex gap-2 pe-2">
+              <a
+                href="https://github.com/joinmarket-webui/joinmarket-webui"
+                target="_blank"
+                rel="noopener noreferrer"
+                className="d-flex align-items-center text-secondary"
+              >
+                <Sprite symbol="github" width="18px" height="18px" className="" />
```
theborakompanioni, line 192:
```suggestion
                <Sprite symbol="github" width="18px" height="18px" />
```
?

src/components/App.jsx
```diff
@@ -180,60 +180,46 @@ export default function App() {
...
+            <div className="d-flex">
+              <rb.OverlayTrigger
+                delay={{ hide: 300, show: 200 }}
+                placement="top"
+                overlay={(props) => (
+                  <rb.Tooltip {...props}>
+                    {websocketConnected ? t('footer.connected') : t('footer.disconnected')}
+                  </rb.Tooltip>
+                )}
+              >
```
theborakompanioni, line 213: Tooltip is nice. :+1:

It is misaligned in my browsers though, but that's an issue with react-bootstrap, so I do not know what to do about it, just wanted to let you know (Not using safari here).
I tired upgrading to v2.3.0, but it is still not fixed. (Possibly related: https://github.com/react-bootstrap/react-bootstrap/issues/6010 but in general there are some open issue regarding Tooltips/Overlays.)

theborakompanioni (approved):

dnlggr:
> One last thing: What about the french translation file? Should it be edited in-place or via Transifex?

That's a good question. Not sure how to best handle this but I suspect to just leave it like it is. We really won't be able to update translations for all languages ever in each PR (and probably shouldn't). I feel like we should just update the default language (English) and hope that the translators catch up with us eventually. That's kind of the problem with doing translations at this stage where a lot of stuff is still going to change. But we knew that when we decided to do it so that's fine imo. What do you think?

src/components/App.jsx
dnlggr: Hm interesting, for me it seems fine on Firefox, Chrome and Safari. Or maybe I'm just not seeing the misalignment? 🤔

dnlggr:
>> We could add it to the settings as some sort of attribution to the server project though?

> I am certainly in favor of doing that in any shape or form. Let's see with what we come up with.

Added ✅

theborakompanioni:
> > One last thing: What about the french translation file? Should it be edited in-place or via Transifex?
> 
> That's a good question. Not sure how to best handle this but I suspect to just leave it like it is. We really won't be able to update translations for all languages ever in each PR (and probably shouldn't). I feel like we should just update the default language (English) and hope that the translators catch up with us eventually. That's kind of the problem with doing translations at this stage where a lot of stuff is still going to change. But we knew that when we decided to do it so that's fine imo. What do you think?

Yes, that certainly feels very reasonable.
Also think it is good to already have one additional language next to English included - but would definitely wait for the first official major release before including new ones. :muscle: 

> > I am certainly in favor of doing that in any shape or form. Let's see with what we come up with.
>
> Added :white_check_mark:

Hope you know I did not mean to have it included immediately - just that I'd be in favor. You are always crazy fast : D

Also - sorry I only noticed it now: Settings page in dark-mode needs a tiny little bit of love.

dnlggr:
> Also - sorry I only noticed it now: Settings page in dark-mode needs a tiny little bit of love.

Damn I forgot to check that. Will update here before merging. Thanks! 🙏

dnlggr:
Ok should be ready now @theborakompanioni! 🙏 ✅ 

Looks pretty clean now imo. What do you think @editwentyone? 

Screenshot 2022-04-25 at 14 06 16
Screenshot 2022-04-25 at 14 06 00

PR #231, opened 2022-04-25
reload session info on Earn and Wallets screen

### LINKED ISSUE #225: Reload Session Information on `Earn` and `Wallets` screens
`Earn` and `Wallets` screens make use of information contained in the session information. This request does not have much overhead and the response should arrive quite fast.

Currently, even if reloading session info is triggered as one of the first requests when the app loads, the screens might display outdated information and it cannot be relied upon that this info is already present when the initial rendering is done.

This is blocked by #218

theborakompanioni:
Closes #225.

Reloads the session information on `Earn` and `Wallets` screens.

Additional change: In order to mitigate a problem when the maker is started/stopped too quickly in succession, the "submit" button is re-enabled with a time delay (currently 2 seconds). This behaviour feels a bit strange, but I think it's still better than the service not working at all – feedback on that is welcome.

### DIFF
--- a/src/components/Earn.jsx
+++ b/src/components/Earn.jsx
@@ -90,7 +90,6 @@ export default function Earn() {
   const [alert, setAlert] = useState(null)
   const [isLoading, setIsLoading] = useState(true)
   const [isSending, setIsSending] = useState(false)
-  const [isWaiting, setIsWaiting] = useState(false)
   const [isWaitingMakerStart, setIsWaitingMakerStart] = useState(false)
   const [isWaitingMakerStop, setIsWaitingMakerStop] = useState(false)
   const [isReportLoading, setIsReportLoading] = useState(false)
@@ -160,12 +159,11 @@ export default function Earn() {
       .then((res) => (res.ok ? true : Api.Helper.throwError(res)))
       // show the loader a little longer to avoid flickering
       .then((_) => new Promise((r) => setTimeout(r, 200)))
-      .then((_) => setIsSending(false))
       .catch((e) => {
         setIsWaitingMakerStart(false)
-        setIsSending(false)
         setAlert({ variant: 'danger', message: e.message })
       })
+      .finally(() => setIsSending(false))
   }
 
   const stopMakerService = () => {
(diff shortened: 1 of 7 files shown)


dnlggr:
I feel like it might be a bit confusing that the start/stop button is disabled/enabled independently (time-wise) from the rest of the UI:

https://user-images.githubusercontent.com/10026790/165289312-39830ac6-4f7a-4ba7-a873-38087ef612a4.mov

What we really want is to make sure there is a delay between stopping the service and restarting it right? If that is the case, would artificially prolonging the `stopMakerService` function work (Assuming we disable the button while that operation is ongoing)? Then it would be synced up with the rest of the UI and not do it's own thing on every reload / operation. Let me know if that makes sense at all.

theborakompanioni:
> I feel like it might be a bit confusing that the start/stop button is disabled/enabled independently (time-wise) from the rest of the UI:
> Screen.Recording.2022-04-26.at.13.22.30.mov
> 
> What we really want is to make sure there is a delay between stopping the service and restarting it right? If that is the case, would artificially prolonging the `stopMakerService` function work (Assuming we disable the button while that operation is ongoing)? Then it would be synced up with the rest of the UI and not do it's own thing on every reload / operation. Let me know if that makes sense at all. thinking

Thanks. Of course it makes sense and I thought about this approach – but ultimately decided against it, because I thought it felt even weirder. Wrong decision I guess. Will adapt!

It might be that we also have to prolong `startMakerService`. If that turns out to be necessary, I will change it to delaying _before_ the `start` request and delaying _after_ the `stop` request.

dnlggr:
That sounds good. It's not something people should do often so it'll be ok if each operation takes 2 seconds longer imo.

src/components/Earn.jsx
```diff
@@ -14,7 +14,11 @@ import styles from './Earn.module.css'
 const OFFERTYPE_REL = 'sw0reloffer'
 const OFFERTYPE_ABS = 'sw0absoffer'
 
-const SUBMIT_BUTTON_ENABLE_DELAY_IN_MS = 2_000
+// In order to prevent observed state mismatch, the 'maker stop' response is delayed shortly.
+// Even though the API states the maker as started or stopped, it seems this is reported prematurely.
+// There is currently no way to know for sure - adding a delay at least migitages the problem.
+// 2022-04-26: With value of 2_000ms, no state corruption could be provoked in a local dev setup.
```
dnlggr, line 20:
```suggestion
// In order to prevent state mismatch, the 'maker stop' response is delayed shortly.
// Even though the API response suggests that the maker has started or stopped immediately, it seems that this is not always the case.
// There is currently no way to know for sure - adding a delay at least mitigates the problem.
// 2022-04-26: With value of 2_000ms, no state corruption could be provoked in a local dev setup.
```

dnlggr (commented):
Awesome feels great now! ✅ 👏 Good job!

dnlggr (approved):

PR #232, opened 2022-04-28
refactor: reusable copy-to-clipboard mechanism

theborakompanioni:
The current figma screens contain copy functionality in some components.
Currently, only the Receive screen has this capability for coping addresses into the clipboard.
This includes a fallback mechanism with an hidden input field in case a browser does not support it natively.

In order to make it easier to implement this feature in other components, a universal `Copyable` component, a `CopyButton` and a `CopyButtonWithConfirmation` has been externalized.

`Copyable`: a div that copies a given value into the clipboard on click (encapsulating the fallback mechansim)
`CopyButton`: a simple reusable button styled div
`CopyButtonWithConfirmation`: a highly opinionated button component that shows a visual confirmation after successful operation

With these changes applied, it should be easier to add copy functionality to other screens while being flexible regarding component style.


src/components/Receive.jsx
```diff
@@ -122,42 +91,11 @@ export default function Receive() {
...
+              <CopyButtonWithConfirmation
                 value={address}
-                style={{
-                  position: 'absolute',
-                  left: '-9999px',
-                  top: '-9999px',
-                }}
+                text={t('receive.button_copy_address')}
+                successText={t('receive.text_copy_address_confirmed')}
+                disabled={!address || isLoading}
```
Djo1e, line 98: I don't think this `disabled` prop is being passed further down

src/utils.ts
```diff
@@ -0,0 +1,46 @@
...
+
+export const walletDisplayName = (name: string) => name.replace('.jmdat', '')
+
+export const displayDate = (string: string) => new Date(string).toLocaleString()
+
+export const btcToSats = (value: string) => Math.round(parseFloat(value) * 100000000)
+
+export const satsToBtc = (value: string) => parseInt(value, 10) / 100000000
+
+export const copyToClipboard = (
```
Djo1e, line 14: Should this function be async?

src/components/Copyable.tsx
```diff
@@ -0,0 +1,43 @@
...
+  onError,
+  className,
+  children,
+  ...props
+}: PropsWithChildren<CopyableProps>) {
+  const valueFallbackInputRef = useRef(null)
+
+  return (
+    <>
+      <div
```
Djo1e, line 23: WDYT about changing this to a `button` instead? For better accessibility

src/components/Copyable.tsx
```diff
@@ -0,0 +1,43 @@
...
+  className?: string
+}
+
+export function Copyable({
+  value,
+  onSuccess,
+  onError,
+  className,
+  children,
+  ...props
```
Djo1e, line 17: I don't think `...props` is doing anything here

src/components/Receive.jsx
theborakompanioni: Ah, thanks – nice catch. Fixed.

src/components/Copyable.tsx
theborakompanioni: Affirmative. Done!

src/components/Copyable.tsx
theborakompanioni: Now it does - passing other properties, e.g. "disabled" : D
Thanks!

src/components/Copyable.tsx
```diff
@@ -0,0 +1,43 @@
...
+import { copyToClipboard } from '../utils'
+
+export interface CopyableProps {
+  value: string
+  onSuccess?: () => void
+  onError?: (e: Error) => void
+  className?: string
+}
+
+export function Copyable({
```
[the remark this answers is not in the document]
theborakompanioni: Yep, I'll do that. As far as I understood, the idea is that all addresses can be copied on click in the current figma screens – but is seems that needs a bit more tim. Now that `Copyable` renders as `button`, as rightfully suggested by @Djo1e , it is also well placed in `CopyButton.tsx`. :+1:

theborakompanioni:
Thank you for your feedback! Hope everything is addressed and working as expected.

src/utils.ts
Djo1e: I generally prefer async/await, as I find it more readable. 

But I was suggesting changing this
```suggestion
export const copyToClipboard = async (
```
because it might return a promise

Djo1e (approved):
looks good, thanks for addressing feedback 👍

src/utils.ts
theborakompanioni: > because it might return a promise

`copyToClipboard` returns `Promise<boolean>` -> You can always use `await` in any `async` function calling it. Hope it is acceptable not using `async`/`await` in the function itself.

dnlggr (approved):

PR #235, opened 2022-05-02
Formik validation in Wallet component

theborakompanioni:
One more step in the direction of using Formik for every form throughout the app.

This PR splits the former unified form for locking/unlocking a wallet into a `WalletUnlockForm` and a `WalletLockForm`.

The `WalletLockForm` really consists of a submit button only, so maybe it's not even necessary to wrap it in a form (I do not know if it has benefits for screen readers, etc.).. The button was in a form before, so it is in a form now – willing to change it if someone has feedback on that, as passing empty `initialValues` and noop `validate` method to Formik feels rather clumsy.

As far as I noticed, there are no visual changes.
Functionality change introduced: Before, the lock button just triggered locking the current wallet. After, the "lock" action now knows which "wallet" it belongs to. In practice this currently does not matter (as the lock button is only visible for the active wallet anyway).


src/components/Earn.jsx
```diff
@@ -327,8 +326,6 @@ export default function Earn() {
     } catch (e) {
       setServiceInfoAlert(null)
       setAlert({ variant: 'danger', message: e.message })
-    } finally {
-      setSubmitting(false)
```
dnlggr, line 331: We're these calls not doing anything before?

dnlggr (approved):
Looks good! ✅ 

Imo there's no reason for keeping the no-op form for locking the wallet. Other buttons in the app also don't have a form so it should be fine to just have a button there. But you can definitely also just keep it the way it is  -- I don't see a problem with the no-op form either. Whatever you prefer is fine!

src/components/Earn.jsx
theborakompanioni: Formik resets the flag automatically when using `async` functions.

dnlggr:
Oh no, I made a type when merging. 🤦

PR #236, opened 2022-05-02
Reload wallet data after send 

### LINKED ISSUE #187: Reload wallet data after send
See #184.

After sending we should reload the wallet data so that the "Send" page is ready to be used again and displays accurate data. We need to do it somehow silently in the background though. Or freeze the UI and reload data in the background. We don't want the page to flicker and reload immediately after sending as this feels spooky and could make the user feel that something with sending went wrong.

Ideally we implement this for the next release (currently 0.1.0 IIRC).

theborakompanioni:
Closes #187.

Main goal of these changes: Users should see the correct balance after a send.

Before this PR, the send page contained outdated information after submitting the form, as relevant information is only loaded once. After this PR, the relevant information is reloaded and the correct balance is shown.

In order to know if the balance is synchronized, it must be waited till the spent utxos are removed from the wallet's utxo set.
This can be done for "direct sends", but not for collaborative transactions.

So the approach is different for both
- after direct-send, wait for wallet's utxo set to be synchronized, then reload data
- after starting taker operation, wait for it to finish, then reload data

### DIFF
--- a/src/components/Send.jsx
+++ b/src/components/Send.jsx
@@ -272,9 +272,13 @@ export default function Send() {
         .catch((err) => {
           if (abortCtrl.signal.aborted) return
 
+          // Stop waiting for wallet synchronization on errors, but inform
+          // the user that loading the wallet info failed
+          setWaitForUtxosToBeSpent([])
+          setIsLoading(false)
+
           const message = err.message || t('send.error_loading_wallet_failed')
           setAlert({ variant: 'danger', message })
-          setIsLoading(false)
         })
     }, initialDelayInMs)
 
--- a/src/components/Send.jsx
+++ b/src/components/Send.jsx
@@ -182,7 +182,7 @@ export default function Send() {
 
   const [paymentSuccessfulInfoAlert, setPaymentSuccessfulInfoAlert] = useState(null)
   const [waitForUtxosToBeSpent, setWaitForUtxosToBeSpent] = useState([])
-  // TODO: It is not obvious when to clear this data, might be better to display this as "toast"
+  const [waitForTakerToFinish, setWaitForTakerToFinish] = useState(false)
   const [takerStartedInfoAlert, setTakerStartedInfoAlert] = useState(null)
 
   const [isLoading, setIsLoading] = useState(true)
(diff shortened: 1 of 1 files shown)


dnlggr:
Initial feedback after testing for a direct send: UX wise, I feel like we should delay the "Transaction successful" alert until the reload is done. Right now it's a bit weird seeing "Transaction successful" but the UI still loading. It'd be artificially making a send take longer (at least from a user's point of view) for the second or so it takes to reload. But the more smooth UI is worth it imo. Then everything would feel like one single operation. What do you think?

theborakompanioni:
> Initial feedback after testing for a direct send: UX wise, I feel like we should delay the "Transaction successful" alert until the reload is done. Right now it's a bit weird seeing "Transaction successful" but the UI still loading. It'd be artificially making a send take longer (at least from a user's point of view) for the second or so it takes to reload. But the more smooth UI is worth it imo. Then everything would feel like one single operation. What do you think?

This is a valid observation and your proposed solution is a great idea! :100: 
I think it takes a little bit of further tweaking to get it right, though Currently, the inputs would turn into placeholders which should probably done differently when waiting for the success message to appear. Also, what should be done in case the reloading fails – just display both, the success message _and_ the "wallet info not loaded" alert?  I try to come up with an implementation which will needs another review anyway. Thank you @dnlggr :rocket:

dnlggr:
> Also, what should be done in case the reloading fails – just display both, the success message and the "wallet info not loaded" alert?

Oh that's a good point. The "send" is definitely the more important operation -- obviously. So yeah probably both would make sense, I guess. 🤔 

I thought it would be an easy change but I haven't really thought it through fully. If it is too much work, don't over think it we can also do that in an improvement story if needed... :)

theborakompanioni:
> > Also, what should be done in case the reloading fails – just display both, the success message and the "wallet info not loaded" alert?
> 
> Oh that's a good point. The "send" is definitely the more important operation -- obviously. So yeah probably both would make sense, I guess. thinking
> 
> I thought it would be an easy change but I haven't really thought it through fully. If it is too much work, don't over think it we can also do that in an improvement story if needed... :)

If you think this is a valid way to move forward, let's do this. Current changes feel like an overall improvement, albeit, as you mentioned, not quite optimal yet. I'd turn your comment above into a new issue once this is merged.

dnlggr:
> If you think this is a valid way to move forward, let's do this. Current changes feel like an overall improvement, albeit, as you mentioned, not quite optimal yet. I'd turn your comment above into a new issue once this is merged.

Sounds good!

Speaking of successful transactions: I think we should probably also think about a better visual indication to make it more clear for the user that a collaborative transaction has finished successfully (or not). Some sort of nice green checkmark or something would definitely be more satisfying than the "Joining" text in the header simply disappearing. Just noting down my thoughts. Completely unrelated to this PR of course.

dnlggr (approved):

PR #239, opened 2022-05-06
ui: decrease brightness of placeholder in light theme

theborakompanioni:
It feels that the placeholders in light mode are little bit too bright. On some screens it is hard to tell the difference from the background. Do you think it would be an improvement if they were a little darker?

## 📸 Before
image

## 📸 After
image

Just for reference, this would be the original Bootstrap style (applying `currentColor`):
image

### DIFF
--- a/src/index.css
+++ b/src/index.css
@@ -502,6 +502,10 @@ h2 {
     center/1em auto no-repeat;
 }
 
+.placeholder {
+  background-color: var(--bs-gray-400);
+}
+
 /* Privacy Level Styles */
 
 .privacy-level-4 {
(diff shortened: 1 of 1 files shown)


editwentyone:
its ok, go a little darker, but not that dark as in the last screenshot. a spinner per block would also help to visualize that something is loading

theborakompanioni:
> its ok, go a little darker, but not that dark as in the last screenshot. a spinner per block would also help to visualize that something is loading

Would you approve the PR if you are happy with the changes? :orange_heart:

dnlggr (approved):
LGTM ✅

PR #240, opened 2022-05-06
refactor: remove catch all style rule for border-color in dark mode

theborakompanioni:
Remove catch all css rule overriding all border colors in dark mode.

Some bootstrap elements apply `border-color: transparent` to elements for a reason, e.g. buttons without borders.
These rules are overwritten by the custom rule included in `index.css`
```css
:root[data-theme='dark'] * {
  border-color: var(--bs-gray-800) !important;
}
```

It has happened more than once now that this rule gets in the way rather than being helpful.
A benefit of this change is that elements are easier to use and more consistent across themes.
A drawback is that all new elements must take the border color into account.

Also, visual effects on the elements when removing the catch all rule:
- Border color of input fields on validation (works in light mode)
  - valid input fields get a green border
  - invalid input fields get a red border

## 📸 Before
image

image

## 📸 After
image
image

(Notice the border colors of the input fields and buttons)

Hopefully all affected elements are covered with these changes.
What do you think? Does it have advantages or would it make things complicated again in the long run?

If these changes do not provide an benefit, simply close the PR.

### DIFF
--- a/src/components/Send.module.css
+++ b/src/components/Send.module.css
@@ -104,8 +104,8 @@ input[type='number'] {
   color: var(--bs-white);
 }
 
-:root[data-theme='dark'] .selected-dark {
-  background-color: var(--bs-gray-dark) !important;
+:root[data-theme='dark'] .collaborators-selector .selected-dark {
+  background-color: var(--bs-gray-dark);
   border-color: var(--bs-gray-dark) !important;
 }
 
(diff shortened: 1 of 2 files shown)


dergigi (approved):
utACK

Looks good to me ✅

editwentyone (approved):
all good so far, a redesign will fix the visibility later on :)

theborakompanioni:
> all good so far, a redesign will fix the visibility later on :)

Thank you. Will merge after the scheduler PRs are included.

PR #241, opened 2022-05-06
ui: Wallet Detail Overlay

theborakompanioni:
Adds a "wallet detail overlay" view.
Current views slightly differ from the designs in figma, but most functionality is already present:
- switching through accounts
- accordion for branches
- abbreviated derivation path, addresses, labels, balances
- copy address

Not integrated yet:
- freeze/unfreeze

https://user-images.githubusercontent.com/3358649/170373054-c5f6a03e-f924-4360-91bb-607cd8b8a014.mov


editwentyone:
I like it so far, it would be cool, if we could shorten the address when the space gets smaller like mobile: 

image

theborakompanioni:
@editwentyone @dnlggr Not entirely sure if this should be kept the way it is. In "Simple View" a user is probably not interested in any of the displayed information except for the xpubs. What about changing it to show utxos instead and enable real coin control for everyone?

editwentyone:
i will add coin control / freeze / unfreeze on UTXO level, soon tm

editwentyone:
I need help to understand addresses and utxo to design coin control, or does someone has a best practice example?

dnlggr:
Coin control, focus on UOXOs etc. plays a lot into the "redesign privacy shields" topic. I feel like we can't solve one without the other. Redesigning the main wallet view and removing the shields is the next big thing to solve after fidelity bonds so we'll have a lot of discussions around that. There isn't really an easy, straight forward best practice to follow here if you ask me.

After thinking about it again, I'd say let's merge this overlay for now since it's a strict improvement to the current "simple" wallet view. It's definitely not the perfect and final solution to all of our problems, but it's a step that we can build on. At least it'll enable us to remove the "advanced" wallet view since the "simple" view will now have the same features as the advanced view. That's already a good step in the right direction if you ask me. (Is that correct @theborakompanioni? I might be missing something.) 

What do you guys think? If you agree, let me know and I'll do a code review so that we can merge this.

editwentyone:
ACK, I'm good with this first step, lets learn fast on it

theborakompanioni:
> I pushed some UI updates. Nothing big. I'd say we merge it now but will revisit all of this later on when we have a more streamlined idea oh how we should do the "wallet divedown".

Do you think it is okay to show the UTXOs with minimal coin control and hide the currently shown details in an "advanced" accordion, like on other pages? I'd really be in favor of doing that.

Having said that, I am also okay with merging it now and create a follow up PR to do showcase these changes. What do you think @dnlggr?

dnlggr:
Let's merge this now. ✅  

I do think the "digging deeper into the wallet" needs a redesign before not too long though. The current state is more or less the the same as the CLI which is fine if you know what you're doing but feels a bit cluttered and overwhelming at first.

Maybe putting more focus on UTXOs at first and have the super detailed wallet information another level deeper can help making it more intuitive? When we talk about Jars we often talk about them containing the UTXOs. Representing that in the UI would be interesting imo.

Issue #248, opened 2022-05-11
Copy for scheduled transactions


dnlggr:
See https://github.com/joinmarket-webui/joinmarket-webui/pull/242.

Copy is preliminary at the moment and we should think about it again before releasing.

What's currently there:

<img width="619" alt="Screenshot 2022-05-11 at 09 45 49" src="https://user-images.githubusercontent.com/10026790/167796243-4d31de6e-96e0-48ac-bf57-b920f3f1cf9a.png">

dergigi:
Typo: autoamtically -> automatically 😜 

In all seriousness, I'll do my best to get this done today.

dnlggr:
Copy for main Scheduler screen resolved in #257. ✅ 

There's another piece of text in https://github.com/joinmarket-webui/joinmarket-webui/pull/262. Maybe have a quick look at that as well @dergigi. 🙏

PR #257, opened 2022-05-11
feat(scheduler): invert external/internal toggle

dergigi:
This PR inverts the external address toggle and adjusts the behavior of the scheduler to "internal" by default.

Also adjusts and updates copy, fixing #248

Screenshot 2022-05-11 at 15 11 07

Screenshot 2022-05-11 at 15 11 18

### DIFF
--- a/src/components/Jam.jsx
+++ b/src/components/Jam.jsx
@@ -200,81 +200,91 @@ export default function Jam() {
           <p className="text-secondary mb-4">{t('scheduler.description_destination_addresses')}</p>
         </>
       )}
-      <Formik
-        initialValues={{ dest1: '', dest2: '', dest3: '' }}
-        validate={(values) => {
-          if (collaborativeOperationRunning) {
-            return {}
-          }
+      {!walletInfo ? (
+        <rb.Placeholder as="div" animation="wave">
+          <rb.Placeholder xs={12} className={styles['input-loader']} />
+        </rb.Placeholder>
+      ) : (
+        <Formik
+          initialValues={getNewAddresses(3, INTERNAL_DEST_ACCOUNT).reduce(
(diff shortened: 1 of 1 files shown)


dnlggr (changes requested):
That's a good change! Much clearer than  before. 👏 👍 

One issue though: When the screen initially loads, the internal destination addresses won't be set and the button "Start" button will be disabled. That's because the internal addresses will only be set once the toggle [is actively un-toggled](https://github.com/joinmarket-webui/joinmarket-webui/pull/257/files#diff-5aa1ef24bb3d55e6829234b19485cd23b10ff61aa8ac49179828246df4857c00R262). Only after toggling the button back and forth will the internal addresses be set and the button be activated. 

With the new logic, we'll need to provide internal destination addresses as default values for the address fields.

https://user-images.githubusercontent.com/10026790/168023614-759bc5e7-70b3-4cb7-9edb-52fd9403dd91.mov

dnlggr:
Pushed a fix that should work (I hope). TL;DR: Waits with rendering the form unil the wallet is loaded. Once the wallet is loaded, [sets the form's initial values](https://github.com/joinmarket-webui/joinmarket-webui/pull/257/commits/e2f1d59c6712c5548f015314dcb12f07e1c4f499#diff-5aa1ef24bb3d55e6829234b19485cd23b10ff61aa8ac49179828246df4857c00R209-R212) to 3 internal addresses.

@theborakompanioni Can you have a quick look please? 🙏  Don't want to approve my own code. 😅

dergigi:
> When the screen initially loads, the internal destination addresses won't be set

Ah, damn. Thanks for catching and fixing that! 🙏

theborakompanioni (approved):
lgtm :+1:

PR #262, opened 2022-05-12
feat: simple progress report for scheduled transactions

### LINKED ISSUE #245: Simple progress report for scheduled transactions

Add a simple progress report based on the transaction schedule to let the user know how long a schedule is roughly going to last and what its progress is.

dnlggr:
Resolves #245.

~Still need to fix a few smaller things like i18n and coordinating the requests and placeholders on initial page load.~

As always, the copy is just off the top of my head. Feel free to change it!

But in principle it should be ready for review I hope. The CSS looks more complicated than it actually is. Sorry for that. I basically adapted an existing progress bar I found for our use case and it just grew into what it is now. 😅 

Let me know what you think!

## 📸 

https://user-images.githubusercontent.com/10026790/168109364-f609cd7e-3b3d-4bcb-8cbd-d37b4ca1083b.mov

### DIFF
--- a/src/components/Jam.jsx
+++ b/src/components/Jam.jsx
@@ -36,7 +36,7 @@ export default function Jam() {
 
   const [alert, setAlert] = useState(null)
   const [isLoading, setIsLoading] = useState(true)
-  const [destinationIsInternal, setDestinationIsInternal] = useState(false)
+  const [destinationIsExternal, setDestinationIsExternal] = useState(false)
   const [collaborativeOperationRunning, setCollaborativeOperationRunning] = useState(false)
   const [schedule, setSchedule] = useState(null)
 
@@ -150,7 +150,7 @@ export default function Jam() {
       setIsLoading(false)
 
       if (!res.ok) {
-        await Api.Helper.throwError(res, t('schedule.error_starting_schedule_failed'))
+        await Api.Helper.throwError(res, t('scheduler.error_starting_schedule_failed'))
       }
 
       setCollaborativeOperationRunning(true)
@@ -173,7 +173,7 @@ export default function Jam() {
       setIsLoading(false)
 
       if (!res.ok) {
-        await Api.Helper.throwError(res, t('schedule.error_stopping_schedule_failed'))
+        await Api.Helper.throwError(res, t('scheduler.error_stopping_schedule_failed'))
       }
 
       setCollaborativeOperationRunning(false)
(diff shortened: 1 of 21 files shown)


editwentyone (commented):
already a very good start! I like it so far. is there a need to add the possibility to read/view "logs" live?

MaxHillebrand (approved):
This is great!

Why are the points on the time line not evenly distributed?

dnlggr:
> Why are the points on the time line not evenly distributed?

The points represent transactions. The schedule includes randomly chosen wait times between each transaction (to prevent timing attacks). The distance between two points roughly corresponds to the wait time between the two respective transactions. Each schedule run will have a different distribution of points on the time line.

dnlggr:
Should be ready now @theborakompanioni. 🙏 Sorry for forgetting to actually reload the schedule first. 😅

src/components/Jam.jsx
```diff
@@ -106,22 +147,21 @@ export default function Jam() {
...
-    //     liquiditywait: 10,
-    //     waittime: 1.0,
-    //     mixdepthsrc: 0,
-    //   }
-    // }
+    if (process.env.NODE_ENV === 'development' && useInsecureTestingSettings && !deactivateTestingToggle) {
+      body.tumbler_options = {
+        addrcount: 3,
+        minmakercount: 1,
+        makercountrange: [1, 1],
```
theborakompanioni, line 154:
```suggestion
        makercountrange: [1, 0],
```

https://github.com/JoinMarket-Org/joinmarket-clientserver/blob/50f5228498c670258740cd3cab485bcb39e177a3/jmclient/jmclient/cli_options.py#L317-L318

Second param is `spread`. We want exactly one maker in regtest.

src/components/Jam.jsx
```diff
@@ -72,6 +77,42 @@ export default function Jam() {
     setCollaborativeOperationRunning(coinjoinInProgress || makerRunning)
   }, [serviceInfo])
 
+  const reloadSchedule = useCallback(
+    ({ signal }) => {
+      return Api.getTumblerSchedule({ walletName: wallet.name, token: wallet.token })
```
theborakompanioni, line 82:
```suggestion
      return Api.getTumblerSchedule({ walletName: wallet.name, token: wallet.token, signal })
```

`signal` is unused.

src/components/Jam.jsx
```diff
@@ -72,6 +77,42 @@ export default function Jam() {
...
+        .then((res) => (res.ok ? res.json() : Api.Helper.throwError(res)))
+        .then((data) => {
+          setSchedule(data.schedule)
```
theborakompanioni, line 85:
```suggestion
          !signal.aborted && setSchedule(data.schedule)
```

src/components/Jam.jsx
```diff
@@ -72,6 +77,42 @@ export default function Jam() {
...
+        })
+        .catch((err) => {
+          // Not finding a schedule is not an error.
+          // It means a single colalborative transaction is running.
+          // Those have no schedule.
```
theborakompanioni, line 90: Not finding a schedule with 404 is not an error. Everything else is an error, isn't it?
Status code can be obtained with `err.response.status` if it is an `JmApiError`.

theborakompanioni (approved):
Looks quite nice overall.

I really laughed out loud when I saw "finished in 53 seconds" and the corresponding translation string `progress_tldr_seconds` – shows how you care about little details, even if very few people will ever see it.

What helped reviewing was to see the raw schedule data, so it would be nice to log it to console or even display it below the form in developer mode.

I'll spare you my nits, most of them can be addressed when working on the other related issues.

Other than that and what was mentioned as comment: Again, really nice. Approved :rocket:

src/components/Jam.jsx
dnlggr: Good point! Updated. ✅

dnlggr:
> even if very few people will ever see it.

No one will ever see it, all prod schedules will be way longer. But we will see it every single time while testing. Worth it. 😄 

> I'll spare you my nits, most of them can be addressed when working on the other related issues.

Don't worry, let's hear the nits! We can always delay addressing them anyway.

PR #263, opened 2022-05-12
fix: prevent starting scheduler when utxo preconditions are not met

### LINKED ISSUE #244: Don't allow scheduled transactions when there are no spendable UTXOs

This can lead to weird states in the backend so we'll simply make sure that users don't start a transaction schedule without having something to transact with.

This should be a s simple as going over all accounts and check if there are any non-locked UTXOs.

theborakompanioni:
Closes #244 

(Summary copied from the comments below)

Preconditions are currently as follows:
```js
const SCHEDULE_PRECONDITIONS = {
  MIN_NUMBER_OF_UTXOS: 1, // min amount of utxos available
  MIN_OVERALL_REMAINING_RETRIES: 1, // amount of overall retries available
  MIN_CONFIRMATIONS_OF_SINGLE_UTXO: 5, // at least one utxo needs X confirmations
}
```

Read: At least `MIN_NUMBER_OF_UTXOS` utxos must be available. One of them needs at least `MIN_CONFIRMATIONS_OF_SINGLE_UTXO` confirmations. All retries combined must be more than `MIN_OVERALL_REMAINING_RETRIES`.

In the current settings this means a minimum requirement is "1 utxo with 5 confirmations with at least 1 retry left".

The json is still rendered so we do not forget to reiterate on the wording. If the preconditions seem fine from a technical perspective, we should come up with a good way to communicate them to the user. Do you have a suggestion?

## 📸 
See comments below.


theborakompanioni:
Related https://github.com/joinmarket-webui/joinmarket-webui/issues/155

theborakompanioni:
Preventing running the scheduler is all good and well, however, providing feedback to the user is the tricky part.
Current working state includes a "precondition summary" with these fields available:

```js
{
  isFulfilled: <boolean>,
  numberOfMissingUtxos: <number>,
  numberOfMissingUtxosWithRetriesRemaining: <number>,
  amountOfMissingConfirmations: <number>,
  amountOfMissingOverallRetries: <number>
}
```

Any change we can display this is a nice way?

Or is a general phrase enough? e.g.
> "tarting the scheduler requires a minimum of `n` unspent transaction outputs with at least `m` confirmations."

Tests with various different limits (e.g. only 1 utxo with 1 remaining retry) still to be done!

Any feedback is already highly appreciated! ping @dnlggr

theborakompanioni:
If I understand correctly, the preconditions must be met for the account the scheduler is started from, in Jam's case, this is always `0` at the time of writing. (As hinted by [Adam here](https://github.com/joinmarket-webui/joinmarket-webui/issues/244#issuecomment-1125168912))

dnlggr:
> If I understand correctly, the preconditions must be met for the account the scheduler is started from,

Yes the way I read it, this is correct. Although, reading [Adam's comment](https://github.com/joinmarket-webui/joinmarket-webui/issues/244#issuecomment-1125168912) again, if the the minimum confirmations are not met, it seems like JM will simply retry after 20 mins. I'm wondering if we should just rely on this retry mechanism and not do too much checks ourselves that we then need to explain to the user. The user won't care if the schedule takes 20mins longer or not.

dnlggr:
Couldn't confirm the retry mechanism. I'd say make 1 UTXO of at least 5 confirmations the required precondition. The 3 UTXOs are [not really needed](https://github.com/joinmarket-webui/joinmarket-webui/issues/244#issuecomment-1128951729) at least for now. And in terms of size reqirement (20% the size of the CJ is required), I think that's also an unlikely edge case (assuming users have only a couple of UTXOs to start with).

If the 5 confirmations requiirement turns out to bee too little, we can still add the other things.

theborakompanioni:
> Couldn't confirm the retry mechanism yet. I'd say make 1 UTXO of at least 5 confirmations the required precondition. The 3 UTXOs are [not really needed](https://github.com/joinmarket-webui/joinmarket-webui/issues/244#issuecomment-1128951729) at least for now. And in terms of size reqirement (20% the size of the CJ is required), I think that's also an unlikely edge case (assuming users have only a couple of UTXOs to start with).
> 
> If the 5 confirmations requiirement turns out to bee too little, we can still add the other things later.

That sounds reasonable. Adapted!

Preconditions are currently as follows:
```js
const SCHEDULE_PRECONDITIONS = {
  MIN_NUMBER_OF_UTXOS: 1, // min amount of utxos available
  MIN_OVERALL_REMAINING_RETRIES: 1, // amount of overall retries available
  MIN_CONFIRMATIONS_OF_SINGLE_UTXO: 5, // at least one utxo needs X confirmations
}
```

Does that sound about right to you?

No utxo avialable:


One utxo with only 3 confirmations (too few):


No retries remaining:


The json is still rendered so we do not forget to reiterate on the wording. If the preconditions seem fine from a technical perspective, we should come up with a good way to communicate them to the user. Do you have a suggestion?

src/components/Jam.jsx
```diff
@@ -31,6 +39,61 @@ const ValuesListener = ({ handler }) => {
   return null
 }
 
+const DEFAULT_PRECONDITION_SUMMARY = {
+  isFulfilled: false,
+  numberOfMissingUtxos: SCHEDULER_PRECONDITIONS.MIN_NUMBER_OF_UTXOS,
+  amountOfMissingConfirmations: SCHEDULER_PRECONDITIONS.MIN_CONFIRMATIONS_OF_SINGLE_UTXO,
+  amountOfMissingOverallRetries: SCHEDULER_PRECONDITIONS.MIN_OVERALL_REMAINING_RETRIES,
```
dnlggr, line 46: How do these retires work exactly? I was under the impression each transaction is a "try". Now it seems like tries are attached to UTXOs. Does that mean if I do a transaction with 5 UTXO, I need 5 tries? Does every commitment give every other UTXO in the same mixdepth 3 tries?

src/components/Jam.jsx
```diff
@@ -31,6 +39,61 @@ const ValuesListener = ({ handler }) => {
...
+
+const useSchedulerPreconditionSummary = (walletInfoOrNull, startAccountIndex) => {
+  const eligibleUtxos = useMemo(() => {
+    if (!walletInfoOrNull) return []
+
+    const utxos = walletInfoOrNull.data.utxos.utxos || []
+    return utxos
+      .filter((it) => it.mixdepth === startAccountIndex)
+      .filter((it) => !it.frozen)
+      .filter((it) => !it.locktime)
```
dnlggr, line 57: Can frozen UTXOs _not_ be the source of a comitment?

dnlggr (commented):
For me there's quite a bit of uncertainty still as to how all these requirements exactly work. The only thin I do fully understand is that there needs to be at least one UTXO with 5 or more confirmations. 

Do you have a better understanding of how what the `tries_remaining` mean exactly and if frozen or unfrozen coins can be the source of a commitment? If we're unsure I'd say let's literally only check for the confirmation and not care about the other things for now.

src/components/Jam.jsx
theborakompanioni: This is controlled by config param `taker_utxo_retries`. By default, this is set to `3`. 

> Now it seems like tries are attached to UTXOs

Yes, this is per utxo. The `/utxo` response from the API has following properties included:
```ts
  tries: number
  tries_remaining: number
```
With an untouched/unused utxo it will return:
```ts
{
  [...]
  "tries": 0,
  "tries_remaining": 3
}
```
> Does that mean if I do a transaction with 5 UTXO, I need 5 tries?

No, I do not think so. I think you need 1 commitment and will have 5 * 3 = 15 tries for that. Not 100% sure.

> Once you've tried 3 times, you have to use a different utxo (some Makers may choose to be more lax, but this will be the usual requirement)`

It is not _totally_ clear how it all works from https://github.com/JoinMarket-Org/joinmarket-clientserver/blob/master/docs/SOURCING-COMMITMENTS.md#commitments-for-initiating-coinjoins, but I think the assumption that you need at least one remaining try, like implemented in the code, is valid and a minimum requirement. I'd be in favor of keeping it.

> As an example, if you send to 3 addresses instead of 1, you will then have 9 commiments available - effectively 9 tries at doing the first transaction, instead of only 3 tries if you send the coins to one address.

src/components/Jam.jsx
theborakompanioni: > Can frozen UTXOs _not_ be the source of a comitment?

I am not super sure, but I do not think so. Reading [Adam's blog post about P(o)ODLE](https://reyify.com/blog/poodle), it says:
> Later on in the protocol you will open the commitment and reveal the utxo.

Frozen utxos wont take part in the collaborative transaction and can therefore not be used in the commitment. Does that make sense?

dnlggr: > Frozen utxos wont take part in the collaborative transaction and can therefore not be used in the commitment.

Not sure if that's the implication from that. Theoretically all you want to do is prove that you have some UTXO right? But I guess if we apply stricter restrictions and they turn out to be too strict we can always loosen them later.

theborakompanioni:
> This is not perfect and oversimplifies it. Especially for edge cases like "no more tries left". But I think this is fine for now. We're still at a stage where the users have to figure out a lot of stuff on their own anyway.

Thank you. Much better now!

> **No old enough UTXO available:** The scheduler requires one of your UTXOs to have **5** or more confirmations. Your oldest UTXO has **3** confirmations. Wait for **2** more blocks.

One thing came to mind though: 
> "Your oldest UTXO has **3** confirmations." 

Can stating "your oldest utxo" be confusing as it only valid for the starting jar?
I'd just remove the whole sentence. Acceptable?

dnlggr:
> I'd just remove the whole sentence. Acceptable?

👍 👌

theborakompanioni:
**No UTXO available:**


**No old enough UTXO available:**


**No retries remaining:**

src/components/Jam.jsx
```diff
@@ -246,6 +316,50 @@ export default function Jam() {
...
+                    more confirmations. Wait for{' '}
+                    <strong>
+                      {{ amountOfMissingConfirmations: schedulerPreconditionSummary.amountOfMissingConfirmations }}
+                    </strong>{' '}
+                    more block(s).
+                  </Trans>
+                ) : (
+                  schedulerPreconditionSummary.amountOfMissingOverallRetries > 0 && (
+                    <Trans i18nKey="scheduler.precondition.hint_missing_overall_retries">
+                      You've tried running the scheduler unsuccessfully to many times in a row. For security reasons,
```
dergigi, line 347:
```suggestion
                      You've tried running the scheduler unsuccessfully too many times in a row. For security reasons,
```

src/components/Jam.jsx
```diff
@@ -246,6 +316,50 @@ export default function Jam() {
...
+                      you need a fresh UTXO to try again. See{' '}
+                      <a
+                        href="https://github.com/JoinMarket-Org/joinmarket/wiki/Sourcing-commitments-for-joins#sourcing-external-commitments"
+                        target="_blank"
+                        rel="noopener noreferrer"
+                      >
+                        the docs
+                      </a>{' '}
+                      for more information on this.
```
dergigi, line 356:
```suggestion
                      for more information.
```

dergigi (commented):
Nice! Looks good! Two minor corrections 📝

src/components/Jam.jsx
```diff
@@ -246,6 +316,56 @@ export default function Jam() {
...
+                    blocks.
+                  </Trans>
+                ) : schedulerPreconditionSummary.amountOfMissingConfirmations > 0 ? (
+                  <Trans i18nKey="scheduler.precondition.hint_missing_confirmations">
+                    The scheduler requires one of your UTXOs to have{' '}
+                    <strong>
+                      {{
+                        /* this comment is a hack for "prettier" and prevents the removal of "{' '}" 
+                           (which is essential for parameterized translations to work). */
+                        minConfirmations: SCHEDULER_PRECONDITIONS.MIN_CONFIRMATIONS_OF_SINGLE_UTXO,
```
dnlggr, line 341: Hm yeah the <Trans> component thing is super annoying. If this works, I'm fine with it.

dnlggr (approved):
🆗 🆒 👍

dnlggr:
Might be worth noting here that all of these precondition checks only consider UTXOs in mixdepth 0 which is not reflected in the translations as of now. Just thinking of this now but I think it's still fine. We have mixdepth 0 as the default for funding addresses.

PR #264, opened 2022-05-13
fix: delay scheduler start/stop

### LINKED ISSUE #249: Delay starting / stopping scheduled transactions UI side

To avoid users starting/stopping the service in rapid succession which can lead to problems.

dnlggr:
Resolves #249.

I used the same delay (2000ms) as in the Earn screen. What do you think @theborakompanioni, is that long enough?

A bit off topic, but I also renamed all API methods to refer to "Scheduler" now.

### DIFF
--- a/src/components/Jam.jsx
+++ b/src/components/Jam.jsx
@@ -79,7 +79,7 @@ export default function Jam() {
 
   const reloadSchedule = useCallback(
     ({ signal }) => {
-      return Api.getTumblerSchedule({ walletName: wallet.name, token: wallet.token, signal })
+      return Api.getSchedule({ walletName: wallet.name, token: wallet.token, signal })
         .then((res) => (res.ok ? res.json() : Api.Helper.throwError(res)))
         .then((data) => {
           if (!signal.aborted) {
@@ -157,7 +157,7 @@ export default function Jam() {
     const body = { destination_addresses: destinations }
 
     if (process.env.NODE_ENV === 'development' && useInsecureTestingSettings && !deactivateTestingToggle) {
-      body.tumbler_options = {
+      body.scheduler_options = {
         addrcount: 3,
         minmakercount: 1,
         makercountrange: [1, 0],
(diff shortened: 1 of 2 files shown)


theborakompanioni:
> I used the same delay (2000ms) as in the Earn screen. What do you think @theborakompanioni, is that long enough?

It seems it will sufficiently keep users from starting/stopping the scheduler in rapid succession.
However, there might still be problems if the servers actual state is not taken into account.

This is for another PR, but what do you think about taking the same approach as in the Earn screen and wait for the server to signal the successful operation, ie. https://github.com/joinmarket-webui/joinmarket-webui/blob/a95ac7a6e419f52150a608e0b66c498240361db7/src/components/Earn.jsx#L156-L177

> A bit off topic, but I also renamed all API methods to refer to "Scheduler" now.

:+1:

src/components/Jam.jsx
```diff
@@ -157,7 +158,7 @@ export default function Jam() {
     const body = { destination_addresses: destinations }
 
     if (process.env.NODE_ENV === 'development' && useInsecureTestingSettings && !deactivateTestingToggle) {
-      body.tumbler_options = {
+      body.scheduler_options = {
```
theborakompanioni, line 161: Are you panning to update [your server branch](https://github.com/dnlggr/joinmarket-clientserver/tree/tumbler-rpc-options) or the [the PR](https://github.com/JoinMarket-Org/joinmarket-clientserver/blob/dd8e08ffb6ebdfe9f73307575079cd2123e51e4d/jmclient/jmclient/wallet_rpc.py#L1119) as well? Both still refer to `tumbler_options`.. would need to remember that till they are in sync again..

theborakompanioni (approved):

src/components/Jam.jsx
dnlggr: Oh yeah that needs to be fixed of course!

dnlggr:
> This is for another PR, but what do you think about taking the same approach as in the Earn screen and wait for the server to signal the successful operation, ie. 

Yeah that would definitely be great! Should we track it as an issue for the next version?

src/components/Jam.jsx
dnlggr: Fixed. Good catch! 👏

PR #265, opened 2022-05-13
fix: possible reference error in catch clause

theborakompanioni:
`err.response` is only populated by actual `JmApiErrors`, it might be `undefined`.
e.g. on aborts (`signal.aborted := true`)

This can be reproduced by artificially limiting your network connection speed, e.g. via Dev Tools (most browser should be able of doing that, I suppose).


dnlggr (approved):
Thanks! 🙏

PR #266, opened 2022-05-16
fix: do not prefill addresses after form validation fails

theborakompanioni:
Before this commit, whenever there was a typo in the fields for destination addresses, the form was populated with new wallet addresses, regardless of the "send to external wallet" toggle. 
After this commit, the form will initialize with empty addresses when "send to external wallet" is enabled.


src/components/Jam.jsx
```diff
@@ -40,14 +45,65 @@ export default function Jam() {
...
+        // on error initialize with empty addresses - form validation will do the rest
+        destinationAddresses = Array(addressCount).fill('')
+      }
+    }
+
+    const destinationAddressFormMap = destinationAddresses.reduce(
+      (obj, addr, index) => ({ ...obj, [`dest${index + 1}`]: addr }),
+      {}
+    )
+    return { ...destinationAddressFormMap }
```
dnlggr, line 105: I'd just return the result of `reduce` directly. Then we don't need to come up with a variable name and spread it out again the next line. :)

dnlggr (approved):
Works ✅ LGTM!

PR #270, opened 2022-05-16
ui: big red button for payments without improved privacy

theborakompanioni:
As the title says: Send button signals non-privacy payment choice as suggested in Figma.

## 📸 
 

edit: Images are outdated – instead of "Normal Send – …", it just says "Send – …".

### DIFF
--- a/src/i18n/locales/en/translation.json
+++ b/src/i18n/locales/en/translation.json
@@ -177,6 +177,7 @@
     "feedback_invalid_amount": "Please provide a valid amount.",
     "toggle_coinjoin": "Send as collaborative transaction for improved privacy",
     "button_send": "Send",
+    "button_send_without_improved_privacy": "Send – without improved privacy",
     "text_sending": "Sending",
     "button_sweep": "Sweep",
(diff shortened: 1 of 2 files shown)


src/i18n/locales/en/translation.json
```diff
@@ -177,6 +177,7 @@
     "feedback_invalid_amount": "Please provide a valid amount.",
     "toggle_coinjoin": "Send as collaborative transaction for improved privacy",
     "button_send": "Send",
+    "button_send_without_improved_privacy": "Send – without improved privacy",
```
dnlggr, line 180:
```suggestion
    "button_send_without_improved_privacy": "Send without privacy",
```

I'd just do this. Other suggestions @dergigi?

dergigi: I think we need the "improved" because "without privacy" isn't necessarily accurate 😅 

What about:

> "Send without privacy improvement"

MaxHillebrand (approved):

dergigi (approved):
Looks good to me ✅

PR #272, opened 2022-05-16
feat: prevent address reuse on Jam screen

### LINKED ISSUE #261: Prevent address reuse scheduled transactions
Currently, nothing is stopping a user from providing the same address to all input fields.
Would it be appropriate to prevent address reuse on the UI level?

theborakompanioni:
Closes #261

Text of error message is: "Reusing addresses has negative implications such as diminished privacy and reduced security. It is highly discouraged and therefore prevented."

Please add your suggestions here if you have a more appropriate phrase in mind.

### DIFF
--- a/src/i18n/locales/en/translation.json
+++ b/src/i18n/locales/en/translation.json
@@ -279,6 +279,6 @@
     "progress_current_state": "Waiting for transaction <1>{{ current }}</1> of <3>{{ total }}</3> to process...",
     "progress_done": "All transactions completed successfully. The scheduler will stop soon.",
     "error_loading_schedule_failed": "Loading schedule progress failed.",
-    "error_address_reuse": "Reusing addresses has negative implications such as diminished privacy and reduced security. It is highly discouraged and therefore prevented."
+    "error_address_reuse": "Reusing addresses is prohibited. Make sure to use unique and unused addresses."
   }
 }
--- a/src/i18n/locales/en/translation.json
+++ b/src/i18n/locales/en/translation.json
@@ -278,6 +278,7 @@
     "progress_description": "This time estimate does not take into account the time spent talking to counterparties and waiting for transactions to get confirmed on the time chain. It is a lower bound estimate. The actual time for processing the whole schedule will likely be longer.",
     "progress_current_state": "Waiting for transaction <1>{{ current }}</1> of <3>{{ total }}</3> to process...",
     "progress_done": "All transactions completed successfully. The scheduler will stop soon.",
-    "error_loading_schedule_failed": "Loading schedule progress failed."
+    "error_loading_schedule_failed": "Loading schedule progress failed.",
+    "error_address_reuse": "Reusing addresses has negative implications such as diminished privacy and reduced security. It is highly discouraged and therefore prevented."
   }
 }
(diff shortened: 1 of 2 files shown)


src/i18n/locales/en/translation.json
```diff
@@ -278,6 +278,7 @@
     "progress_description": "This time estimate does not take into account the time spent talking to counterparties and waiting for transactions to get confirmed on the time chain. It is a lower bound estimate. The actual time for processing the whole schedule will likely be longer.",
     "progress_current_state": "Waiting for transaction <1>{{ current }}</1> of <3>{{ total }}</3> to process...",
     "progress_done": "All transactions completed successfully. The scheduler will stop soon.",
-    "error_loading_schedule_failed": "Loading schedule progress failed."
+    "error_loading_schedule_failed": "Loading schedule progress failed.",
+    "error_address_reuse": "Reusing addresses has negative implications such as diminished privacy and reduced security. It is highly discouraged and therefore prevented."
```
dnlggr, line 282: Bad for privacy, yes! Bad for security, I'm not so sure.

Main reason for why we need this feature is that [amount correlation](https://github.com/JoinMarket-Org/joinmarket-clientserver/blob/master/docs/tumblerguide.md#basic-examples) will ruin any privacy gained by using the scheduler in the first place. 

Maybe just use something short like:

> "Reusing addresses is bad for privacy. Make sure to enter unique addresses."

I'd be great to educate the user on why this is needed though. But I don't know how to say "amount correlation attack" in a nice way. 😅  cc @dergigi

dnlggr (commented):
Nice feature! 👏 Just not sure about the wording yet.

src/i18n/locales/en/translation.json
dergigi: I'd go with:

> "Reusing addresses is prohibited. Make sure to use unique and unused addresses."

We should probably add a "Learn more" button that links to something that explains the why behind it at a later stage.

MaxHillebrand (approved):
Good addition!

Maybe this should be even impossible to send to same address in same tumbler...

src/i18n/locales/en/translation.json
theborakompanioni: > Bad for privacy, yes! Bad for security, I'm not so sure.

Tbh, got inspired by [Swan's support page](https://help.swanbitcoin.com/hc/en-us/articles/360055920494), I am obviously fine with any phrasing.

> I'd go with:
> 
> > "Reusing addresses is prohibited. Make sure to use unique and unused addresses."

Applied. :+1:

> We should probably add a "Learn more" button that links to something that explains the why behind it at a later stage.

If you think this is worthwhile – let's add an Issue for that!

theborakompanioni:
> Good addition!
> 
> Maybe this should be even impossible to send to same address in same tumbler...

Yes, it is currently really only applied to the actual from inputs, but checking all already known wallet addresses would be possible and actually would be the most Jam can do to prevent address reuse. (Obviously, it cannot be prevented in general. Users will always find a way). Would you open an Issue for that? Might be a "good first issue"! :muscle:

src/i18n/locales/en/translation.json
dnlggr: > Bad for privacy, yes! Bad for security, I'm not so sure.

I take that back after thinking about it again. See https://github.com/joinmarket-webui/joinmarket-webui/pull/277#discussion_r874597319.

PR #274, opened 2022-05-16
fix: handle new "total_balance" format

### LINKED ISSUE #271: fix: handle new "total_balance" format
Some changes got merged upstream that trigger errors in Jam.
Specifically: `total_balance` property of response data from `wallet/<wallet_name>/display` is not _always_ a number anymore and hence, cannot be parsed _sometimes_. 
It is either "<total_and_available_balance>" or "<available_balance> (<total_balance>)".

See PR: https://github.com/JoinMarket-Org/joinmarket-clientserver/commit/d08985043bf64a60bc379d26e3ebb597dde44e3b

Response *before* the commit:
```json
{
  "account": "0",
  "total_balance": "146.29621113",
 [...]
}
```

Response *after* the commit:
```json
{
  "account": "0",
  "total_balance": "143.82707535 (146.29621113)",
 [...]
}
```
```python
if available_balance != total_balance:
  return "{0:.08f} ({1:.08f})".format(available_balance, total_balance)
```

This is an unreleased sever version but we must take it into account, even if we ignore the newly provided information.

The Balance component errors with:
```
Uncaught TypeError: fractionalPart is undefined
    formatBtc Balance.jsx:31
    balanceComponent Balance.jsx:138
    Balance Balance.jsx:149
    React 12
```

### LINKED ISSUE #273: Show available balance on Jam screen
From a [TODO in code](https://github.com/joinmarket-webui/joinmarket-webui/blob/5764fc5e840e9b392aa813d95a8aa1dd521b5bd9/src/components/Jam.jsx#L267):
> // Todo: Subtract frozen or locked UTXOs from amount shown.

The Jam screen should show the balance used by scheduler and hence not show the total, but the available balance.

theborakompanioni:
Fixes #271.
Fixes #273.

This PR handles a recent format change in the backend api. Please read https://github.com/joinmarket-webui/joinmarket-webui/issues/271 for more information. 
New format handling is done in a backward compatible way, so both v0.9.6 and current master are now supported.

Before this commit, every component that needed a balance read the raw data of the `/wallet/<wallet_name>/display` response.
After this commit, the balance can be fetched from a hook `useBalanceSummary` which parses the raw data and prepares it for the component.

This can and should probably be externalized into a Context, as it will prepare a full "balance summary" for every component using the hook. For a quickfix, this should be good enough and is acceptable hopefully. (Could not find the an appropriate existing Context and did not want to introduce another one on such short notice)

If you use `useBalanceSummary` you get an object that looks like this:
```js
{ 
  totalBalance: "0.00000000",
  availableBalanceDontUseYet: "0.00000000",
  calculatedAvailableBalanceInSats: 0,
  calculatedFrozenOrLockedBalanceInSats: 0,
  accountBalances: [{
    accountIndex: 0,
    totalBalance: "0.00000000",
    availableBalanceDontUseYet: "0.00000000",
    calculatedAvailableBalanceInSats: 0,
    calculatedFrozenOrLockedBalanceInSats: 0,
  }, [...] ]
}
```

`availableBalanceDontUseYet` will contain the value reported by JM as "available balance", which should not be (and is not) used as of now and is only populated for testing. Also, this value is currently "wrong" – see Adams comment here: https://github.com/JoinMarket-Org/joinmarket-clientserver/pull/986#discussion_r872604383

`calculatedAvailableBalanceInSats` and `calculatedFrozenOrLockedBalanceInSats` are populated with the logic from the "balanceBreakdown" formerly only done on the Send screen (and displayed only on sweeps). It has been moved to the `useBalanceSummary` hook and can now be used in any component. `calculatedAvailableBalanceInSats` is now used on Jam screen to fix #273! Both properties are already marked as `@deprecated` - but given the above, this might not really be true as it seems we must calc the actual available balance ourselves. What do you think?

Last but not least: The `AccountBalanceSummary` does not contain `BranchBalanceSummary`s, because they are needed only in component `DisplayAccounts`. For this case, the method to parse the newly formatted string  `parseTotalBalanceString` had to be exported – it is not pretty but it works. We already parse the Account balance, even if the component does not make use of it, also parsing all branches would be the bigger downside than exporting the method temporarily.


src/components/Send.jsx
```diff
@@ -502,7 +479,7 @@ export default function Send() {
                     <td>{t('send.sweep_amount_breakdown_total_balance')}</td>
                     <td className={styles['balance-col']}>
                       <Balance
-                        valueString={breakdown.totalBalance.toString()}
+                        valueString={accountBalanceOrNull.totalBalance}
```
dnlggr, line 482: It's a bit confusing maybe that `totalBalance` is a string and `calculatedFrozenOrLockedBalanceInSats` below is a number. Would it be easy to make all these balances the same type? If not then it's probably not a big deal if we just keep it as is.

src/components/DisplayAccounts.jsx
```diff
@@ -35,70 +36,89 @@ export default function DisplayAccounts({ accounts, ...props }) {
   const { t } = useTranslation()
   const settings = useSettings()
 
+  const parseTotalBalanceFromRawStringOrEmpty = (rawBalanceString, source) => {
+    try {
+      return parseTotalBalanceString(rawBalanceString).totalBalance
+    } catch (e) {
+      console.warn(`cannot determine balance format for ${source}`)
+      return ''
```
dnlggr, line 44: Returning `''` would result in no balance being shown for this account/branch/address right?

src/hooks/BalanceSummary.ts
```diff
@@ -0,0 +1,140 @@
+import { useMemo } from 'react'
+import { btcToSats } from '../utils'
+import { useCurrentWalletInfo, BalanceString, Utxos } from '../context/WalletContext'
+
+interface BalanceSummary {
+  totalBalance: string | null
+  /**
+   * @description available balance (total - frozen - locked); this value is incorrect for backend versions <= 0.9.6
+   */
+  availableBalanceDontUseYet: string | null
```
dnlggr, line 10: So if I understand it correctly, once we switch to 0.9.7 we will use this value only and remove the custom calculation completely, right?

src/hooks/BalanceSummary.ts
```diff
@@ -0,0 +1,140 @@
...
+    const utxos = currentWalletInfo.data.utxos.utxos
+
+    try {
+      const walletBalanceSummary = parseTotalBalanceString(rawTotalBalance)
+      const utxosByAccount = utxos.reduce((acc, utxo) => {
+        const key = `${utxo.mixdepth}`
+        acc[key] = acc[key] || []
+        acc[key].push(utxo)
+        return acc
+      }, {} as { [key: string]: Utxos })
```
dnlggr, line 97: Do we need this (as `calculateFrozenOrLockedBalance` will already filter for UTXOs of the respective account anyway)?

dnlggr (commented):
Whoo good job, sorry you have to go through all this hassle. 😅 

It seems like it works correctly from my tests. I almost feel this would be a good candidate to add some unit tests for, what do you think?

src/hooks/BalanceSummary.ts
theborakompanioni: Sadly, not quite yet. [As hinted by Adam](https://github.com/JoinMarket-Org/joinmarket-clientserver/pull/986#issuecomment-1128681700), this is still subject to change. `availableBalanceDontUseYet` and `calculatedAvailableBalanceInSats` can still differ, whereby the calculated value is "more correct" and needs to be used even after v0.9.7, unless the server handling is adapted.

So in short: It is still needed at the time of writing.

dnlggr: Ok I see. From a gut feeling I'd say let's remove the deprecation flags for the calculated fields and also remove the `availableBalanceDontUseYet` field and fully rely on our own calculations then. We don't really know when (or if) the backend will implement this correctly I'd say we're better off (for now at least) not worrying about it and do it ourselves. What do you think?

src/components/Send.jsx
theborakompanioni: Yes, it is a bit confusing, I am sorry.
No, unfortunately this is not that easy or better put: It would need much additional work and _testing_. I refrained from doing so as this should just be fixing the format change issue without touching too much other code.

Is it okay to handle the string/number and representation issue in a tailored issue?
It needs improvements in general, e.g. the methods `btcToSats` and `satsToBtc` have floating point issues.

e.g.
```ts
const satsToBtc = (value: string) => parseInt(value, 10) / 100000000

> satsToBtc("1")
> // outputs: 1e-8
```

src/hooks/BalanceSummary.ts
theborakompanioni: You are right in the sense that there is some redundancy. But it is rather the other way around, `calculateFrozenOrLockedBalance` does not need to filter anymore as it is already provided with filtered utxos.
Thank you – will simplify! :+1:

src/components/Send.jsx
dnlggr: 👍 Yeah let's do it like this then!

src/components/DisplayAccounts.jsx
theborakompanioni: Yes it would look like this:


Instead of this:


The only case this can occur, if I am not mistaken, is, if the format has changed again. Which hopefully does not happen, but better safe than sorry. FYI, `amount` value in the `branches[]->entries[]` object does not change and is still a btc value represented as string, e.g. `{ amount: "150.00000000", [...] }`. Hope that also does not change (otherwise the UI will break again as this is not handled).

dnlggr:
With #1285 probably landing soon, this probably needs to be re-thought right @theborakompanioni?

theborakompanioni:
> With [#1285](https://github.com/JoinMarket-Org/joinmarket-clientserver/pull/1285) probably landing soon, this probably needs to be re-thought right @theborakompanioni?

Since the value from any new format (previously in the string, after [#1285](https://github.com/JoinMarket-Org/joinmarket-clientserver/pull/1285) in a distinct `available_balance`) will be unused for now anyway, this can be merged regardless.
I'd merge it and make adaptions once the upstream PR lands, is that okay?

dnlggr:
> I'd merge it and make adaptions once the upstream PR lands, is that okay?

Yeah sure. Just need some tests and then it can go in. Let me know if I should help you out.

src/hooks/BalanceSummary.test.tsx
```diff
@@ -0,0 +1,173 @@
...
+        data,
+        setup({
+          data: {
+            utxos: {
+              utxos: [],
+            },
+            display: {
+              walletinfo: {
+                wallet_name: 'test.jmdat',
+                total_balance: '1.00000001',
```
dnlggr, line 53: > it('should handle new format', () => {

That's the old format as well right?

dnlggr (approved):
Cool thanks for adding tests. If you want to add more feel free to do so and just merge whenever you think it is ready to be merged.

src/hooks/BalanceSummary.test.tsx
theborakompanioni: > Cool thanks for adding tests. If you want to add more feel free to do so and just merge whenever you think it is ready to be merged.

Yes, I want to add more tests. It is not quite ready yet. Will propably merge tomorrow.

> That's the old format as well right?

Yes, copy paste error, thanks.

theborakompanioni:
Still waiting for https://github.com/JoinMarket-Org/joinmarket-clientserver/pull/1285 to be integrated before this will be merged.

Issue #256, opened 2022-05-11
Privacy Level dropdown does not respect bitcoin/sats setting


dergigi:
Will always show the amount in bitcoin

![Screenshot 2022-05-11 at 14 45 58](https://user-images.githubusercontent.com/109058/167853098-98a15d7d-6e33-4841-b87c-208820ac2ac3.png)

secondl1ght:
I put a PR #276 for this

PR #276, opened 2022-05-17
feat: add sat conversion to send screen privacy levels based on settings

secondl1ght:
Requested in issue #256, I have added a sat conversion to the Send screen Privacy Levels dropdown menu that is dynamic to the Settings. This turned out to be a bit more tricky than I thought because the value needs to go inside the `option` element. Therefore you cannot use the `Balance` component because it will show up in the browser as `object`, it seems you cannot put another element inside an `option`. So I wrote a new conditional to render the balance based on the Unit setting. For the same reason as above I also could not include the Sprite for the sat symbol because it was also showing up as `object`. So instead I just used the word 'sats', I know it is not consistent with the other components but maybe there is no other workaround? I moved a few formatting functions into the Utils file because we need to use them in multiple files now. There also wasn't formatting on the BTC amount or '*****' for a hidden balance before so I added that as well. 

Anyway I have been super busy lately so haven't had time to contribute in a while but let me know what you think of these changes and approach for this task!

Thanks

image


src/components/Send.jsx
```diff
@@ -640,7 +640,11 @@ export default function Send() {
                         {settings.useAdvancedWalletMode
                           ? t('send.account_selector_option_dev_mode', { number: account })
                           : t('send.account_selector_option', { number: account })}{' '}
-                        {settings.showBalance && `(\u20BF${balance})`}
+                        {settings.showBalance
+                          ? settings.unit === 'sats'
+                            ? `(${formatSats(balance * 100000000)} sats)`
+                            : `(\u20BF${formatBtc(balance)})`
+                          : '*****'}
```
dnlggr, line 647: Let's just remove that case for when the blance is hidden. Will look best if we just don't show anything there like before.

src/components/Send.jsx
```diff
@@ -640,7 +640,11 @@ export default function Send() {
                         {settings.useAdvancedWalletMode
                           ? t('send.account_selector_option_dev_mode', { number: account })
                           : t('send.account_selector_option', { number: account })}{' '}
-                        {settings.showBalance && `(\u20BF${balance})`}
+                        {settings.showBalance
+                          ? settings.unit === 'sats'
+                            ? `(${formatSats(balance * 100000000)} sats)`
```
dnlggr, line 645: We have a `btcToSats` function in `src/utils.ts`. It's already imported in this component, so you should be able to just do:

```suggestion
                            ? `(${formatSats(btcToSats(balance))} sats)`
```

dnlggr (commented):
Looks good! 👌 

Eventually we'll need a custom selector component so that we can render the `Balance` component (since that's not possible in a plain `<option>` element as you mentioned) but for now this is a good trade off.

I'm a bit wary of blindly assuming that `balance` is BTC but we did this before as well so it's fine to leave it like this in this PR.

See my two comments inline. They're easy to address. 

Good job! 👏

theborakompanioni:
Please wait for https://github.com/joinmarket-webui/joinmarket-webui/pull/274 to be merged before integrating this. I'll fix all merge conflicts afterwards :pray:

theborakompanioni (approved):

PR #277, opened 2022-05-17
feat: improve copy

dergigi:
Went over the main pages:

- Receive
- Jam
- Earn
- Send

### Screenshots 📸

01-receive
02-jam
02b-jam
03a-earn
03b-earn
03c-earn
04-send


src/components/Earn.jsx
```diff
@@ -258,6 +258,11 @@ export default function Earn() {
 
           {serviceInfoAlert && <rb.Alert variant={serviceInfoAlert.variant}>{serviceInfoAlert.message}</rb.Alert>}
 
+          {!serviceInfo?.coinjoinInProgress &&
+            !serviceInfo?.makerRunning &&
+            !isWaitingMakerStart &&
+            !isWaitingMakerStop && <p className="text-secondary">{t('earn.market_explainer')}</p>}
```
dnlggr, line 264:
```suggestion
            !isWaitingMakerStop && <p className="text-secondary mb-4">{t('earn.market_explainer')}</p>}
```

src/i18n/locales/en/translation.json
```diff
@@ -151,11 +151,11 @@
   },
   "send": {
     "loading": "Loading",
-    "title": "Send bitcoin",
-    "subtitle": "Collaborative transactions increase the privacy of yourself and others.",
+    "title": "Send",
+    "subtitle": "Use a collaborative transaction for increased security and privacy.",
```
dnlggr, line 155: What's the security aspect to it?

dnlggr (approved):
Awesome! 👏 ✅ 💬

src/i18n/locales/en/translation.json
dergigi: Two sides of the same coin, see [Lopp's List](https://github.com/jlopp/physical-bitcoin-attacks).

There is a reason why the *s* in `https` stands for "secure" even though it is technically "private" - I think we should slowly but surely get this into the heads of bitcoiners too.

dnlggr: Fair point! 👍

dergigi:
Added screenshots for easier review. Please have another look @dnlggr - nothing set in stone of course, but I tried to be as careful and consistent as possible with the copy.

I think pointing out that Jam allows you to "improve the privacy of yourself and others" is important. I also was careful to not mention "bitcoin" too often, since it is always confusing (the network? the currency? UTXOs? what?).

If you don't like the "security" mention we can remove it of course, but it is only mentioned once. I hope that we will be able to improve the wording and terminology over time everywhere, but I feel like this is already a big step in the right direction.

src/i18n/locales/en/translation.json
```diff
@@ -202,22 +203,23 @@
     "text_getting_address": "Getting new address"
   },
   "earn": {
-    "title": "Earn bitcoin",
-    "subtitle": "By making your bitcoin available for others, you help them improve their privacy and can also earn a yield.",
+    "title": "Earn",
+    "subtitle": "By providing liquidity to the market, you improve the privacy of yourself and others. Your sats always remain under your control and never leave your wallet.",
+    "market_explainer": "You can choose how much you want to charge for your offer. Note that you participate and compete in an open market—there is no guarantee that other market participants will take you up on your offer. Being chosen for a collaborative transaction can take some time. Low fees increase your chances.",
```
dnlggr, line 208: > Low fees increase your chances.

Low fees and fidelity bonds, but we can't say that (yet). 😄

dnlggr (approved):
Looks good! ✅ 👍 

> If you don't like the "security" mention we can remove it of course,

After thinking about it again, I do get why "security" makes sense. Especially since we're trying to work towards improved terminology in the longer term.

> I think pointing out that Jam allows you to "improve the privacy of yourself and others" is important.

💯

PR #283, opened 2022-05-18
feat: split up scheduler destiantion addresses over 3 mixdepths

### LINKED ISSUE #282: Split up internal scheduler destination addresses accross mixdepths
As suggested by @openoms in [the last community call](https://github.com/joinmarket-webui/joinmarket-webui/wiki/community-call-%2315), splitting up the destination addresses of the scheduler across three mixdepths is the safer route to go compared to having it be three addresses from a single mixdepth. This will make it harder for the user to consolidate the three UTXOs e.g. via a sweep of single mixdepth with the three UTXOs.

dnlggr:
Resolves #282.

Splits up the internal destination addresses of the scheduler over 3 mixdepths. This is to make it harder for a consolidation to happen later on (e.g. imagine all 3 UTXOs on one mixdepth and the user then sweeping this mixdepth).

### DIFF
--- a/src/components/Jam.jsx
+++ b/src/components/Jam.jsx
@@ -49,7 +49,7 @@ export default function Jam() {
   const [schedule, setSchedule] = useState(null)
 
   // Returns one fresh address for each requested mixdepth.
-  const getNewAddresses = useCallback(
+  const getNewAddressesForAccounts = useCallback(
     (mixdepths) => {
       if (mixdepths.length !== 3) {
         throw new Error('Can only handle 3 destination addresses for now.')
@@ -88,7 +88,7 @@ export default function Jam() {
     } else {
       try {
         // prefill with addresses marked as "new"
-        destinationAddresses = getNewAddresses(INTERNAL_DEST_ACCOUNTS)
+        destinationAddresses = getNewAddressesForAccounts(INTERNAL_DEST_ACCOUNTS)
       } catch (e) {
         // on error initialize with empty addresses - form validation will do the rest
         destinationAddresses = Array(addressCount).fill('')
@@ -96,7 +96,7 @@ export default function Jam() {
     }
 
     return destinationAddresses.reduce((obj, addr, index) => ({ ...obj, [`dest${index + 1}`]: addr }), {})
-  }, [destinationIsExternal, getNewAddresses])
+  }, [destinationIsExternal, getNewAddressesForAccounts])
 
   useEffect(() => {
     const abortCtrl = new AbortController()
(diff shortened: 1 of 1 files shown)


src/components/Jam.jsx
```diff
@@ -46,34 +48,28 @@ export default function Jam() {
   const [collaborativeOperationRunning, setCollaborativeOperationRunning] = useState(false)
   const [schedule, setSchedule] = useState(null)
 
+  // Returns one fresh address for each requested mixdepth.
   const getNewAddresses = useCallback(
```
theborakompanioni, line 52:
```suggestion
  const getUnusedAddressPerAccount = useCallback(
```

Small nit.

src/components/Jam.jsx
```diff
@@ -46,34 +48,28 @@ export default function Jam() {
...
-    (count, mixdepth) => {
+    (mixdepths) => {
+      if (mixdepths.length !== 3) {
+        throw new Error('Can only handle 3 destination addresses for now.')
+      }
       if (!walletInfo) {
         throw new Error('Wallet info is not available.')
       }
```
theborakompanioni, line 59: Hmm.. is this check _really_ necessary?
It feels not right.
e.g. what about:

```suggestion
      if (!walletInfo) {
        throw new Error('Wallet info is not available.')
      }
      if (mixdepths.length > walletInfo.data.display.walletinfo.accounts) {
        throw new Error('Can only provide one address per account.')
      }
      if (mixdepths.length <= 3) {
        throw new Error('Can only handle at least 3 destination addresses for now.')
      }
```

Also, nothing is stopping anybody from providing [0,0,0], which might be a better thing to check, right?

src/components/Jam.jsx
```diff
@@ -46,34 +48,28 @@ export default function Jam() {
...
 
-      externalBranch.entries.every((entry) => {
-        if (entry.status === 'new') {
-          newAddresses.push(entry.address)
-        }
+        const newEntry = externalBranch.entries.find((entry) => entry.status === 'new')
 
-        if (newAddresses.length >= count) {
-          return false
+        if (!newEntry) {
```
theborakompanioni, line 67: Just a small notice here. If the exception is thrown, the user interface behaves very weird. Maybe try it for yourself. This is not something that should ever happen, but maybe try to wrap all invocations in a try/catch block if it can be mitigated. What do you think?

theborakompanioni (approved):
LGTM. :rocket:

src/components/Jam.jsx
dnlggr: The reason why I put the check in is to avoid anyone thinking they can just do something like:

```
INTERNAL_DEST_ACCOUNTS = [0, 1, 2, 3]
```

because the UI doesn't support anything else than exactly 3 destination addresses yet. I didn't want to put in the work to make this fully dynamic quite yet, that's why I added this check. Basically an assert. But I agree it's not super clean in this place.

src/components/Jam.jsx
dnlggr: Updated it to `getNewAddressesForAccounts`.

src/components/Jam.jsx
dnlggr: Should we remove the check and rely on the comment that says `INTERNAL_DEST_ACCOUNTS` must have length 3?

PR #284, opened 2022-05-18
chore: enable schedule

### LINKED ISSUE #255: Remove scheduled transactions feature flag
Before releasing 0.0.6.

dnlggr:
Resolves #255.

Reused the feature flag for the "schedule testing" toggle.

The `process.env.NODE_ENV === 'development'` will be `false` in our production deployment right? 😅


src/components/Jam.jsx
```diff
@@ -187,7 +185,8 @@ export default function Jam() {
 
     const body = { destination_addresses: destinations }
 
-    if (process.env.NODE_ENV === 'development' && useInsecureTestingSettings && !deactivateTestingToggle) {
+    // Make sure schedule testing is really only used in dev mode.
```
theborakompanioni, line 188:
```suggestion
```

What do you think about making the "insecure testing settings" a feature flag that is enabled by default on `env := 'development'` instead? I feel like this is more appropriate and would do away with the "double check", which was very surprising to me. Also, it _should in theory_ be possible to run Jam production build with a regtest backend, right?

theborakompanioni:
> The `process.env.NODE_ENV === 'development'` will be `false` in our production deployment right? sweat_smile

Yes. It is `test` when running `npm test` and `production` in prod builds. You're fine :wink: 
> You cannot override NODE_ENV manually. This prevents developers from accidentally deploying a slow development build to production.

theborakompanioni:
Should [scheduler_options](https://github.com/joinmarket-webui/joinmarket-webui/blob/master/src/components/Jam.jsx#L191) be renamed to [tumbler_options](https://github.com/JoinMarket-Org/joinmarket-clientserver/blob/master/docs/api/wallet-rpc.yaml#L618) in this PR?

src/components/Jam.jsx
dnlggr: I had that initially but I wasn't sure if I liked it so I removed it again. My thinking was that this isn't a feature that will ever be turned on. It's more of a conditional thing that will always be on in dev and never in prod. 

What about adding a second "thing" to the feature flags code called "debug features" or something?

dnlggr:
> Should [scheduler_options](https://github.com/joinmarket-webui/joinmarket-webui/blob/master/src/components/Jam.jsx#L191) be renamed to [tumbler_options](https://github.com/JoinMarket-Org/joinmarket-clientserver/blob/master/docs/api/wallet-rpc.yaml#L618) in this PR?

No imo. We should avoid tumbler throughout the app as much as possible. We're already renaming it just before we send it off already here: [`JmWalletApi.ts`](https://github.com/joinmarket-webui/joinmarket-webui/blob/master/src/libs/JmWalletApi.ts#L286)

theborakompanioni:
> > Should [scheduler_options](https://github.com/joinmarket-webui/joinmarket-webui/blob/master/src/components/Jam.jsx#L191) be renamed to [tumbler_options](https://github.com/JoinMarket-Org/joinmarket-clientserver/blob/master/docs/api/wallet-rpc.yaml#L618) in this PR?
> 
> No imo. We should avoid calling it tumbler throughout the app as much as possible since that's just not really the most accurate wording for it. It's just a scheduler. We're already renaming it just before we send it off here: [`JmWalletApi.ts`](https://github.com/joinmarket-webui/joinmarket-webui/blob/master/src/libs/JmWalletApi.ts#L286)

Uff.. sorry for not getting that. However, this is _very_ surprising to say the least. It is even assigned as property to a variable named `body` and then not used as such? Maybe it is more obvious if it uses camelCase like `schedulerOptions` instead of snake_case `scheduler_options`? As far as I know it is tried to have the [JmWalletApi.ts](https://github.com/joinmarket-webui/joinmarket-webui/blob/master/src/libs/JmWalletApi.ts#L286) reflect the raw API, otherwise it is really tightly coupled to Jam.
I'd be very much in favor of changing this and not let such things slip in in the first place. I think it is a small detail with possible high consequences if it is not handled with discipline.

src/components/Jam.jsx
theborakompanioni: > I had that initially but I wasn't sure if I liked it so I removed it again. My thinking was that this isn't a feature that will ever be turned on. It's more of a conditional thing that will always be on in dev and never in prod.

That's a good argument. I understand :+1:

> What about adding a second "thing" to the feature flags code called "debug features" or something?

Nah, let's see if it comes up again. I dont want to create more work artificially. If something like this comes up again, we can consider doing this.

dnlggr:
> As far as I know it is tried to have the [JmWalletApi.ts](https://github.com/joinmarket-webui/joinmarket-webui/blob/master/src/libs/JmWalletApi.ts#L286) reflect the raw API,

Yeah I see what you mean. Reverted the property to `tumbler_options`. I guess I was getting ahead of myself hoping we could call it `scheduler_option` in the backend as well.

dnlggr:
Let me know what you think of `https://github.com/joinmarket-webui/joinmarket-webui/pull/284/commits/a814465d7ebd059d03ab7da09bf64b95296da6e6` @theborakompanioni. Not sure if it makes things better.

theborakompanioni (approved):
Nice :rocket:

PR #288, opened 2022-05-19
chore: enable cheatsheet

theborakompanioni:
Removes cheatsheet from feature flags, thereby enabling it for everyone.
Also fixed the link to the Jam page.

Did **not** change the order of the entries. But that can be done quickly. Thoughts?

### DIFF
--- a/src/components/App.jsx
+++ b/src/components/App.jsx
@@ -22,7 +22,6 @@ import { setSession, clearSession } from '../session'
 import Onboarding from './Onboarding'
 import Cheatsheet from './Cheatsheet'
 import { routes } from '../constants/routes'
-import { isFeatureEnabled } from '../constants/features'
 import packageInfo from '../../package.json'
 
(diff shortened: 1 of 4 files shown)


dnlggr:
Yeah let's change the order like we said. That would be IIRC:

1. Fund
2. Schedule
3. Lock (disabled)
4. Earn
5. Repeat

theborakompanioni:
> Yeah let's change the order like we said. That would be IIRC:
> 
>     1. Fund
> 
>     2. Schedule
> 
>     3. Lock (disabled)
> 
>     4. Earn
> 
>     5. Repeat

dnlggr (approved):
LGTM 🚀

editwentyone:
maybe this? I posted the rework in a different issue: https://www.figma.com/file/kfejZJFlwBywvLEnPEmJo1/?node-id=2850%3A67638

PR #299, opened 2022-05-27
Fetch config values only once per wallet session

### LINKED ISSUE #279: Fetch config values only once per wallet session
Config values cannot be changed currently via Jam, hence, it is not needed to fetch them every time.
e.g. on the Send page checks existence of `minimum_makers` on every render and `max_cj_fee_rel` and `max_cj_fee_abs` on request errors. If a users changes the values in the config file directly,  `jmwalletd` must restart anyway and a new wallet session will be started. Hence, once per wallet session (or initial rendering) is enough.

theborakompanioni:
Closes #279 

Before: Config values were fetched individually on every send page.
After: Config values will be loaded only once, then stored and reused for the entire wallet session.

Only the `Send` component currently needs a config value for form validation (`minimum_makers`).
Others are loaded conditionally in case of an error (`max_cj_fee_rel` and `max_cj_fee_abs`)).

There are also other potential worthwhile solutions.. this seems to be a very straightforward one. Let me know, what you think.

The config update are logged to console in development mode:
```js
service config updated Object { 
  POLICY: Object { minimum_makers: "1",  max_cj_fee_rel: "0.0003", max_cj_fee_abs: "300000" }
​​}
```

### DIFF
--- a/src/components/Send.jsx
+++ b/src/components/Send.jsx
@@ -8,7 +8,7 @@ import Sprite from './Sprite'
 import Balance from './Balance'
 import { useReloadCurrentWalletInfo, useCurrentWallet, useCurrentWalletInfo } from '../context/WalletContext'
 import { useServiceInfo, useReloadServiceInfo } from '../context/ServiceInfoContext'
-import { useLoadConfigValueIfAbsent } from '../context/ServiceConfigContext'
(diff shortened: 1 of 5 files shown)


dnlggr (approved):
Works! Nice! ✅

PR #301, opened 2022-05-31
ui: blur Send form if service is running

theborakompanioni:
Suggestions by @dnlggr : https://github.com/joinmarket-webui/joinmarket-webui/pull/300#issuecomment-1141940249

If the whole form should be blurred, apply the PR, if it is okay the way it is, just close it.


dnlggr (approved):
See [my comment](https://github.com/joinmarket-webui/joinmarket-webui/pull/300#issuecomment-1144842433) over at #300 for why I think it makes sense to merge this now. Let me know if you strongly disagree @editwentyone. We can always revert this and discuss here. 🙏

PR #304, opened 2022-06-01
ui: Earn screen figma alignment

theborakompanioni:
Changes:
- Add a "Simple/Advanced View" toggle at the top to individually switch into a more configurable view.
- Do not show any config options when not necessary. Results in: only the "start" button is visible in "Simple View".
- Center "abs/rel offer" toggle and add label for disabled state.
- Added prefix text "%" (rel offer) or "sat" symbol (abs offer; min amount) to input fields.

Differences to figma screen:
- Icons for "Simple/Advanced View" toggle are not on bitcoinicons.com and were taken from the "use normal mode"/"use developer mode" toggle on the settings screen.
- The "Simple/Advanced View" toggle is placed _above_ the page title. This is because the `PageTitle` component is not customizable yet. That will be done in a subsequent PR. Hope that is okay for now.
- Unit cannot be changed yet and is always in "sats".

Small refactoring:
- `ToggleSwitch` from `.jsx` to `.tsx`.
- Moved all Earn component related styles from `index.css` to `Earn.module.css`.

## 📸
 

(screenshots slightly outdated: font size of input prefix (with "%") is increased to `1.2rem` to match the sat symbol size)


dnlggr:
Screenshot 2022-06-02 at 15 51 07

Thoughts on using an "Advanced" collapsible thing similar to how we have it on the Receive page? I personally feel like this would make it a bit easier / more modular.

dnlggr:
Screenshot 2022-06-02 at 15 54 16

I like this addition! 🙌

theborakompanioni:
> I like this addition!

:rocket: 

> Thoughts on using an "Advanced" collapsible thing similar to how we have it on the Receive page? I personally feel like this would make it a bit easier / more modular.

Okay with me! I like your suggestion as it will actually make it "harder" to do something complex. As any "Advanced View" should nonetheless strive to  be as simple as possible, a dropdown area would constraint the design and probably leads to rethinking how and what should effectively be achieved.

What's your take @editwentyone?

dnlggr:
Screenshot 2022-06-02 at 16 18 16

One more thing: I've never seen a switch used like this before. It somehow feels a bit off to me. In my experiences switches are best used for settings that can be toggled on and off (where the setting is described on the left off the switch). E.g. "Keep funds internal -- On/Off", "Use testing mode -- On/Off". This case is a bit different. We're basically switching between two subviews here.

Should we use some other UI element for switching between absolute offer and relative offer configurations? I feel like a segmented control like the following would be a good choice:

Screenshot 2022-06-02 at 16 22 18

Any thoughts on this?

editwentyone:
@dnlggr you are absolutely right, it should be a tab-switcher. I will add / change that 

@theborakompanioni could we move advanced/simple view on the same height as the headline i.e.: "earn"?

theborakompanioni:
Converted to draft. Will update with a tab-switcher and improved "Advanced View" action.

theborakompanioni:
Updated:
 


What do you think?

dnlggr (commented):
Works great! Awesome job @theborakompanioni! 👏 

I have some styling suggestions for the new UI element I'd love to hear some opinions on. What would you think of something like this:

**Light theme**

Screenshot 2022-06-06 at 15 51 04

**Dark theme**
Screenshot 2022-06-06 at 15 52 38

Basically makes the control element full width and narrower. The tabs equal width with a lighter color on dark mode and a small drop shadow. Kind of resembles what people already know from mobile (similar to how we reused the toggle switch from mobile).

The relevant changes are:

``` diff
diff --git a/src/components/RadioTabs.module.css b/src/components/RadioTabs.module.css
index 7b9597f..4a514ae 100644
--- a/src/components/RadioTabs.module.css
+++ b/src/components/RadioTabs.module.css
@@ -2,6 +2,7 @@
   background-color: var(--bs-gray-300);
   border-radius: 0.25rem;
   padding: 0.25rem;
+  width: 100%;
 }
 
 :root[data-theme='dark'] .radio-tabs {
@@ -9,15 +10,21 @@
   background-color: var(--bs-gray-800);
 }
 
+.radio-tab {
+  flex: 1 1 0;
+}
+
 .radio-tab input[type='radio'] {
   display: none;
 }
 
 .radio-tab label {
   background-color: white;
-  padding: 0.5rem 1rem;
+  padding: 0.25rem 1rem;
   border-radius: 0.25rem;
   font-weight: 500;
+  width: 100%;
+  text-align: center;
 }
 
 .radio-tab input[type='radio']:disabled ~ label {
@@ -41,8 +48,9 @@
 
 .radio-tab input[type='radio']:checked:not(:disabled) ~ label {
   background-color: var(--bs-gray-100);
+  box-shadow: 1px 1px 3px 1px rgba(0, 0, 0, 0.1);
 }
 
 :root[data-theme='dark'] .radio-tab input[type='radio']:checked:not(:disabled) ~ label {
-  color: var(--bs-gray-900);
+  background-color: var(--bs-gray-600);
 }

```

src/components/RadioTabs.module.css
```diff
@@ -0,0 +1,48 @@
...
+}
+
+.radio-tab input[type='radio'] {
+  display: none;
+}
+
+.radio-tab label {
+  background-color: white;
+  padding: 0.5rem 1rem;
+  border-radius: 0.25rem;
```
dnlggr, line 19: This is obviously a nit since it's such a small value here -- but in general, nested border radii should always be smaller than the parent border radii. See [here](https://css-tricks.com/public-service-announcement-careful-with-your-nested-border-radii/) for more details.

Normally a good [rule of thumb](http://web.archive.org/web/20210506121037/https://joshnh.com/weblog/getting-inner-border-radius-just-right/) is that the inner radius should be the outer radius minus the padding between the two elements. This doesn't work since 0.25 (outer radius) minus 0.25 (padding) equals 0. So I'd just go with something like 0.15. It'll look fine as well:

**Before:**
Screenshot 2022-06-06 at 15 59 45

Look kinda off right?

**After:**
Screenshot 2022-06-06 at 15 59 36

editwentyone: love the eye for details :D

editwentyone:
> > I like this addition!
> 
> 🚀
> 
> > Thoughts on using an "Advanced" collapsible thing similar to how we have it on the Receive page? I personally feel like this would make it a bit easier / more modular.
> 
> Okay with me! I like your suggestion as it will actually make it "harder" to do something complex. As any "Advanced View" should nonetheless strive to be as simple as possible, a dropdown area would constraint the design and probably leads to rethinking how and what should effectively be achieved.
> 
> What's your take @editwentyone?

I'm ok with this view for now, but we should make it consistent and the idea with advanced/simple view concept is still early. 
lets finish feature parity first and concentrate on UI later :)

dnlggr:
Pushed my feedback as discussed @theborakompanioni. Again, great job.

PR #305, opened 2022-06-01
feat: enable report overlay

theborakompanioni:
Before this commit, only the "developer mode" had the ability to open the earn report.
After this commit, the earn report button will always be shown, hence enabling it for all users by default.


dnlggr (approved):
LGTM! ✅

PR #307, opened 2022-06-06
feat: Fidelity Bonds

theborakompanioni:
TBD.


src/components/fidelity_bond/AccountSelector.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+  const selectableAccounts = useMemo(() => {
+    return accounts.filter((it) => !it.disabled)
+  }, [accounts])
+
+  const totalAmount = useMemo(() => {
+    return accounts
+      .map((it) => it.utxos.reduce((acc, curr) => acc + curr.value, 0))
+      .reduce((acc, curr) => acc + curr, 0)
+  }, [accounts])
```
dnlggr, line 32: Is there a benefit of recalculating this vs. taking it from `currentWallet`?

src/components/fidelity_bond/FidelityBondDetailsSetupForm.tsx
```diff
@@ -0,0 +1,582 @@
...
+  type Milliseconds = number
+  type UnitKey = 'year' | 'month' | 'day' | 'hour' | 'minute' | 'second'
+
+  type Units = {
+    [key in UnitKey]: Milliseconds
+  }
+
+  const units: Units = {
+    year: 24 * 60 * 60 * 1_000 * 365,
+    month: (24 * 60 * 60 * 1_000 * 365) / 12,
```
dnlggr, line 34: A month is not always (actually never) 365 / 12 days long. A year might not be 365 days long. Does that matter for how this util is used? I'm guessing this is used for rough user feedback only right?

src/components/fidelity_bond/FidelityBondDetailsSetupForm.tsx
```diff
@@ -0,0 +1,582 @@
...
+    return accounts.filter((it) => accountsWithUtxos.includes(it.account))
+  }, [utxosByAccount, accounts])
+
+  const selectableAccounts = useMemo(() => {
+    return accounts.map((it) => ({
+      ...it,
+      disabled: !availableAccounts.includes(it),
+      utxos: utxosByAccount[it.account] || [],
+    }))
+  }, [accounts, availableAccounts, utxosByAccount])
```
dnlggr, line 91: Is there ever an account that's available but not selectable? If not, I'd say let's not differentiate those two concepts and just have a single list of accounts including all info necessary.

src/components/fidelity_bond/FidelityBondDetailsSetupForm.tsx
```diff
@@ -0,0 +1,582 @@
...
+      <h4>Select Account</h4>
+      {availableAccounts.length === 0 ? (
+        <>
+          <Link to={routes.receive} className="unstyled">
+            <rb.Alert variant="info" className="mb-4">
+              <rb.Row className="align-items-center">
+                <rb.Col>
+                  <>
+                    No suitable account available. Fund your wallet and run the scheduler, before you create a Fidelity
+                    Bond.
```
dnlggr, line 111: We can't really detect if the user has actually run the scheduler right?

src/components/fidelity_bond/FidelityBondDetailsSetupForm.tsx
```diff
@@ -0,0 +1,582 @@
...
+}
+const SelectLockdateStep = ({ utxos, onChange }: SelectLockdateStepProps) => {
+  const { i18n } = useTranslation()
+  const settings = useSettings()
+
+  const [lockdate, setLockdate] = useState<Api.Lockdate | null>(null)
+  const yearsRange = useMemo(() => {
+    if (process.env.NODE_ENV === 'development') {
+      // allow creating lockdate in the past in dev mode
+      return toYearsRange(-1, DEFAULT_MAX_TIMELOCK_YEARS)
```
dnlggr, line 227: That's what I meant above. Do we still need the development mode now assuming we also add an Advanced collapsible showing the address (logging to console for now is fine as well). 😄

src/components/fidelity_bond/LockdateForm.tsx
```diff
@@ -0,0 +1,136 @@
...
+  }, [lockdateYear, lockdateMonth, isLockdateYearValid, isLockdateMonthValid, onChange])
+
+  return (
+    <rb.Row>
+      <rb.Col xs={6}>
+        <rb.Form.Group className="mb-4" controlId="locktimeYear">
+          <rb.Form.Label form="fidelity-bond-form">
+            <Trans i18nKey="fidelity_bond.form_create.label_locktime_year">Year</Trans>
+          </rb.Form.Label>
+          <rb.Form.Control
```
dnlggr, line 90: Let's use dropdowns here. We can't use the default date picker since Firefox and Safari won't support that (see #221) but 10 years and 12 months are an ok enough range for a dropdown.

dnlggr (commented):
It tested it a couple of times and it worked nicely for me. I left some comments inline but I have to say it is almost impossible for me to do a thorough code review of a PR this large in a reasonable amount of time.

The main, general questions I have are:

- We're currently using a sweep to send to the time-locked address which is good. There's still some open questions on how to handle cases where we don't want to use all UTXOs from a single jar (in which case we need to freeze some UTXOs and unfreeze them later on). The unfreezing is a bit tricky as we already talked about. For a v1 we could also think about just allowing users to send the entire contents of a jar if that makes things easier, similar to the [prototype on Figma](https://www.figma.com/proto/kfejZJFlwBywvLEnPEmJo1/JoinMarket-UI?page-id=2850%3A67638&node-id=2850%3A71655&viewport=48%2C-1032%2C0.23&scaling=min-zoom&starting-point-node-id=2850%3A71655&show-proto-sidebar=1).
- Privacy-wise, does it make a difference if the UTXO used is the output of a CJ done previously (A -> CJ -> B -> FB) or if we use a CJ directly to fund the fidelity bond (A -> CJ -> FB)?
- Can we remove the separate dev mode? 😄 
-

src/components/FidelityBond.tsx
```diff
@@ -0,0 +1,468 @@
...
+    setWaitForTakerToFinish(isCoinjoinInProgress)
+  }, [isCreating, isCreateSuccess, isCreateError, isCoinjoinInProgress])
+
+  const utxos = useMemo(
+    () => (currentWalletInfo === null ? [] : currentWalletInfo.data.utxos.utxos),
+    [currentWalletInfo]
+  )
+  const fidelityBonds = useMemo(() => (utxos === null ? null : utxos.filter((utxo) => utxo.locktime)), [utxos])
+  const activeFidelityBonds = useMemo(
+    () => (fidelityBonds === null ? null : fidelityBonds.filter((utxo) => utxo.frozen)),
```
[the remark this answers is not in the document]
theborakompanioni: No.

src/components/fidelity_bond/FidelityBondDetailsSetupForm.tsx
theborakompanioni: > A month is not always (actually never) 365 / 12 days long. A year might not be 365 days long. Does that matter for how this util is used? I'm guessing this is used for rough user feedback only right?

I will add tests for this to make it more clear. Short answer: Should not matter here.
But you are right that I did not take into account what happens on the last day of a month.
Q: Should a "normal" user be able to create a FB that is only valid a few hours, minutes eventually?
It feels like a mistake as the bond will not help much in being chosen more often, but then again, someone might have valid reasons to do so (?)

src/components/fidelity_bond/FidelityBondDetailsSetupForm.tsx
theborakompanioni: Yes. You should have experienced this by yourself in regtest: At first, only Jar Zero contains utxos and it is the only one "selectable".
What is your suggestion: Should only Jar Zero be shown? Or should all be "selectable" and "No utxos" phrase shown in the next step?

src/components/FidelityBond.tsx
```diff
@@ -0,0 +1,468 @@
...
+      const coinControlSetup = createCoinControlSetup(currentWalletInfo, selectedUtxos)
+
+      console.info(
+        `Fidelity Bond Setup: Freezing ${coinControlSetup.freeze.length} utxos, unfreezing ${coinControlSetup.unfreeze.length} utxos`
+      )
+
+      try {
+        const frozenUtxoIds = await prepareUtxosForSweep(requestContext, coinControlSetup)
+
+        // TODO: consider storing utxo id hashes in local storage..
```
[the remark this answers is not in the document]
dnlggr: Thinking about it again, maybe let's also try to make it super simple first and maybe only let them chose whole jars and then revisit this topic to be able to put a better focus onto the nuances here.

src/components/fidelity_bond/FidelityBondDetailsSetupForm.tsx
theborakompanioni: Yes. Gonna remove `DevOnly` component. :+1:

src/components/FidelityBond.tsx
```diff
@@ -0,0 +1,468 @@
...
+      const minimumMakers = await loadConfigValue({
+        signal: abortCtrl.signal,
+        key: { section: 'POLICY', field: 'minimum_makers' },
+      }).then((data) => parseInt(data.value, 10))
+
+      const coinControlSetup = createCoinControlSetup(currentWalletInfo, selectedUtxos)
+
+      console.info(
+        `Fidelity Bond Setup: Freezing ${coinControlSetup.freeze.length} utxos, unfreezing ${coinControlSetup.unfreeze.length} utxos`
+      )
```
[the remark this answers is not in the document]
theborakompanioni: Can do! Just for curiosity, is it discouraged to log info messages to console?

src/components/FidelityBond.tsx
```diff
@@ -0,0 +1,468 @@
...
+  }
+  if (!waitForTakerToFinish && isCoinjoinInProgress) {
+    return <>Creating Fidelity Bonds is temporarily disabled: A collaborative transaction is in progress.</>
+  }
+
+  return (
+    <div className={styles['fidelity-bond']}>
+      <PageTitle title={t('fidelity_bond.title')} subtitle={t('fidelity_bond.subtitle')} />
+
+      {featureAdvancedEnabled && (
```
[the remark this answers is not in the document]
theborakompanioni: `DevOnly` was previously named `Advanced` as the first intention was to allow expert users to retrieve the actual timelocked address. Has been removed and is not possible anymore.

src/components/fidelity_bond/AccountSelector.tsx
```diff
@@ -0,0 +1,87 @@
...
+    [selected]
+  )
+
+  const addOrRemove = useCallback(
+    (account: Account) => {
+      setSelected((current) => {
+        if (isSelected(account)) {
+          return current.filter((it) => it !== account)
+        }
+        return type === 'checkbox' ? [...current, account] : [account]
```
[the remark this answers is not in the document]
theborakompanioni: Removed :+1:
Maybe I was a little bit too optimistic about reusability :sweat_smile:

src/components/fidelity_bond/AccountSelector.tsx
theborakompanioni: I don't know exactly what you mean. The actual total value in sats is not stored anywhere. I was about to add a total value in sats to the `useBalance` hook (think there is a TODO in the comments iirc).
But in this case, the accounts can be provided and it felt wrong to bring in a total balance when it can easily be calculated from the input values. Furthermore, total value in btc is in `WalletDisplayInfo`.. however, the current `satsToBtc` util function is not to be trusted.  It parses a float string and does rounding.. nothing good can ever come from that. It is all good and well when the outputs are used for displaying stuff (notice the `Display` in the name).. For Fidelity Bonds, I did not want to take the chance.

Or maybe I misunderstood and you had something completely different in mind. Can you elaborate?

src/components/FidelityBond.tsx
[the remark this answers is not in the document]
theborakompanioni: Source: https://github.com/JoinMarket-Org/joinmarket-clientserver/blob/master/docs/fidelity-bonds.md#spending-time-locked-coins

src/components/fidelity_bond/AccountSelector.tsx
dnlggr: Yeah I meant `WalletDisplayInfo`. I mean it doesn't really matter here probably but it's a red flag that goes off in my head when there's two different sources of some value since then there's the (at least theoretical) possibility of  stuff getting out of sync. That's why I raised it but you can best judge if it's ok to do it like this or somehow get access to `WalletDisplayInfo`. Up to you! 😄

dergigi:
Awesome work @theborakompanioni! Will do my best to review and test this today.

dnlggr (commented):
Awesome, great starting point to further build this out and make it look pretty. ✨ 

Obviously we're going to need some sort of UTXO selection mechanism in the future. My biggest questions right now are around the collaborative transaction used to fund the fidelity bond:
- Is it enough to make a fidelity bond private privacy?
- Do we need to make sure users run the schedule first and fund the fidelity bond with an already private UTXO (as noted in the cheatsheet)?
- If so, do we even need the collaborative transaction to fund the fidelity bond?

src/components/FidelityBond.tsx
```diff
@@ -0,0 +1,268 @@
...
+  counterparties: number
+): Promise<void> => {
+  const amount_sats = 0 // sweep
+
+  await Api.postCoinjoin(requestContext, {
+    mixdepth: parseInt(account.account, 10),
+    destination: timelockedDestinationAddress,
+    amount_sats,
+    counterparties,
+  }).then((res) => (res.ok ? true : Api.Helper.throwError(res)))
```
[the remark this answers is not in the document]
theborakompanioni: No.. I did it again.. damnit. Fill be fixed. Sorry : /

src/components/fidelity_bond/AccountSelector.tsx
```diff
@@ -0,0 +1,103 @@
...
+    <rb.Row xs={1} className="gap-2">
+      {accounts.map((it) => {
+        if (!displayDisabledAccounts && it.disabled) {
+          return <></>
+        }
+
+        const accountIndex = parseInt(it.account, 10)
+        const availableAccountBalance = balanceSummary.accountBalances
+          .filter((balance) => balance.accountIndex === accountIndex)
+          .reduce((acc, curr) => acc + curr.calculatedAvailableBalanceInSats, 0)
```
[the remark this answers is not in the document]
theborakompanioni: You are right. This annoyed me as well. Will rework `accountBalances` to be an object with `accountIndex` as keys instead of an array.

src/components/fidelity_bond/FidelityBondDetailsSetupForm.tsx
```diff
@@ -0,0 +1,304 @@
...
+    })
+      .then((res) => (res.ok ? res.json() : Api.Helper.throwError(res)))
+      .then((data) => data.address)
+      .then((timelockedAddress) => console.info(`Address with lockdate '${selectedLockdate}':`, timelockedAddress))
+      .catch((error) => console.warn(`Could not fetch address with lockdate '${selectedLockdate}':`, error))
+
+    return () => {
+      abortCtrl.abort()
+    }
+  }, [currentWallet, selectedLockdate])
```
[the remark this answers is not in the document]
theborakompanioni: The address will be fetched once the user actually wants to submit, as all the form elements do not know it is about getting a timelocked address. I thought this is okay as it is only done if the debug feature "logFidelityBondAddressToConsole" is enabled. Do you have a suggestion, _when_ it should be done? Should a in-between step be introduced?

src/components/fidelity_bond/LockdateForm.tsx
```diff
@@ -0,0 +1,144 @@
...
+          <rb.Form.Label form="fidelity-bond-form">
+            <Trans i18nKey="fidelity_bond.form_create.label_lockdate_month">Month</Trans>
+          </rb.Form.Label>
+          <rb.Form.Select
+            defaultValue={initialMonth}
+            onChange={(e) => setLockdateMonth(parseInt(e.target.value, 10))}
+            required
+            isInvalid={!isLockdateMonthValid}
+          >
+            {[1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12].map((month) => (
```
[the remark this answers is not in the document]
theborakompanioni: Done!

src/components/fidelity_bond/AccountSelector.tsx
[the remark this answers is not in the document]
theborakompanioni: Done!

src/components/fidelity_bond/FidelityBondDetailsSetupForm.tsx
```diff
@@ -0,0 +1,304 @@
...
+
+const ConfirmationStep = ({ balanceSummary, account, lockdate, confirmed, onChange }: ConfirmationStepProps) => {
+  const { t } = useTranslation()
+  const settings = useSettings()
+
+  const accountAvailableBalanceInSats = useMemo(
+    () =>
+      balanceSummary.accountBalances
+        .filter((it) => it.accountIndex === parseInt(account.account, 10))
+        .reduce((acc, curr) => acc + curr.calculatedAvailableBalanceInSats, 0),
```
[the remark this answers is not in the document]
theborakompanioni: Done!

src/i18n/locales/en/translation.json
```diff
@@ -268,6 +272,17 @@
       "description": "Still confused? Dig into the <2>documentation</2>."
     }
   },
+  "fidelity_bond": {
+    "title": "Fidelity Bonds (Experimental)",
+    "subtitle": "Fidelity Bonds are a feature to improve the resistance to sybil attacks by deliberately locking up bitcoin value to make a cryptographic identity expensive to obtain. Therefore improving your privacy and the whole network. You will have a higher probability to take part in collaborative transaction and earn more yield.",
```
dergigi, line 277:
```suggestion
    "subtitle": "Fidelity Bonds prevent Sybil attacks by deliberately increasing the cost of creating cryptographic identities. Creating a fidelity bond will increase your chances of being picked for a collaborative transaction.",
```

src/components/FidelityBond.tsx
```diff
@@ -0,0 +1,272 @@
...
+        setIsInitializing(false)
+      })
+
+    return () => abortCtrl.abort()
+  }, [currentWallet, reloadCurrentWalletInfo, t])
+
+  /**
+   * Initiate sending funds to a timelocked address.
+   * Defaults to sweep with a collaborative transaction.
+   * If the selected utxo is a single expired FB, "diret-send" is used.
```
dergigi, line 140:
```suggestion
   * If the selected utxo is a single expired FB, "direct-send" is used.
```

src/components/FidelityBond.tsx
```diff
@@ -0,0 +1,268 @@
...
+
+      if (useDirectSend) {
+        await directSweepToFidelityBond(requestContext, selectedAccount, timelockedDestinationAddress)
+      } else {
+        const minimumMakers = await loadConfigValue({
+          signal: abortCtrl.signal,
+          key: { section: 'POLICY', field: 'minimum_makers' },
+        }).then((data) => parseInt(data.value, 10))
+
+        // TODO: how many counterparties to use? is "minimum" for fbs okay?
```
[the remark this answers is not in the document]
theborakompanioni: Yeah, you are right. Users should be in full control of this. Is it okay to introduce a step to select direct-send/collaborative-tx and the number of collaborators in a follow-up PR?

src/components/FidelityBond.tsx
```diff
@@ -0,0 +1,268 @@
...
+
+  const isCoinjoinInProgress = useMemo(() => serviceInfo && serviceInfo.coinjoinInProgress, [serviceInfo])
+  const isMakerRunning = useMemo(() => serviceInfo && serviceInfo.makerRunning, [serviceInfo])
+  const isOperationDisabled = useMemo(
+    () => isMakerRunning || isCoinjoinInProgress,
+    [isMakerRunning, isCoinjoinInProgress]
+  )
+
+  const [alert, setAlert] = useState<AlertWithMessage | null>(null)
+  const [isInitializing, setIsInitializing] = useState(true)
```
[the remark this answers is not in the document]
theborakompanioni: Good catch. Leftover from coin control. Removed!

dnlggr:
Ok let's merge this now and work do any updates in separate PRs. Unless you're still reviewing @dergigi? 🤔 

Imo I'd be nice to then:

- move the whole thing [to the Earn screen](https://www.figma.com/file/kfejZJFlwBywvLEnPEmJo1/JoinMarket-UI?node-id=2850%3A71655) as per Figma; and 
- make the collaborative transaction magic in the background more transparent to the user

src/components/fidelity_bond/FidelityBondDetailsSetupForm.tsx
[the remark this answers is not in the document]
theborakompanioni: Debug feature `logFidelityBondAddressToConsole` removed.

src/components/fidelity_bond/FidelityBondDetailsSetupForm.tsx
```diff
@@ -0,0 +1,304 @@
...
+    return () => {
+      abortCtrl.abort()
+    }
+  }, [currentWallet, selectedLockdate])
+
+  const _onSubmit = async (account: Account, lockdate: Api.Lockdate) => {
+    if (!currentWallet) return
+
+    const { name: walletName, token } = currentWallet
+    const timelockedDestinationAddress = await Api.getAddressTimelockNew({
```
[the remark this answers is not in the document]
theborakompanioni: Done.

src/hooks/BalanceSummary.ts
```diff
@@ -63,48 +94,60 @@ const useBalanceSummary = (currentWalletInfo: WalletInfo | null): WalletBalanceS
...
+          }
+          const accountTotalCalculated: AmountSats = totalCalculatedByAccount[account] || 0
+          const accountFrozenOrLockedCalculated: AmountSats = frozenOrLockedCalculatedByAccount[account] || 0
+          const accountAvailableCalculated: AmountSats = accountTotalCalculated - accountFrozenOrLockedCalculated
+          return {
+            ...accountBalanceSummary,
+            calculatedTotalBalanceInSats: accountTotalCalculated,
+            calculatedFrozenOrLockedBalanceInSats: accountFrozenOrLockedCalculated,
+            calculatedAvailableBalanceInSats: accountAvailableCalculated,
+            accountIndex: parseInt(account, 10),
```
dergigi, line 122: NIT: Any reason for the explicit `10` in all the `parseInt` calls? 2nd param could be removed, 10 is default afaik.

src/libs/JmWalletApi.ts
```diff
@@ -16,8 +16,12 @@ type ApiToken = string
 type WalletName = string
 
 type Mixdepth = number
-type AmountSats = BigInt
-type BitcoinAddress = string
+export type AmountSats = number // TODO: should be BigInt! Remove once every caller migrated to TypeScript.
```
dergigi, line 19: 👍

src/components/fidelity_bond/fb_utils.ts
```diff
@@ -0,0 +1,79 @@
...
+    throw new Error('Invalid values for range of years.')
+  }
+  return { min, max }
+}
+
+// A maximum of years for a timelock to be accepted.
+// This is useful in simple mode - when it should be prevented that users
+// lock up their coins for an awful amount of time by accident.
+// In "advanced" mode, this can be dropped or increased substantially.
+export const DEFAULT_MAX_TIMELOCK_YEARS = 10
```
dergigi, line 21: 10 years is quite a lot 😅  -- that being said, I don't know what a sane maximum should be. 10 years is probably good!

dergigi (approved):
Nice! Good job on all the tests, and on getting the first iteration through the door! 👏 Also: thank you @dnlggr for all the work on this in terms of sparring and review.

Tested locally on regtest, similarly to @theborakompanioni's demo in [Call 17](https://github.com/joinmarket-webui/joinmarket-webui/wiki/Community-Call-%2317).

I only have nits, thus this can be merged I think. Looks good to me ✅

dergigi:
> Ok let's merge this now and work do any updates in separate PRs.

Agree, also with next steps 👍 :shipit:

MaxHillebrand (approved):
This looks awesome.

src/hooks/BalanceSummary.ts
theborakompanioni: Over the years, I have made a habit of always indicating the radix to get reliable results.

> Be careful—this does not default to 10!
See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt

JavaScript is strange.

Edit: In this case, it would probably be fine, as we should not expect "malformed" inputs from JM.. nonetheless.. I guess I'll sleep better with the radix included. Is that okay for you?

src/components/fidelity_bond/fb_utils.ts
theborakompanioni: Yeah.. as far as I can see most fidelity bonds are not beyond the year 2026.. should we decrease to a maximum of 5 years and wait for user feedback?

dnlggr (approved):
Great stuff. Let's merge this, get feedback and improve it incrementally. 👏 👏 👏

src/hooks/BalanceSummary.ts
dergigi: > JavaScript is strange.

Ain't that the truth! 😅

PR #311, opened 2022-06-09
fix: mobile layout issues

dnlggr:
Fixes some smaller UI glitches on mobile. There's still some left, especially a scrolling issue where the main wallet view is scrolled down all the way on first load on a mobile phone. Couldn't figure out what causes this in a reasonable amount of time so leaving this alone for now.

Fixed:
- Footer layout on mobile.
- Page title font size
- Larger top padding for pages on mobile
- Cheatsheet height


src/components/Cheatsheet.module.css
```diff
@@ -1,12 +1,12 @@
 .cheatsheet {
   height: auto !important;
   /* page height - navbar height - some spacing*/
-  max-height: calc(100vh - 76px - 1rem) !important;
+  max-height: calc(100% - 4.75rem - 1rem) !important;
```
dnlggr, line 4: `100vh` causes issues on iOS since browser UI elements are part of the view port. Before, the cheatsheet would grow underneath the navbar. Now it doesn't anymore.

PR #313, opened 2022-06-09
chore: pre-alpha -> beta

dergigi:
Beta soon! (tm)

### DIFF
--- a/src/components/App.jsx
+++ b/src/components/App.jsx
@@ -34,7 +34,7 @@ export default function App() {
   const sessionConnectionError = useSessionConnectionError()
 
   const [websocketConnected, setWebsocketConnected] = useState()
-  const [showAlphaWarning, setShowBetaWarning] = useState(false)
+  const [showBetaWarning, setShowBetaWarning] = useState(false)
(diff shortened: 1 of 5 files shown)


dnlggr:
We're already on _alpha_. (See [here](https://github.com/joinmarket-webui/joinmarket-webui/blob/master/src/i18n/locales/en/translation.json#L17) for the current translation string). The hardcoded string here is just an artifact that the translation library needs. Its content's don't matter though. You can read about it [here](https://react.i18next.com/latest/trans-component) but I don't recommend it, its super confusing and annoying.

dnlggr:
Should we go straight to beta though? 😄

dergigi:
God damnit 😅

dergigi:
Let's rip out the i18n stuff for now and move straight to beta?

dnlggr:
No need for ripping, you can just update the translations string here: https://github.com/joinmarket-webui/joinmarket-webui/blob/master/src/i18n/locales/en/translation.json#L17

dergigi:
Should be good now, I think.

dnlggr (approved):
LGTM

PR #321, opened 2022-06-09
ui: tab dividers

dnlggr:
Adds dividers between the main tabs to hint towards the expected flow a user should/could take.

Trade off is that the tabs won't be of equal length anymore but that is ok imo.

Side note: Eventually we need an "official" color palette. Right now we're mixing colors from bootstrap and Figma screens.

## 📸 

Screenshot 2022-06-09 at 17 25 47

### DIFF
--- a/src/components/Navbar.jsx
+++ b/src/components/Navbar.jsx
@@ -51,7 +51,7 @@ const CenterNav = ({ makerRunning, cjRunning, onClick }) => {
           {t('navbar.tab_receive')}
         </NavLink>
       </rb.Nav.Item>
-      <div className="d-flex align-items-center center-nav-link-divider">»</div>
+      <div className="d-none d-md-flex align-items-center center-nav-link-divider">»</div>
       <rb.Nav.Item className="d-flex align-items-stretch">
         <NavLink
           to={routes.jam}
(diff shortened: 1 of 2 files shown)


theborakompanioni:
Looks a bit weird on small screens. Should we hide the dividers in the mobile menu?

dnlggr:
Very good point! 😄

editwentyone:
colours based on a real and consistent design system is something for the future. just go with the greys from bootstrap that look similar to figmas, everything else is time consuming right now

editwentyone (approved):

dnlggr:
> colours based on a real and consistent design system is something for the future. just go with the greys from bootstrap that look similar to figmas, everything else is time consuming right now

I agree but we shouldn't put it off for too long. The longer we wait, the more complicated it'll become later on and at some point it'll be too much work and never actually happen. We're probably still fine with waiting a bit but we need to keep it in mind.

editwentyone:
sure but there will/ should be a complete overhaul of the ui at some point and I think then its better to touch everything with a clearer design in mind

dnlggr:
Sure we'll have to go over the screens and make them consistent. This is where we can also take care of colors then. I'm thinking something like 0.1.0 the latest is where we'll want to to this i.e. after fidelity bonds and jar visuals are sorted out.

theborakompanioni (approved):

Issue #228, opened 2022-04-21
Make one wallet the standard ?!


editwentyone:
what do you think about sliming down the overall use of passwords for JAM and wallets by giving the user in the normal journey the following:

- password protected JAM app [preview](https://www.figma.com/file/kfejZJFlwBywvLEnPEmJo1/?node-id=3994%3A73345)
- automatically creating one wallet **without** password after intro (is this even possible with JM API?) (no preview, created screens wouldn't be needed anymore)
- and move "switch wallet" / "create another wallet"  into the settings for more advanced people. [preview](https://www.figma.com/file/kfejZJFlwBywvLEnPEmJo1/?node-id=3283%3A85955l)

later we could also maybe build in import/ export of wallets for different cross usage like joininbox if needed. 
the user would have an easier experience without too many passwords to remember/ secure and there is no lock/unlock issue with running processes in the background.

openoms:
On most systems the wallet password is the only encryption  standing between needing a password or having the bitcoin keys readily available for anyone with physical access to the disk.

Be aware that getting rid of wallet password there will be no security left.
A slightly better solution could be to reuse the wallet password to be the app password so then it can be stored in the memory to unlock the wallet also and disappears when the node is switched off.
See the raspiblitz LND wallet for example:
the apps are locked with the PasswordB (which is the bitcoin password reused)
the LND wallet has it's own PasswordC which is not stored by default and locks all funds on a restart.

Think of what happens when someone takes the node away. 
Passwords not stored (or only in RAM) are flushed and strong encryption saves the user from losing funds.

The problem with loosening the security is that JM is meant for higher amounts by nature so autounlocks and passwordless setups risk a lot potentially.

dnlggr:
I wouldn't be comfortable with implementing completely passwordless wallets for the reasons @openoms mentioned above. 

I do like the general idea of having a single wallet flow by default (with the option to create more wallets if needed). I feel like most of our users would be perfectly fine with having only one wallet. Currently the "choose wallet" screen is really prominent and I don't think it needs to be such a central thing in the UX when most users (I assume; might be wrong of course) have a single wallet only.

[...]

PR #325, opened 2022-06-14
feat: improve wallet control in settings

### LINKED ISSUE #323: Question: Lock wallet flow
UI flow to lock a wallet does not feel very natural.
Currently, you have to go to `Settings` > `Switch wallets` > `Lock wallet`.

Would it be okay to introduce a "Lock/Leave wallet" functionality directly on the Settings page?

dnlggr:
Resolves #323. Lays some groundwork for #228.

Adds actions for locking the wallet and creating a new wallet to the settings. We're not at a single wallet flow yet but this lays some groundwork for it and makes features available through the settings screen, which were previously accessible only on the wallets screen.

Also updates the UI of the modal used to warn the user before locking a wallet where a service is running.

## 📸 
Screenshot 2022-06-14 at 14 48 01
Screenshot 2022-06-14 at 14 48 21


src/components/Settings.jsx
```diff
@@ -97,10 +107,56 @@ export default function Settings() {
...
+        } else if (!res.ok) {
+          Api.Helper.throwError(res)
+        }
+
+        navigate(destination)
+      } catch (e) {
+        setAlert({ variant: 'danger', dismissible: false, message: e.message })
+      } finally {
+        setLockingWallet(false)
+      }
```
theborakompanioni, line 134:
```suggestion
        setLockingWallet(false)
        navigate(destination)
      } catch (e) {
        setLockingWallet(false)
        setAlert({ variant: 'danger', dismissible: false, message: e.message })
      }
```

Prevents `Can't perform a React state update on an unmounted component`.

src/components/Settings.jsx
```diff
@@ -97,10 +107,56 @@ export default function Settings() {
...
+
+      setLockingWallet(true)
+
+      try {
+        const { name: walletName, token } = currentWallet
+        const res = await Api.getWalletLock({ walletName, token })
+
+        if (res.ok || res.status === 401) {
+          stopWallet()
+        } else if (!res.ok) {
```
theborakompanioni, line 125:
```suggestion
        } else {
```
Is there any other possibility?

theborakompanioni:
> I refrained from using it here for such a light component with no real API and complex logic. Mostly because I feel like it adds a lot of complexity everywhere. But I won't die on that hill. I can update it if you feel like we should use it more consistently?

No, it's okay.  :+1: 
I don't want to create any additional work.

I think IDE support improves a lot with TypeScript. Will do it myself once we are polishing things up, if it pays off at all.

dnlggr:
> I think IDE support improves a lot with TypeScript.

Good point! 👍

editwentyone:
new proposal that I came up with right now: 

image

instead of "create wallet" directly on settings page, we could remove that, and rename "switch wallet" to "manage wallets", there the user can directly lock, unlock and create stuff. 

and in the future it makes more sense because there the user could also import and maybe export wallets. what do you think?

dnlggr:
I feel like what we have implemented atm is good enough for now. It's a rather niche feature -- I wouldn't want to spend too much time on it right now. We can definitely keep the idea in the backlog for now, though. Feel free to create an issue for it. :)

editwentyone:
thanks, you are right, I will open a new issue

PR #331, opened 2022-06-15
feat: make jars interactive

dnlggr:
Makes the jars a bit more engaging with some hover effects. Clicking on them will show the complete wallet details. Making this more easy to understand (e.g. first show UTXOs and only on a second layer show complete wallet details) is something we can work on in the future.

## 📸 

https://user-images.githubusercontent.com/10026790/173845884-1d246e9e-f5cd-4fec-9a90-abcf0c7d3359.mov


MaxHillebrand (approved):
Beautiful, ACK

theborakompanioni (approved):

PR #332, opened 2022-06-16
fix: checked state of ToggleSwitch can be controlled by caller

### LINKED ISSUE #309: fix: value of `ToggleSwitch` should be controlled by caller
The `ToggleSwitch` component can not be controlled by the caller.
This would enable resetting the toggle programmatically, which is not possible currently.

Currently, only the `initialValue` can be set. It would be a good improvement to have complete control over it by the calling component.

theborakompanioni:
Fixes #309.

Small change to control the `checked` value by the parent component.
Previously a state change in the calling component was not visually displayed by the `ToggleSwitch` component.
After this change the `ToggleSwitch` will reflect the actual value correctly.

### DIFF
--- a/src/components/ToggleSwitch.tsx
+++ b/src/components/ToggleSwitch.tsx
@@ -5,7 +5,7 @@ interface ToggleSwitchProps {
   label: string
   subtitle?: string
   onToggle: (isToggled: boolean) => void
-  checked: boolean
+  toggledOn: boolean
   disabled?: boolean
 }
 
@@ -13,7 +13,7 @@ export default function ToggleSwitch({
   label,
   subtitle = undefined,
   onToggle,
-  checked,
+  toggledOn,
   disabled = false,
 }: ToggleSwitchProps) {
   const onChange = (e: React.ChangeEvent<HTMLInputElement>) => {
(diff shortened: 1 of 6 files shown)


src/components/ToggleSwitch.tsx
```diff
@@ -5,18 +5,18 @@ interface ToggleSwitchProps {
   label: string
   subtitle?: string
   onToggle: (isToggled: boolean) => void
```
dnlggr, line 7: For consistency with the new prop, let's rename this to `isChecked`.

```suggestion
  onToggle: (isChecked: boolean) => void
```

theborakompanioni: Good point! I am tempted to rename the `checked` property to `toggled`.. would that also be okay?

dnlggr: Let's do `toggledOn` then? Since a switch can be toggled either way, with just `toggled` it would not be clear if it's toggled on or off.

dnlggr (approved):

src/components/ToggleSwitch.tsx
theborakompanioni: Perfect :ok_hand:

PR #334, opened 2022-06-16
fix: disable 'create wallet' link when unlocking wallet

theborakompanioni:
Disable "Create Wallet" button while unlocking a wallet is in progress.
Based on and blocked by https://github.com/joinmarket-webui/joinmarket-webui/pull/351.

### DIFF
--- a/src/components/Wallets.jsx
+++ b/src/components/Wallets.jsx
@@ -203,18 +203,22 @@ export default function Wallets({ startWallet, stopWallet }) {
           </div>
         ) : (
           walletList?.map((wallet, index) => {
+            const noneActive = !serviceInfo?.walletName
+            const isActive = serviceInfo?.walletName === wallet
+            const hasToken = currentWallet && currentWallet.token && currentWallet.name === serviceInfo?.walletName
+
+            const showLockOptions = isActive && hasToken
+            const showUnlockOptions =
+              noneActive || (isActive && !hasToken) || (!hasToken && !makerRunning && !coinjoinInProgress)
             return (
               <Wallet
                 key={wallet}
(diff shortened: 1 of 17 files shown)


src/components/Wallets.jsx
```diff
@@ -54,16 +56,20 @@ export default function Wallets({ currentWallet, startWallet, stopWallet }) {
...
 
+        setIsUnlocking(false)
+
         const { walletname: unlockedWalletName, token } = body
         startWallet(unlockedWalletName, token)
         navigate(routes.wallet)
       } catch (e) {
         const message = e.message.replace('Wallet', walletName)
         setAlert({ variant: 'danger', dismissible: false, message })
+        setIsUnlocking(false)
```
dnlggr, line 72: Maybe put it in a `finally` clause?

theborakompanioni: I think this is not possible due to `navigate` unmounting the component. Will try and report back.

theborakompanioni: Funny how I get 
```
Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a `useEffect` cleanup function.
```
in Chrome, but not in Firefox. It seems I have not fully grasp the component lifecycle as of yet. Would have thought it is the same in both browsers. Is it okay to leave it unchanged?

dnlggr: > I think this is not possible due to navigate unmounting the component. 

Makes sense! ✔️

dnlggr (approved):

PR #345, opened 2022-06-22
feat(jars): destination jar selector on send screen

### LINKED ISSUE #340: Destination jar selector on the send screen
The destination can be an address (manually entered) or a jar.

See designs in [Figma](https://www.figma.com/file/kfejZJFlwBywvLEnPEmJo1/JoinMarket-UI?node-id=2850%3A71153).

Screenshot 2022-06-20 at 17 03 05
Screenshot 2022-06-20 at 17 03 11
Screenshot 2022-06-20 at 17 04 23

dnlggr:
Resolves #340.

## 📸 

https://user-images.githubusercontent.com/10026790/175030752-68d9fd46-945a-4d3a-afa6-78a900a9de9c.mov


package.json
```diff
@@ -21,6 +21,7 @@
     "@types/react-dom": "^17.0.14",
     "bootstrap": "^5.1.3",
     "buffer": "^6.0.3",
+    "classnames": "^2.3.1",
```
dnlggr, line 24: Added this since constructing class name strings is getting messy for some components. It's a widely used and package that's actively maintained. 

Before we usually did something like that:

``` js
className={`${styles['selectableJarContainer']} ${isSelectable ? styles['selectable'] : ''} ${isSelected ? styles['selected'] : ''} `}
```

which is not even fully correct because it will generate trailing whitespaces in the class names string if `isSelectable` and/or `isSelected` is false: `selectableJarContainer  `.

Now we can write the same thing like this:

``` js
className={cx('selectableJarContainer', { selectable: isSelectable, selected: isSelected })}
```

Which is shorter, more readable, and more generates a formally correct class name string for all cases.

tsconfig.json
```diff
@@ -6,6 +6,7 @@
     "forceConsistentCasingInFileNames": true,
     "strict": true,
     "jsx": "react-jsx",
-    "skipLibCheck": true
+    "skipLibCheck": true,
+    "allowJs": true
```
dnlggr, line 10: So that we don't have to do `// @ts-ignore` for every jsx/js import.

editwentyone:
looking very good, just maybe some adjustments:

- the X in the input field could be separated and function to reset/empty the field (not combined with the Jar) (like 2 buttons, X and Jar)
- clicking the Jar again would open modal, selecting another jar is closing the modal & overwriting the value inside the input field, no need to reset before
- the number inside the Jar after choosing is a good idea but needs some formatting, more semi bold and maybe a little bit smaller

editwentyone:
thought about the number on the jar a little bit more. it kind of looks like a pillbox with that number on it.
maybe an arrow indicating that the jar will be filled is enough?

and we could maybe use as an input value the jar #X number?  
image

image

MaxHillebrand (approved):
Sexy!

dnlggr:
> the X in the input field could be separated and function to reset/empty the field (not combined with the Jar) (like 2 buttons, X and Jar)

I think two separate buttons would be a bit fiddly to click honestly. Let's keep it simple and remove the jar symbol and instead indicate which jar we're dealing with in the field itself next to the address. How does that sound?

https://user-images.githubusercontent.com/10026790/175277474-7081a089-5e55-4e80-8b3c-bc82593abbd0.mov

theborakompanioni:
> I think two separate buttons would be a bit fiddly to click honestly. Let's keep it simple and remove the jar symbol and instead indicate which jar we're dealing with in the field itself next to the address. How does that sound?

I think this looks awesome and is an improvement to the previous version, which might have been _a tiny little bit_ confusing. Also, this fixed an issue, where the selected jar was not reset after submitting the form. :+1: 

ACK.

theborakompanioni (approved):

src/components/JarSelectorModal.jsx
```diff
@@ -0,0 +1,84 @@
...
+import styles from './JarSelectorModal.module.css'
+import Sprite from './Sprite'
+import { calculateFillLevel, SelectableJar } from './jars/Jar'
+
+export default function JarSelectorModal({
+  isShown,
+  title,
+  accountBalances,
+  totalBalance,
+  disabledJar = undefined,
```
theborakompanioni, line 13: NIT, as it is not used here and can be changed easily on demand.
Should be an array and named `disabledJars`  – "zero, one, many". 
Disabling zero or one is an edge case derived from the current task.

dnlggr: > Disabling zero or one is an edge case

Isn't disabling zero or one the only actual use case we have right now?

theborakompanioni: > > Disabling zero or one is an edge case
> 
> Isn't disabling zero or one the only actual use case we have right now?

Yes, that's why it is a NIT : D
But it is an edge case because we select the destination jar.. would the component be used for selecting the source jar, it might be that all empty jars should be disabled.. 
It is meant as a general purpose component, right?
That's what I meant when I said this is "derived from the current task" at hand.

dnlggr: Ah ok now I see what you mean. 👍 

Yeah make sense. But let's cross that bridge when we get to it.

PR #349, opened 2022-06-24
fix: Check preconditions before send request

### LINKED ISSUE #155: Check sufficient utxo confirmations exist before submitting send request
I tried sending funds with collaberative transaction enabled. 
The interface indicated that a CoinJoin was in progress...
After some time, the 'Joining...' status indicator disapeared from the dashboard.
The CoinJoin was not successful.
After some time, I realised it was most likely because the UTXO being spent was still unconfirmed.

My suggestion is to add a check, prior to processing the send request, to verify that the UTXO(s) being spent have sufficient confirmations.

theborakompanioni:
Closes #155.

Before this commit, the `Send` page did not validate minimal preconditions that must be met before executing a collaborative transaction. Only the `Jam` page (invoking the scheduler, hence multiple collaborative transactions) displays a warning with further information.

In this commit, the minimal preconditions currently in place (min. 1 spendable utxo [not frozen, not timelocked], min. 5 confirmations of at least one utxo, at least 1 retry left) are externalized to `hooks/CoinjoinPrecondition.tsx`. These prerequisites are currently not very well verified and tested, and can probably be improved a lot – this _is not part_ of this PR.

After this commit these preconditions are also used on the `Send` page when a collaborative transaction is attempted, and if not met, a warning is displayed and the submit button disabled.

Currently `Send` and `Jam` will show different messages, maybe the wording can be harmonized to reuse the warning text across all components in a follow-up PR.

### DIFF
--- a/src/components/Jam.jsx
+++ b/src/components/Jam.jsx
@@ -7,13 +7,15 @@ import { useSettings } from '../context/SettingsContext'
 import { useServiceInfo, useReloadServiceInfo } from '../context/ServiceInfoContext'
 import { useCurrentWallet, useCurrentWalletInfo, useReloadCurrentWalletInfo } from '../context/WalletContext'
 import { isDebugFeatureEnabled } from '../constants/debugFeatures'
-import styles from './Jam.module.css'
+import { useBalanceSummary } from '../hooks/BalanceSummary'
+import { COINJOIN_PRECONDITIONS, useCoinjoinPreconditionSummary } from '../hooks/CoinjoinPrecondition'
 import PageTitle from './PageTitle'
 import ToggleSwitch from './ToggleSwitch'
 import Sprite from './Sprite'
 import Balance from './Balance'
 import ScheduleProgress from './ScheduleProgress'
-import { isLocked, useBalanceSummary } from '../hooks/BalanceSummary'
+
+import styles from './Jam.module.css'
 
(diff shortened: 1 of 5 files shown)


src/components/Jam.jsx
```diff
@@ -114,7 +55,13 @@ export default function Jam() {
   const [collaborativeOperationRunning, setCollaborativeOperationRunning] = useState(false)
   const [schedule, setSchedule] = useState(null)
 
-  const schedulerPreconditionSummary = useSchedulerPreconditionSummary(walletInfo, SCHEDULER_START_ACCOUNT)
+  const startJarUtxos = useMemo(() => {
+    if (!walletInfo) return null
+
+    return walletInfo.data.utxos.utxos.filter((it) => it.mixdepth === SCHEDULER_START_ACCOUNT)
+  }, [walletInfo])
```
dnlggr, line 62: Side note, unrelated to this PR: We have to be careful to harmonize terminology now that we have "Jars". We had "Privacy Levels" in the past and there's still "Accounts" and "Mixdepths". All referring to the same thing.

dnlggr (approved):
tACK looks good. ✅

src/components/Jam.jsx
dergigi: Yes, we should make sure to only use one term. Should be easy enough to clean up, at least on the frontend side of things. Will also be important for documentation.

Issue #337, opened 2022-06-20
Concept: Manage Wallets (simplify wallet mangement further)


editwentyone:
new proposal that I came up with right now:

![](https://user-images.githubusercontent.com/96262473/174605738-509fbf97-f051-47ae-b715-ee546f893ecd.png)

instead of "create wallet" directly on settings page, we could remove that, and rename "switch wallet" to "manage wallets", there the user can directly lock, unlock and create stuff.

and in the future it makes more sense because there the user could also import and maybe export wallets. what do you think?

dnlggr:
After looking deeper into this, I think a good way forward would be to move the wallet management to the ["lock screen"](https://www.figma.com/file/kfejZJFlwBywvLEnPEmJo1/JoinMarket-UI?node-id=3994%3A73345). Turns out that there's some issues--technical and imo also UX wise--with having it in the settings page. The main problem is that to unlock a wallet you need to lock the current one first. If the wallet management screen lives in the settings screen which in turn requires an unlocked wallet to access, we're ending up in no-man's-land as soon as we lock the current wallet from the wallet management screen. We can't be in settings anymore (requires a wallet) so we can't return to where we came from.

This is not impossible to work around but cumbersome and not worth the effort right now imo. A simpler solution would be to have wallet management on the lock screen (kind of like we had before), but make it so that it defaults to a "single wallet" workflow. I'm thinking we could do something similar to [how Google does it](https://www.google.com/search?q=google+account+switcher&source=lnms&tbm=isch&sa=X&ved=2ahUKEwivwqis3834AhUMzqQKHW4iBxgQ_AUoAXoECAEQAw&biw=1920&bih=970&dpr=2#imgrc=DvzBSGlSUK73SM) where you have one default account that's always selected and you just have to enter the password to unlock it. But you also have a dropdown where you can select other accounts (wallets in our case), change the default, create new ones etc.
What do you think--is that something that would make sense here? It would certainly ease a lot off implementation pain and be a pretty straight forward solution that works well with the desired "single wallet" flow.

Could look something like this where the wallet name is a dropdown where you can select other wallets. Also needs a make default and create new option of course.
<img width="617" alt="Screenshot 2022-06-27 at 15 35 03" src="https://user-images.githubusercontent.com/10026790/175954534-27f7ab3f-0ec3-4ae0-bbd2-72a5bc5df2d1.png">

editwentyone:
I like the approach. will think about it.

[...]

PR #351, opened 2022-06-27
refactor: move lock/unlock requests from `Wallet` to `Wallets`

theborakompanioni:
Refactoring for easier testing and in preparation of https://github.com/joinmarket-webui/joinmarket-webui/pull/334, enables reacting to "lock/unlock" attempts in `Wallets` component.  

Instead of passing `startWallet` and `stopWallet` to the `Wallet` component, this will perform the API requests in the parent `Wallets` component and just provide `unlockWallet` and `lockWallet` callbacks to the child.
This in turn greatly simplifies the `Wallet` component: Instead of passing multiple properties to determine whether "lock" and "unlock" buttons should be shown, this decision is made by the caller. 

No functionality has been added or removed.

### DIFF
--- a/src/components/App.jsx
+++ b/src/components/App.jsx
@@ -123,7 +123,10 @@ export default function App() {
           {!sessionConnectionError && (
             <>
               <Route element={<Layout />}>
-                <Route path={routes.home} element={<Wallets startWallet={startWallet} stopWallet={stopWallet} />} />
+                <Route
+                  path={routes.home}
+                  element={<Wallets currentWallet={currentWallet} startWallet={startWallet} stopWallet={stopWallet} />}
+                />
                 {currentWallet && (
                   <>
                     <Route path={routes.jam} element={<Jam />} />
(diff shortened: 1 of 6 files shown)


dnlggr (approved):
Nice refactor. ✅

Just as a heads up, I think much of this is going to change with https://github.com/joinmarket-webui/joinmarket-webui/issues/337 so maybe don't put too much blood, seat, and tears into it. Still n ice to have it simplified like this though!

theborakompanioni:
> Nice refactor. white_check_mark
> 
> Just as a heads up, I think much of this is going to change with #337 so maybe don't put too much blood, seat, and tears into it. Still nice to have it simplified like this though!

Yeah, thanks! That's what I thought when I read what you wrote about the concept. That's why I refrained from doing anything more. Was thinking about a larger refactoring, but then again.. I am trying to behave and have smaller change sets lately. This one does not seems so small, but it is really just moving code around.

Thanks for the approval - will merge!

dnlggr:
> This one does not seems so small, but it is really just moving code around.

No it's totally fine, it's just moving code around -- as you say! 👌

PR #357, opened 2022-06-28
ui: evenly spacing between jars in select modal

theborakompanioni:
Add more spacing to the jar selector modal.

Approve: Yes, looks better.
Decline: Let's keep it how it is.

### DIFF
--- a/src/components/JarSelectorModal.module.css
+++ b/src/components/JarSelectorModal.module.css
@@ -58,7 +58,7 @@
   display: flex;
   flex-wrap: wrap;
   flex-direction: row;
-  justify-content: center;
+  justify-content: space-evenly;
   align-items: center;
   gap: 2rem;
   color: var(--bs-body-color);


editwentyone:
better

dnlggr:
How did you get to this state where some jars' balances are hidden and others' are not? Those balances aren't clickable are they?

theborakompanioni:
> How did you get to this state where some jars' balances are hidden and others' are not? Those balances aren't clickable are they?

They are! Introduced in https://github.com/joinmarket-webui/joinmarket-webui/pull/247

dnlggr:
Oh you're right! I feel like we should give the balance a minimum width then. Should take up the same space than the shortest balance possible (`0.00 000 000`). That way we'd get less UI issue with jars being super narrow.

dnlggr:
Just tried out clicking the balances. It only works when you start out in "balance hidden" mode. 🤔 That's why I couldn't get it to work at first.

theborakompanioni:
> Oh you're right! I feel like we should give the balance a minimum width then. Should take up the same space than the shortest balance possible (`0.00 000 000`). That way we'd get less UI issue with jars being super narrow.

Addressed in https://github.com/joinmarket-webui/joinmarket-webui/pull/360. Let me know what you think.

PR #358, opened 2022-06-28
refactor: pass body of `ConfirmModal` as child node instead of parameter

theborakompanioni:
Harmonizes the color of buttons in a modal's footer.
Currently, a modal is used to select jars or when asking for confirmation to lock a wallet while a service is running.
Before this commit, these modals both had a "cancel" and "confirm" button that looked differently.

After this commit, the buttons will look the same.

In a future PR, maybe the `JarSelectorModal` can make use of the `ConfirmModal` to reuse it's functionality. Refrained from doing so for now to keep the change set small. Also, let's see if it is even feasible to do so.

Additional small refactorings: 
- move `Modal.jsx` to `Modal.tsx`
- use `children` child node to pass the modals body instead of a `body` property

### DIFF
--- a/src/components/JarSelectorModal.jsx
+++ b/src/components/JarSelectorModal.jsx
@@ -73,7 +73,6 @@ export default function JarSelectorModal({
       </rb.Modal.Body>
       <rb.Modal.Footer className={styles.modalFooter}>
         <rb.Button variant="light" onClick={cancel} className="d-flex justify-content-center align-items-center">
-          <Sprite symbol="cancel" width="26" height="26" />
           {t('modal.confirm_button_reject')}
         </rb.Button>
         <rb.Button disabled={selectedJar === null} variant="dark" onClick={confirm}>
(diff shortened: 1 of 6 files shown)


dnlggr (commented):
Not sure if both of those need to have exactly the same buttons. I feel like for the jar selector it makes sense to urge the user to confirm a jar whereas for the wallet-lock warning having both buttons be equally emphasized options seems fair to me:

Screenshot 2022-06-29 at 09 40 07

vs. 

Screenshot 2022-06-29 at 09 40 27

theborakompanioni:
> Not sure if both of those need to have exactly the same buttons. I feel like for the jar selector it makes sense to urge the user to confirm a jar whereas for the wallet-lock warning having both buttons be equally emphasized options seems fair to me.

Ah, that seems legit. I honestly did not think of that. Thank you.
I have reverted the 2 commits targeting the styling. What's left now is just the refactoring:
- move `Modal.jsx` to `Modal.tsx`
- use `children` child node to pass the modals `body` instead of a parameter

dnlggr (approved):

Issue #18, opened 2022-01-24
Citadel integration


dergigi:
Make JoinMarket Web UI available as an app for [Citadel](https://runcitadel.space/).

Probably related to #17, as the underlying architecture should be the same.

theborakompanioni:
Quote from @AaronDewes:
> There is a new version of the app.yml format coming very soon (It's basically finished, it just needs documentation), so I recommend waiting.

Will start as soon as we have confirmation from Aaron.

AaronDewes:
https://github.com/runcitadel/apps/blob/v2/PORTING_APPS.md

Here's a porting guide from docker-compose.yml (Umbrel) to app.yml. Please tell me if you run into any issues.

theborakompanioni:
WIP PR for reference: https://github.com/runcitadel/apps/pull/9

theborakompanioni:
https://github.com/runcitadel/apps/pull/9 (v0.0.3) has not been merged, but implemented separately by @AaronDewes . Closing this now as https://github.com/runcitadel/apps/pull/10 (v0.0.4) was integrated :rocket:

PR #361, opened 2022-07-04
feat: move fidelity bonds to earn screen

### LINKED ISSUE #339: Move fidelity bonds to earn screen and update UI
See designs in [Figma](https://www.figma.com/file/kfejZJFlwBywvLEnPEmJo1/JoinMarket-UI?node-id=2850%3A67638). 

Not all details of the designs will be possible to implement. For example, selecting a fidelity bond before starting the maker is not needed. The server [will automatically select the most valuable one](https://github.com/JoinMarket-Org/joinmarket-clientserver/blob/master/docs/fidelity-bonds.md#how-to-create-and-use-a-fidelity-bond-as-a-yield-generator). Other things, such as the "chances in the market" part, will have to be discussed further and are advanced details that we can implement later. Just use some common sense and pick whatever makes sense, following the general theme of the UI in Figma.

The UI doesn't have a UTXO selector yet so ignore that for now, we can do that separately once we figured out how that should look/work like.

**Some screenshots from the designs:**

Screenshot 2022-06-20 at 16 56 25
Screenshot 2022-06-20 at 16 56 37
Screenshot 2022-06-20 at 16 57 14

dnlggr:
Resolves #339.

Moves fidelity bonds to the earn screen, updates the UI, and adds UTXO selection. Consider this the base version of the 
feature that we can build upon in separate PRs.

👇  A fair chunk of the new lines of code are tests and CSS, don't worry too much about that number. 😄 
Screenshot 2022-07-05 at 15 05 35

## 📸 How does it look?

https://user-images.githubusercontent.com/10026790/177336144-1106ea14-eaea-4b67-9413-1f26c2d89220.mov

## 💡 Possible Improvements

I'll track these points as separate issues if you guys agree on them.

_Things we _MUST_ do before releasing_

- Redo all of the wording and move it to the `translation.json` file. At the moment everything is either dummy text or copied from Figma or elsewhere without much thought.

_Things we _SHOULD_ consider doing before releasing if we have the time_

- Reload the earn page after creating a fidelity bond.
- De-emphasize the UI for creating a second fidelity bond if another one already exists. It's not really useful to do so.
- Use TypeScript in the main `CreateFidelityBond` UI component and redo some of the logic to be more concise and less all-over-the-place.
- Offer to unfreeze the UTXOs that were frozen during the flow.
- Warn the user when selecting non `cj-out` UTXOS (e.g. by making the final button red similarly to how we do it when sending a "normal" transaction on the Send screen.)

_Things we _MIGHT_ consider as improvements for a future releases_

- Offer to make an intermediate CJ transaction before funding the fidelity bond (see [community call #18](https://github.com/joinmarket-webui/joinmarket-webui/wiki/community-call-%2318))
- Paginated (or otherwise more space efficient) UTXO selection.

Let me know what else you would like to see.


theborakompanioni:
Tested it a bit and must say: It rocks :rock: :rock: :rock: :rocket:

Aside from the things you are already aware of or you have even mentioned, what follows are random thoughts I am sure you also had, in no particular order. However, it is always beneficial to take note of them:
- UTXO list can potentially be quite long, let's consider how UX can be improved
  - e.g. Should the list be displayed in a modal? Should it be paginated?
  - I guess every option has it's own benefits and drawbacks, it just has to be agreed upon.
  
- Should utxos be sorted in some way?

- Should we add a warning or a hint to only select "cj-out" utxos, or encourage users
  to run the scheduler before creating a FB?
  
- What are your thoughts on unfreezing previously frozen during the FB setup?
  Can this be done in a non-confusing way without destroying UX?

- Have you thought about externalizing coin control completely in the first version?
  i.e. Coin control must be performed before the FB setup is started
  Would this simplify the form and make it easier to grasp?
  (e.g. can be linked from the FB form but user realize it is a separate functionality)
 
- Creating an additional Fidelity Bond is very prominent, let's not forget to inform the user what this means
  (e.g. only one FB is "active"; it does not make sense to create another FB with lower date/smaller amount)

- By glancing over the code:
  - Noticed async/await inconsistencies. Can be tackled later on.
  - Error state handling: What would really happen if some steps fail?
    - e.g.: `freezeUtxos` or `directSweepToFidelityBond`?
  
Small issues:
- dark mode artifacts
- loading state inconsistencies

Other thoughts:
- The utxos selector items "look like" radio buttons, which implies that only one is selectable.
  I think it is neglectible, but if UTXO selection is kept, are checkboxes more appropriate?

I guess the most important question is whether to keep coin control in the FB setup or complete externalize it.

Great job by the way. Definitely better than anything I could come up with. Really nice :orange_heart:

dnlggr:
Thanks for the feedback @theborakompanioni, I've added a couple more things and updated the description accordingly. I'm marking this as ready for review now. It's still not perfect but I feel like anything else can be done in distinct improvement PRs. Let me know what you think! 🙏

dergigi (approved):
Great work @dnlggr! 👏 

Code looks good to me in general, see NITs inline; only the one TODO is something that should be removed I believe.

I did my best to test this quite thoroughly, and out of the 6-or-so back-to-back test runs I think I managed to break it once. I created a fidelity bond and made some transactions and did some freeze/unfreeze operations, and suddenly the fidelity bond was gone. Not sure what happened, will try to reproduce it again tomorrow. I also managed to reproduce the Gateway timeout error around the same time, so maybe something just broke completely (something that is unrelated to this PR).

I'll give this a tACK now anyway, since we are running with scissors (and I wasn't able to reproduce whatever I did, and it's absolutely possible that nothing is wrong and I'm just stupid in the first place). So: LGTM ✅ 

I agree on the improvements listed, and here are some additional things that could be tracked for follow-up PRs:
- Add a back button for every step (currently there is no way to go back a step)
- Improve confusing iconography ("frozen" uses a lock; on the next step, "UTXOs that will be locked up" shows a lock again; same symbol for two things that are the opposite --> very confusing!)

UX wise the biggest improvements would be a button to batch-unfreeze all frozen UTXOs that were just frozen (or all UTXOs in the jar that were selected). Pre-selecting the largest `cj-out` or something like that would be cool too, that way you can just go "next next next". The batch-unfreeze is almost a must for `0.1.0` I'd say.

In any case, great job, this is already amazing for a first iteration! 💪 🚀 

As mentioned inline, I will take care of the final copy in a separate PR.

src/components/fb/FidelityBondSteps.tsx
```diff
@@ -0,0 +1,294 @@
...
+      ))}
+    </div>
+  )
+}
+
+const ReviewInputs = ({ lockDate, jar, utxos, selectedUtxos, timelockedAddress }: ReviewInputsProps) => {
+  const settings = useSettings()
+
+  const UtxoSummary = ({ title, icon, utxos }: { title: string; icon: string; utxos: Array<Utxo> }) => {
+    const settings = useSettings()
```
theborakompanioni, line 202: Should either `UtxoSummary` be moved out of `ReviewInputs` or the line `const settings = useSettings()` deleted?

src/components/fb/CreateFidelityBond.jsx
```diff
@@ -0,0 +1,395 @@
...
+  }
+
+  useEffect(() => {
+    const utxos = walletInfo.data.utxos.utxos
+
+    const utxosByAccount = utxos.reduce((res, utxo) => {
+      const { mixdepth } = utxo
+      res[mixdepth] = res[mixdepth] || []
+
+      // todo: remove
```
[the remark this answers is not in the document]
dnlggr: Yep definitely needs to go before merging. Removed now. ✅

src/components/fb/FidelityBondSteps.tsx
dnlggr: Removed it. ✅

src/components/fb/FidelityBondSteps.tsx
```diff
@@ -0,0 +1,290 @@
...
+          <div>frozen</div>
+        </div>
+      )}
+      {!isLoading && utxo.locktime && (
+        <div className={cx('utxoLabel', 'utxoFidelityBond')}>
+          <Sprite symbol="lock" width="18" height="18" />
+          <div>locked</div>
+        </div>
+      )}
+      {!isLoading && !utxo.frozen && utxo.label === 'cj-out' && (
```
theborakompanioni, line 145: Just noticed that the classification (`new`, `deposit`, `cj-out`, etc.) are not provided in the data from `/utxo`, but only in `/display`, where they are not on a "utxo", but on address level..

Nonetheless, I think displaying the label here (even when it cannot be set by Jam yet), should be displayed here.

This is an object in `entries` of the `display` response:
```json
{
	"0": {
		"hd_path": "m/84'/1'/1'/1/0",
		"address": "bcrt1qeme6gzjlj5s95jfe8hwauagy8y45teh4rej8jw",
		"status": "cj-out",
		"label": "",
		"extradata": "",
		"amount": "149.99994620",
		"available_balance": "149.99994620"
	}
}
```
Compare with the utxo data structure (no `status` here):
```json
{
	"0": {
		"address": "bcrt1qeme6gzjlj5s95jfe8hwauagy8y45teh4rej8jw",
		"path": "m/84'/1'/1'/1/0",
		"label": "",
		"value": 14999994620,
		"tries": 1,
		"tries_remaining": 2,
		"external": false,
		"mixdepth": 1,
		"confirmations": 7,
		"frozen": false,
		"utxo": "9dbd6db083b3fb23c936bd16dbc52efa64330a3145d4c8c9492136a00c8e4c80:1"
	}
}
```

dnlggr:
> Utxo with expired lock is displayed as "locked", but can be selected. This feels odd and might be a bit confusing - funds are not locked anymore. After freezing (when not selected for the next fb) it still says "locked", now however, "frozen" seems to be more appropriate.

We'd need a way to check if the lockdate is in the past. Do you know what timezone the lockdate returned from JM is? I'm guessing UTC but can't find any docs. 😢

dnlggr:
> async/await "issues": Some functions are marked as async with a single await statement that continues to use promise-style syntax. These functions are in turn called by synchronous functions (without subscribing to the promise). This seems a bit awkward. If this pattern is seen, one can either remove async/await or remove the promise-style syntax. "At least" subscribe to the returned promise.

You are right of course. Removed the async/await keywords. Please have another look if you can. 🙏

src/components/fb/FidelityBondSteps.tsx
dnlggr: If I understand correctly, the API does not return the UTXO labels in the UTXO endpoint but only for the `/display` wallet endpoint on an address label and we need to manually match up UTXOs and addresses to find out which UTXO has which label? 🤔

theborakompanioni:
> We'd need a way to check if the lockdate is in the past. Do you know what timezone the lockdate returned from JM is? I'm guessing UTC but can't find any docs. cry

It's a unix timestamp, so no timezone. `Date.UTC` (albeit in milliseconds) can be used to compare the value: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/UTC

Already present, just not well placed in `BalanceSummary.ts`: https://github.com/joinmarket-webui/joinmarket-webui/blob/master/src/hooks/BalanceSummary.test.tsx#L10-L36 
Move it wherever you see fit!

src/components/fb/FidelityBondSteps.tsx
theborakompanioni: > If I understand correctly, the API does not return the UTXO labels in the UTXO endpoint but only for the `/display` wallet endpoint on an address label and we need to manually match up UTXOs and addresses to find out which UTXO has which label? thinking


Correct.


> Do you know what the difference between a label and a status is?

A `label` is manually applied by the user (https://github.com/JoinMarket-Org/joinmarket-clientserver/blob/3140616d322c85e71591d8d4d46cbb4cea374002/docs/release-notes/release-notes-0.9.4.md#user-chosen-address-labelling). Whereas the `status` a metric applied by the backend (https://github.com/theborakompanioni/joinmarket-clientserver/blob/36bf36bd65f6365cb282a8abf3512e8e11df548a/jmclient/jmclient/wallet_utils.py#L493-L531).

dnlggr:
Thanks for the detailed feedback @dergigi and @theborakompanioni! I've tracked a whole bunch of follow up issues [here](https://github.com/joinmarket-webui/joinmarket-webui/issues?q=is%3Aissue+is%3Aopen+label%3A%22Fidelity+Bonds%22).

I've roughly split them up between v0.0.9 (issues that improve the current flow) and v0.1.0 (issues that are either only nice to haves or extend the current flow) although I do think that we can do an intermediate release and move some of the "not so high prio but still kinda important" issues to an intermediate v0.9.1 before we get going on the big v0.1.0.

theborakompanioni (approved):

PR #374, opened 2022-07-06
regtest: delete wallet lockfiles before startup

theborakompanioni:
Same adaptions as in https://github.com/joinmarket-webui/joinmarket-webui-docker/pull/39

Future changes to the `joinmarket-webui-docker` repo should get rid of much of the duplicated code.
This is a work-in-progress as of yet, so we'd still need to maintain both a bit longer.

Also:
- adapt to use the same command style (e.g. `[ ! ... ]` instead of `! [ ]` for conditions, etc.)
- move docker-entrypoint.sh to root directory (instead of polluting jm `/src` dir)

After these changes are merged, the setup  **must be rebuilt** with `npm run regtest:rebuild`!


dnlggr:
Should I move this out of draft state?

theborakompanioni:
@dnlggr Ready for review!

dnlggr (approved):
Everything seems to work fine after a rebuild. ✅

PR #376, opened 2022-07-07
fix: link to new fidelity bonds form in Cheatsheet component

theborakompanioni:
Fix the link from old home of Fidelity Bonds `/fidelity-bonds` to new location on `/earn`.


dnlggr (approved):
👌 Good catch! Thanks!

PR #378, opened 2022-07-07
fix: Expired fidelity bonds are still shown as locked in UTXO selector

### LINKED ISSUE #373: Expired fidelity bonds are still shown as locked in UTXO selector
Check if the `utxo.lockdate` is in the past and show the "locked" label only when the UTXO is in fact still locked.

theborakompanioni:
Resolves #373.

Small refactoring: Move function `isLocked` to check expiry date of timelocked utxo to proper place from `BalanceSummary` to `fb/utils`.


dnlggr (approved):
LGTM ✅

PR #382, opened 2022-07-07
fix: Correctly fetch status for fidelity bond UTXOs

### LINKED ISSUE #375: Correctly fetch status for fidelity bond UTXOs
For some reason, when loading UTXOs from the backend, a UTXOs `status` (e.g. `cj-out`) is not included. The `status` field seems to only be there for addresses when fetching the entire wallet. So we somehow have to fetch the whole wallet and then match UTXOs to addresses to find out which status a UTXO has.

See https://github.com/joinmarket-webui/joinmarket-webui/pull/361#discussion_r914678203 for more details.

theborakompanioni:
Resolves #375.

Displays the "correct" status in the "utxo selection" step when creating a fidelity bond.

Please note that in https://github.com/joinmarket-webui/joinmarket-webui/pull/383, contents of `hooks/AddressSummary.ts` are moved to `context/WalletContext.tsx` and data placed into type `WalletInfo`. But the approach needs to be agreed upon (and might eventually not be merged). This PR can either be merged independently or closed in favor of #383.

##### Addendum
The mapping is not `utxo -> label`, but rather `address -> label` which leads to the following thinking:
Maybe "reused" addresses can be handled differently.. Should only partly spending from reused addresses be prevented somehow or discouraged? If there are multiple outputs controlled by the same pubkey, they probably should be spent together, meaning, they also shouldn't be frozen independently. Should an utxo selector really be an "address selector", or at least behave like one?

### DIFF
--- a/src/hooks/AddressSummary.ts
+++ b/src/hooks/AddressSummary.ts
@@ -1,0 +1,31 @@
+import { useMemo } from 'react'
+import { WalletInfo, AddressStatus } from '../context/WalletContext'
+import { BitcoinAddress } from '../libs/JmWalletApi'
+
+type AddressInfo = {
+  status: AddressStatus
+  address: BitcoinAddress
+}
+
+type AddressSummary = {
+  [key: BitcoinAddress]: AddressInfo
+}
+
+const useAddressSummary = (currentWalletInfo: WalletInfo | null): AddressSummary | null => {
+  return useMemo(() => {
+    if (!currentWalletInfo) {
+      return null
+    }
+
+    const accounts = currentWalletInfo.data.display.walletinfo.accounts
+    return accounts
+      .flatMap((it) => it.branches)
+      .flatMap((it) => it.entries)
+      .reduce((acc, { address, status }) => {
+        acc[address] = { address, status }
+        return acc
+      }, {} as AddressSummary)
+  }, [currentWalletInfo])
+}
+
+export { useAddressSummary, AddressSummary, AddressInfo }
--- a/src/components/fb/CreateFidelityBond.jsx
+++ b/src/components/fb/CreateFidelityBond.jsx
@@ -155,6 +155,7 @@ const CreateFidelityBond = ({ otherFidelityBondExists, accountBalances, totalBal
       case steps.selectUtxos:
         return (
           <SelectUtxos
+            walletInfo={walletInfo}
             jar={selectedJar}
             utxos={utxos[selectedJar]}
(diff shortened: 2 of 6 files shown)


src/hooks/AddressSummary.ts
```diff
@@ -0,0 +1,31 @@
...
+type AddressInfo = {
+  status: AddressStatus
+  address: BitcoinAddress
+}
+
+type AddressSummary = {
+  [key: BitcoinAddress]: AddressInfo
+}
+
+const useAddressSummary = (currentWalletInfo: WalletInfo | null): AddressSummary | null => {
```
dnlggr, line 14:
```suggestion
const useAddressSummary = (currentWalletInfo: WalletInfo | null): AddressSummary | null => {
```

What's the reasoning behind allowing `currentWalletInfo` to be `null`? Everywhere where we use `useWalletSummary` we do a non-null assertion (`!`) anyway. Would it work to just do this signature and remove these assertions?

``` suggestion
const useAddressSummary = (currentWalletInfo: WalletInfo): AddressSummary => {
```

dnlggr (commented):
The fact that addresses are labelled `cj-out` is weird to me. As you say this is strange in case of (what will hopefully never occur in a normal JM flow) reused addresses. But UTXOs on reused addresses are frozen automatically anyway so that should already make it so that users won't spend them accidentally.

src/hooks/AddressSummary.ts
theborakompanioni: Yes, you are right. This would be possible. I was thinking, as this is currently implemented as a hook, following scenario can occur:
```
const currentWallet = useCurrentWallet()
const addressSummary = useAddressSummary(walletInfo)
```
Since `useCurrentWallet` can return `null`, `useAddressSummary` should also accept it as parameter..

But you are completely right, this should never happen in this case, as it is even prevented to load some pages (Send, Earn, etc.) if no wallet is present.

The upcoming changes in https://github.com/joinmarket-webui/joinmarket-webui/pull/383 already do away the `null` param. Would it be okay for you, if you are satisfied with the changes in general, to close this PR in favor of #383?

dnlggr: Would you prefer to have #383 replace this PR?

theborakompanioni: Yes.

theborakompanioni:
Closing this in favor of https://github.com/joinmarket-webui/joinmarket-webui/pull/383.

PR #384, opened 2022-07-07
fix: do not show FB create form when maker is running

theborakompanioni:
Creating a fidelity bond funding transaction includes freezing utxos, which should not be done when a taker or maker is running.
Before this commit, the FB form is hidden when the taker service is running, but not when the maker is running. 
After this commit the form is hidden when any service is running, while still displaying existing FBs when the maker is active).

### DIFF
--- a/src/components/Earn.jsx
+++ b/src/components/Earn.jsx
@@ -492,13 +492,13 @@ export default function Earn() {
                         </div>
                       )}
 
-                      <hr />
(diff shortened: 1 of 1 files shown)


dnlggr (approved):
LGTM ✅ 

Added a small spacing adjustment so that the button's top spacing remains the same no matter whether the maker is running or not.

PR #388, opened 2022-07-11
Batch unfreeze

### LINKED ISSUE #366: Batch unfreeze UTXOs after creating a fidelity bond
After a fidelity bond has been created, a user will likely want to unfreeze any UTXOs that were frozen while creating the fidelity bond. We should offer a button at the end of the FB flow that allows them to do so.

dnlggr:
Resolves #366.

Give the user the option to batch-unfreeze those UTXOs that were frozen during setting up the fidelity bond.

Only those UTXOs that were batch-frozen during the creation of a fidelity bond will be unfrozen. Those that were frozen before starting the fidelity bond flow will remain frozen.

Not super pretty but good enough for now I would say.

## 📸 

https://user-images.githubusercontent.com/10026790/178225730-db626d9c-5484-4c87-8199-39fa00efc6a0.mov


dergigi:
That looks fantastic! Will do my best to test & review soon.

theborakompanioni:
Looks good. :muscle: 

Only concern is the timeout after which wallet data is reloaded (`TIMEOUT_RELOAD_UTXOS_AFTER_FB_CREATE_MS`). On mainnet, this might be very different from a local regtest setup. Would it be difficult to wait for utxos used for the FB to be removed from the wallet (just like on the Send page)? You'd get the inputs used from the `direct-send` response (or alternatively, the used utxos are already known before sending). 

UI suggestions:
- probably remove the `margin-bottom` of `.utxoSummaryIcon` class.

dnlggr:
> Would it be difficult to wait for utxos used for the FB to be removed from the wallet (just like on the Send page)?

Good point. Posted a quick fix. Hope I got it right. Could you be so kind and have another look if it all makes sense?

theborakompanioni (approved):

MaxHillebrand (approved):
Well done!

PR #389, opened 2022-07-11
fix: encode wallet name param in url path

theborakompanioni:
Resolves https://github.com/joinmarket-webui/joinmarket-webui/issues/387.

Quickfix for the issue by encoding the wallet name in the url path.

An alternative solution would be prevent certain characters as proposed in https://github.com/joinmarket-webui/joinmarket-webui/issues/352. Nonetheless, if other chars than [a-zA-Z0-9] are allowed, it would have to be escaped anyway.


dnlggr (approved):
LGTM! ✅

PR #383, opened 2022-07-07
refactor: add `AddressSummary` to `WalletInfo`

theborakompanioni:
First attempt to enhance`WalletInfo` with data that is used through-out the app. Based on https://github.com/joinmarket-webui/joinmarket-webui/pull/382.

Data that needs to be loaded together (otherwise one or the other is outdated) is reflected by `CombinedRawWalletData` (naming suggestions welcome). Idea is that `WalletInfo` should be able to hold every information derived with function `toWalletInfo` from this input.

Next candidate is `BalanceSummary`, which can be done once this changes are reviewed, agreed upon and merged.

Feedback is highly appreciated.

Issue #362, opened 2022-07-06
Fidelity bonds wording


dnlggr:
- [ ] Wording; Relevant files are `Earn.jsx`, `CreateFidelityBond.jsx`, `FidelityBondSteps.tsx`, `ExistingFidelityBond.jsx`
- [x] Translations; Move all strings to the `translation.json` file.

Assigning to you @dergigi for the wording part. Let me know if you need help with the translation technicalities. It can be quite annoying.

PR #392, opened 2022-07-11
feat: visual warning when selecting non cj-out UTXOs for fidelity bond

### LINKED ISSUE #369: Visually warn the user when selecting non cj-out UTXOs for fidelity bonds
Similarly to how we show a red button when sending a transaction without collaborators we could indicate "danger" when creating a FB from a non cj-out UTXO.
In the future we might want to offer doing a collaborative transaction to an intermediate address before funding the FB from there. See #367.

dnlggr:
Resolves #369. Depends on #383 (merge that one first).

Visually warns the user if non cj-out UTXOs are selected for a fidelity bond. Similar to how it's done on the _Send_ screen.

Wording still todo (see #362).

## 📸 

Screenshot 2022-07-11 at 15 42 03

---

Screenshot 2022-07-11 at 15 42 31


theborakompanioni:
Good change, imho!

As you said, wording can probably still be a bit improved, but this will be done in #362.

Only thing I can think of is, that this message should probably not be shown, if a Fidelity Bond is renewed, ie. if the only selected utxo is from an expired timelocked address. What do you think?

dnlggr:
> Only thing I can think of is, that this message should probably not be shown, if a Fidelity Bond is renewed, ie. if the only selected utxo is from an expired timelocked address. What do you think?

Good point. Fixed. ✅

PR #401, opened 2022-07-14
style: improve fidelity bond wording

dergigi:
Definitely not perfect yet, but it might be better than before 😅 

### Things to note:

The capitalization of jar vs. Jar depends on whether a specific jar is referred to or not:

```
One jar, many jars. 
Put it in Jar #0
```

I used all-caps case for UTXO:

```
One UTXO will be frozen, two UTXOs will remain untouched.
```

I used `time-lock` instead of `timelock`; not sure about this one, but it is nicer on the eyes in most parts of the interface imho.

I changed some other small things related to wording here and there, but mostly tracked new issues for the small NITs I found.

---

Also contains a small fix and a tiny improvement, see 148060cbc6dc5252c28574c9df4362c58291cf66 and 71a8aa4f76d4494d5375aed72086969173033be7


src/i18n/locales/en/translation.json
```diff
@@ -289,44 +289,45 @@
...
-        "description_selected_utxos": "Selected UTXOs",
-        "description_unselected_utxos": "The following UTXOs will not be used for the fidelity bond.",
-        "description_selected_utxos_to_freeze": "They will be frozen to remain in jar #{{ jar }}. You can unfreeze them anytime after creating the fidelity bond."
+        "description_selected_utxos": "Selected UTXOs:",
+        "description_unselected_utxos": "The following UTXOs will not be used for the fidelity bond:",
+        "description_selected_utxos_to_freeze": "They will be frozen to remain in Jar #{{ jar }}. You can unfreeze them anytime after creating the bond."
       },
       "review_inputs": {
         "text_primary_button": "Create fidelity bond",
         "text_primary_button_unsafe": "Create fidelity bond with potentially less private UTXOs",
```
dergigi, line 301: I'd love to come up with something clever that replaces "potentially less private" without saying "non-cjout UTXOs" 🤣 

---

Just spitballing here, but we could rename the `cjout` label to `jam` or `jammed` (hmm... "jammed" has a negative connotation, so probably don't use that) so we could call "bad" UTXOs "non-jam" or something like that.

Sticking with the jam (and making jam) metaphor, we could call "bad" UTXOs "lumpy" or "clumpy", and "good" ones (mixed, `cj-out`) "smooth" or "silky" or "kosher" or something. "Smooth" is probably best, since it is used in cooking, as in "blend until smooth."

dergigi: Again: this is just a thinking-out-loud kind of thing, not saying we should absolutely do this. I'd also be happy with less playful language, i.e. revolving around privacy, or even the "default" lingo of mixed and unmixed (which works well with the jam metaphor too).

dnlggr: I like the idea of inverting it. Highlight the non-jammed UTXOs as bad instead of the other way around. Naming is tricky. _Jammed_ would be the ideal word but you're right -- it's too negative. 🤔 I wouldn't go too far with the lingo. With something like "smooth", "silky" I feel like it might be unclear what we mean unless we also use it in lots of other places in the app...

dergigi: We could also use the "pre-mix" and "post-mix" differentiation, and thus we could say something like:

> ⚠️  Create fidelity bond with pre-mix UTXOs

dnlggr: I'm a bit hesitant with using "mixing" as well. IIRC we once said to avoid it since it might be confused with the old school coin mixers from back in the days. I don't have a better suggestion though so until we come up with something better it's definitely better than "potentially less private" imo.

dnlggr (approved):
Looks good to me! I'll leave it up to replace the "potentially less private" with something else before merging or merge now and replace it later @dergigi.

LGTM ✅

dergigi:
Awesome, thx. Let's leave it as is for now. I'd like to sleep on it, and there are many NITs and smaller issues still anyways. LFG v0.0.9! 🚀

PR #402, opened 2022-07-14
feat: time-lock-icon

dnlggr:
Replaces the normal lock icon that labels fidelity bonds with a "time-lock":

Screenshot 2022-07-14 at 11 20 21


dergigi (approved):
Beautiful! 😍

PR #411, opened 2022-07-20
feat: show address reuse warning

### LINKED ISSUE #404: Show address reuse warning
Currently, it is not only possible to reuse addresses in Jam, there is also no indication that there's anything wrong with that. We probably want to, at the very least, show a warning. We should be able to easily do this for all "internal" Jam addresses, i.e. if you accidentally paste the same address twice (e.g. to fund a jar with multiple transactions).

It would also be good to show a warning in general, for any address, but we can't query the backend for that as of now.

We could also think of a "strict" setting that doesn't allow any reuse at all, no matter what.

dnlggr:
Resolves #404.

Warns the user in case a destination address on the Send screen is a known "internal" address that's not labelled as `new`.

Edit: Currently, we'd not allow sending to that address. Alternatively, we could also just show the warning but allow sending Any preferences @dergigi @theborakompanioni?

## 📸 

Screenshot 2022-07-20 at 11 17 05

### DIFF
--- a/src/components/Send.jsx
+++ b/src/components/Send.jsx
@@ -266,6 +266,7 @@ export default function Send() {
   const [isSweep, setIsSweep] = useState(false)
   const [destinationJarPickerShown, setDestinationJarPickerShown] = useState(false)
   const [destinationJar, setDestinationJar] = useState(null)
+  const [destinationIsReusedAddress, setDesitnationIsReusedAddress] = useState(false)
 
   const [waitForUtxosToBeSpent, setWaitForUtxosToBeSpent] = useState([])
   const [paymentSuccessfulInfoAlert, setPaymentSuccessfulInfoAlert] = useState(null)
@@ -424,6 +425,17 @@ export default function Send() {
 
     return () => abortCtrl.abort()
   }, [isOperationDisabled, wallet, reloadCurrentWalletInfo, reloadServiceInfo, loadConfigValue, t])
+
+  useEffect(() => {
+    if (walletInfo?.addressSummary[destination]) {
+      if (walletInfo?.addressSummary[destination].status !== 'new') {
+        setDesitnationIsReusedAddress(true)
+        return
+      }
+    }
+
+    setDesitnationIsReusedAddress(false)
+  }, [walletInfo, destination])
 
   const sendPayment = async (account, destination, amount_sats) => {
     setAlert(null)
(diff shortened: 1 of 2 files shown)


src/components/Send.jsx
```diff
@@ -266,6 +266,7 @@ export default function Send() {
   const [isSweep, setIsSweep] = useState(false)
   const [destinationJarPickerShown, setDestinationJarPickerShown] = useState(false)
   const [destinationJar, setDestinationJar] = useState(null)
+  const [destinationIsReusedAddress, setDesitnationIsReusedAddress] = useState(false)
```
theborakompanioni, line 269: Small typo: `setDesitnationIsReusedAddress` -> `setDestinationIsReusedAddress`

src/components/Send.jsx
```diff
@@ -425,6 +426,17 @@ export default function Send() {
     return () => abortCtrl.abort()
   }, [isOperationDisabled, wallet, reloadCurrentWalletInfo, reloadServiceInfo, loadConfigValue, t])
 
+  useEffect(() => {
+    if (walletInfo?.addressSummary[destination]) {
```
theborakompanioni, line 430:
```suggestion
    if (destination && walletInfo?.addressSummary[destination]) {
```
?

Apparently..
```
var dict = { null: 1 }
dict[null]
// yields 1
```

dnlggr: Yeah but will `addressSummary` ever have a `null` entry we'd accidentally access that way?

theborakompanioni: I guess not, just trying to come up with stuff that breaks things. Maybe in the future "null" is a valid Bitcoin invoice? : D
Chances are for that are basically zero, yes. I'd say it is just good practice (apparently, `{"null": 1}[null]` yields also `1`).
I am with you, though, it's a nit.

dnlggr: Added the null check, you're right it's good practice.

> Maybe in the future "null" is a valid Bitcoin invoice? : D

If that'd be the case, then we'd want to check if the `null` address is reused though right? :D

theborakompanioni: > If that'd be the case, then we'd want to check if the `null` address is reused though right? :D
`null` is falsy, `"null"` is truthy, oh my..  : D

theborakompanioni (approved):
Approved. :muscle: 

I guess we can do the same on the Jam page in another PR.

dergigi:
> Currently, we'd still allow sending to that address. Alternatively, we could also disable sending in this case. Any preferences @dergigi @theborakompanioni?

+1 for disabling.

PR #421, opened 2022-07-21
fix: spacing in jar overlay header and `onKeyDown`

theborakompanioni:
Fixes a small artifact in the jar overlay header and the `onKeyDown` handling (did not fire when the outer most element lost focus).

##### Before
image

##### After
image


dnlggr (approved):
LGTM ✅

PR #427, opened 2022-07-26
feat(cheatsheet): link to jamdocs.org

dergigi:
Since [jamdocs.org](https://jamdocs.org/) is now live, let's link to the appropriate site there instead.

--> https://jamdocs.org/interface/00-cheatsheet/


dnlggr (approved):

dnlggr:
👌 

Thinking of it, we should probably also add a link to the docs to the settings page in Jam. Tracked https://github.com/joinmarket-webui/joinmarket-webui/issues/428.

PR #430, opened 2022-07-27
feat: utxo list

### LINKED ISSUE #338: Jar drill down view
Currently when you click on a jar on the main wallet view, you'll get a big view with a bunch of information which is basically equivalent to the output of the `wallet-tool.py` CLI script.

Screenshot 2022-06-20 at 16 19 11

I think we could be a bit more innovative here and do something more intuitive. In the past we've talked about showing the individual UTXOs in each jar instead of too much on addresses of a wallet. Obviously we still need to show the addresses somehow. I think this is an opportunity to be quite creative here. CC @dergigi and @editwentyone for some creative and conceptional input.

Things the drill down view should support related to UTXOs:
- Freezing/unfreezing UTXOs
- Showing UTXO labels

Things the drill down view should support related to addresses:
- Showing fresh, unused addresses
- Getting a derivation path's xpub
- Showing address labels

Let me know what is missing here. ☝️ 

---

<table>
<thead>
<th>Current Implementation</th>
<th>CLI Output</th>
</thead>
<tr>
<td width="50%">
Screenshot 2022-06-20 at 16 28 33
</td>
<td width="50%">
Screenshot 2022-06-20 at 16 28 33
</td>
</tr>
</table>

dnlggr:
Resolves #338.

Still in draft as some cleanup, i18n, etc. is still missing. Should be ready to play around with, though. Let me know what you think. More advanced stuff like filtering the table by tags or so can be done later imo.

⚠️ The diff is only so huge because I installed [a library](https://github.com/table-library/react-table-library) to handle the table.

## 📸 

Screenshot 2022-07-27 at 17 27 52


dnlggr:
Btw, since it came up in the ccommunity call: The tags are colored as follows:

- `cj-out`: 🟢 Green to indicate success. We want people to have lots of green UTXOs.
- `change-out`: 🟡 Yellow to indicate "be careful". Getting change out of a collaborative transaction is something completely normal (unless you sweep). So it's fine to have these UTXOs, but you have to be careful with them.
- `reused`: 🔴 Red to indicate danger. You're likely doing something wrong if you have a reused address.

All other tags won't have any color. Thoughts on that? cc @dergigi

package.json
```diff
@@ -41,7 +41,11 @@
     "react-router-dom": "^6.3.0",
     "react-scripts": "^5.0.1",
     "typescript": "^4.7.2",
-    "web-vitals": "^2.1.4"
+    "web-vitals": "^2.1.4",
+    "@emotion/react": "^11.9.3",
+    "@table-library/react-table-library": "^4.0.10",
+    "install": "^0.13.0",
+    "npm": "^8.15.0"
```
theborakompanioni, line 48: This might have been a mistake. Is it really necessary?

dnlggr:
Ye, I’m just cleaning up these things. Code is not yet ready. ;)

package.json
dnlggr: Yep I'm just cleaning up these things. Code is not quite ready for review. 😉

src/components/jar_details/JarDetailsOverlay.tsx
```diff
@@ -0,0 +1,374 @@
...
+                        utxos={props.utxosByAccount[accountIndex] || []}
+                        walletInfo={props.walletInfo}
+                        setSelectedUtxoIds={setSelectedUtxoIds}
+                        setDetailUtxo={setDetailUtxo}
+                      />
+                    </div>
+                  )}
+                </div>
+              ) : (
+                <div className={styles.utxoListContainer}>
```
theborakompanioni, line 351: `className` also `utxoListContainer` here? Maybe just a semantic issue.

theborakompanioni:
Dark mode issue in the "Account Details" view. This might have been introduced by my while updating bootstrap.



As I am the culprit, I will fix it myself later on. Just noting it down.

package.json
theborakompanioni: Ah, okay sorry! Take your time. : D

dnlggr:
@theborakompanioni Added the confirmations to the table as discussed. ✅ Also added a refresh button that reloads the wallet info and thus refreshes the utxos.


Screenshot 2022-07-28 at 12 18 26

src/components/jar_details/JarDetailsOverlay.tsx
```diff
@@ -0,0 +1,429 @@
...
+            )}
+            <div>
+              <strong>{t('jar_details.utxo_list.utxo_detail_label_status')}</strong>: {status}
+            </div>
+          </div>
+        </div>
+      </rb.Modal.Body>
+      <rb.Modal.Footer className="d-flex justify-content-center">
+        <rb.Button variant="light" onClick={close} className="w-25 d-flex justify-content-center align-items-center">
+          Close
```
theborakompanioni, line 176:
```suggestion
          <span>{t('global.close')}</span>
```

src/components/jar_details/UtxoList.tsx
```diff
@@ -0,0 +1,270 @@
...
+    var rawStatus = walletInfo.addressSummary[utxo.address]?.status
+
+    var status: string | null = null
+    var locktime: string | null = null
+
+    if (rawStatus) {
+      const indexOfLockedTag = rawStatus.indexOf('[LOCKED]')
+
+      if (indexOfLockedTag !== -1) {
+        locktime = rawStatus.substring(0, indexOfLockedTag).trim()
```
theborakompanioni, line 86: `utxo.locktime` is also populated for time-locked utxos. Is there a specific reason why locktime is parsed from the `status` value? Just asking, because It has not been _that_ reliable in the past.

theborakompanioni:
What follows are some observations while playing with the UI. First of all: Great job. Amazing for a first iteration. It is very informative, functional, visually appealing, and hopefully quite useful for users. I'm eager to hear the feedback from the community.

### Observations
- Currently it is possible to unfreeze non-expired fidelity bonds.
  - It is prevented in current `CurrentWalletAdvanced` view. Cannot remember exactly right now what the issue is, but I think this leads to some problems. At best, it is handled in the backend, but till that is the case, we should keep this behaviour imho (was introduced in c68e4c5c64).
  
- If the overlay is opened, switched to tab "Jar Details" and subsequently closed. The next time the overlay is opened tab "UTXOs" is activated, but the contents of tab "Jar Details" are still shown. This also applies to the active Jar -> Open Jar#0 -> Switch to Jar#2 (active) -> Close overlay -> Click Jar#0 -> Jar#2 still shown :x: .

What follows are things that can probably be addressed in follow-up PRs:

- Can sorting the utxo list by confirmations be added? I think there is a use case for users seeking to "spend my most recent coins" or "spend oldest coins".

### Nits
- Hovering over addresses in "Jar Details" displays a border - is this intended? If so, a left padding is missing. I think it actually looks good.. 

- "Wallet Info reload" function is great. But it feels like a visual feedback is missing (On reload success -> nothing changes -> user might not know what happened or are "unsatisfied"/). 
  Thinking of two things:
  - showing a :heavy_check_mark:  icon like we do on successful "address copy"
  - introducing toasts (https://react-bootstrap.github.io/components/toasts/) - e.g. "Successfully reloaded wallet informatioN"
    - Thought: They might be needed in the future anyway, e.g. displaying infos from websockets (incoming transaction, etc.)
  
- Now that the switching of Jars in the overlay looks nice, do you think it should switch places with the "close" button?
  - Thought: This is the only view having the close button on the left hand side. Would be more consistent to have it on the right, imho.
  
- If the first account in a jar is empty, the accordion body is visually empty (other show an xpub, but JM treats it specially and does not provide the xpub for the first account)
  - As a user, this would "surpise" me, or at least raise questions. What do you think, should some sort of message be shown?
    "e.g. internal account is empty"

- Current max. confirmations is 99999999, then the value is visually cut off. We have time till 3909-09-11 10:10pm till block #99,999,999 is mined.. so maybe this can be fixed later.

### Questions
- Do you think "freeze all non-selected coins" is a common use case?

- Should the same visual representation of tags as in the "UTXOs" tab be used in the "Jar Details" tab?

- Can advanced mode be removed? Maybe the flag `useAdvancedWalletMode` can be kept, but maybe the actual "CurrentWalletAdvanced" view (and "DisplayAccount" respectively) is not needed anymore?
  - Moving all used sub-views to into `jar_details` folder (e.g. `DisplayBranch.tsx`)

dergigi:
Nice! Great job, that looks really awesome already. Love the confirmation indicator. 

> All other tags (new, used, non-cj-change, etc.) won't have any color as they feel somewhat neutral to me. Thoughts on that?

I like the green/yellow/red distinction, and I think having no colors on the other tags is fine. Let's keep it like that and see how it is perceived by users.

---

Only thing I found during testing is that there can be a label/text overlap in the "Jar Details" view, but I guess that's unrelated to this PR. Just bringing it up here so we don't forget.

Screenshot 2022-08-01 at 09 03 48


---

> Jam is a web UI for JoinMarket with focus on user-friendliness. It aims to provide sensible defaults and be easy to use for beginners while still having the features advanced users expect.

We're getting there! 😊 🚀

src/components/jar_details/UtxoList.tsx
dnlggr: That is a good point actually. Let me change that. 👍

dnlggr:
Awesome, thanks for all the detailed feedback! 🙌 🙏 It's great to know that everything's been thoroughly tested. 👌 

I think I addressed everything that needs immediate attention. Please have another look. @theborakompanioni.

Below my comments on the things I didn't address yet.

> "Wallet Info reload" function is great. But it feels like a visual feedback is missing

There's _some_ feedback in terms of a loading spinner being shown and hidden again after loading. I personally don't think much more feedback is required (cf. a reload button in a browser which also only has two states usually: loading and not loading.)

> If the first account in a jar is empty, the accordion body is visually empty 

Jar details needs a redesign and rewrite imo. Haven't touched it beyond some obvious issues.

> Current max. confirmations is 99999999

Let's push that fix to an upcoming version. 😄 

> Have you considered showing the sum of all selected utxos?

Good idea, let's track that as an improvement: https://github.com/joinmarket-webui/jam/issues/440.

> Do you think "freeze all non-selected coins" is a common use case in general?

No I think that'd be confusing to be honest. 😄 Is that a common pattern elsewhere -- performing an operation on non-selected items?

> Should the same visual representation of tags as in the "UTXOs" tab be used in the "Jar Details" tab?

Yes. As mentioned, the Jar Details tab needs a redesign.

> Can advanced mode be removed? Maybe the flag useAdvancedWalletMode can be kept, but maybe the actual "CurrentWalletAdvanced" view (and "DisplayAccount" respectively) is not needed anymore?

Yes I think we're ready to finally remove it now. I tracked https://github.com/joinmarket-webui/jam/issues/441.

src/components/jar_details/UtxoList.tsx
```diff
@@ -0,0 +1,274 @@
...
+const TABLE_THEME = {
+  Table: `
+    --data-table-library_grid-template-columns: 2rem 3.5rem 2fr 4fr 6rem 3fr 1fr;
+    font-size: 0.9rem;
+  `,
+  BaseCell: `
+    &:nth-of-type(1) {
+      text-align: center;
+    }
+  `,
```
theborakompanioni, line 42:
```suggestion
  BaseCell: `
    &:nth-of-type(1) {
      text-align: center;
    }
    &:nth-of-type(3) button {
      display: flex;
      justify-content: end;
    }
    &:nth-of-type(5) button {
      display: flex;
      justify-content: center;
    }
  `,
```

This change aligns the column header accordingly. Just a thought. Yes/no?

dnlggr: Oh yeah good point. We can remove the code below as well then. `BaseCell` will also apply to `Cell`.

src/components/jar_details/JarDetailsOverlay.module.css
```diff
@@ -0,0 +1,236 @@
...
+:root[data-theme='dark'] .overlayBody .accountItem {
+  background-color: var(--bs-gray-800) !important;
+  box-shadow: 0px 1px 2px rgba(0, 0, 0, 0.5);
+}
+
+.overlayBody .accountItem > h2 > button {
+  border-radius: 0.5rem !important;
+  background-color: var(--bs-gray-100);
+  box-shadow: none;
+}
```
theborakompanioni, line 232:
```suggestion
.overlayBody .accountItem > h2 > button {
  border-radius: 0.5rem !important;
  background-color: var(--bs-gray-100);
  box-shadow: none;
}

:root[data-theme='dark'] .overlayBody .accountItem > h2 > button {
  background-color: transparent;
}
```

Fixes background color of accordion header in dark-mode.

theborakompanioni: Maybe we should strive to have more general style rules and move these to `index.css`.

e.g. something like this in `index.css`:
```css
.accordion {
  --bs-accordion-bg: var(--bs-gray-100);
}

.accordion-button {
  background-color: transparent;
}
```

editwentyone:
https://www.figma.com/file/kfejZJFlwBywvLEnPEmJo1/JoinMarket-UI?node-id=5071%3A87607

image

https://user-images.githubusercontent.com/96262473/184351051-60867ad0-af57-4861-a041-68231bff357f.mov

editwentyone:
> There's _some_ feedback in terms of a loading spinner being shown and hidden again after loading. I personally don't think much more feedback is required (cf. a reload button in a browser which also only has two states usually: loading and not loading.)
> 
> >

I think its better **to have a green Checkmark replacing the spinning loader / refresh icon for 2 seconds** as an indicator. 
your example with a reload for a browser window has a visual feedback on success: the whole window goes blank and rebuilds when you do a browser refresh. because we don't do that, a little success indicator (also possible as a toast) in our case would be helpful

PR #432, opened 2022-07-29
fix: text and border colors after bootstrap update

theborakompanioni:
Fixes small oversights after updating bootstrap in dark theme.

Concerning components:
- font and border colors in accordion body in `CurrentWalletAdvanced`
- font color of cancel button in Fidelity Bond form (Earn page)

Before/After


dnlggr (approved):
Bootstrap is weird...

PR #433, opened 2022-07-29
feat: prevent address reuse on Jam page

theborakompanioni:
Add a similar check for reused addresses on the Jam page as it already exists on the Send page.

Before this PR, only duplicate address entries in the form inputs have been checked.
After this PR is applied, address validation is extended to all addresses already known to Jam.


dergigi (commented):
Nice! Looks good to me.

I'd err on the side of brevity when it comes to wording, but that's just me. I plan to go over all strings before v0.1 anyway, so let's keep the wording as is for now.

dergigi (approved):

PR #434, opened 2022-07-29
feat: ability to sort and filter orderbook

theborakompanioni:
Adds sorting, filtering and freshing functionality to the Orderbook component.

Keep in mind: This is still hidden behind a feature flag! It can be enabled for all users by merging https://github.com/joinmarket-webui/jam/pull/445

All "fullscreen-offcanvas" component have been generalized, that's why the style of the Earn Report component is affected by these changes.


dnlggr:
Let me know once this is ready to review. :)

theborakompanioni:
> Let me know once this is ready to review. :)

Ready for review! :pray:

src/components/EarnReport.tsx
```diff
@@ -158,27 +159,47 @@ export function EarnReport() {
...
+      placement="bottom"
+    >
+      <rb.Offcanvas.Header>
+        <rb.Container>
+          <div className="w-100 d-flex">
+            <div className="d-flex align-items-center flex-1">
+              <rb.Offcanvas.Title>{t('earn.report.title')}</rb.Offcanvas.Title>
+            </div>
+            <div>
+              <rb.Button variant="link" className="unstyled pe-0 ms-auto me-auto me-md-0" onClick={onHide}>
```
dnlggr, line 192:
```suggestion
              <rb.Button variant="link" className="unstyled pe-0" onClick={onHide}>
```

I don't think you need these margins. The enclosing div will always be just as large as the cancel button. It has no room to grow as the other element in the flexbox is `flex-1` but itself is not. So its child doesn't need those margins.

The reason why they're needed in the Jar Details header is because there the enclosing div of the cancel button _can_ grow (it's `flex-1`). Therefore we need to push the cancel button to the right with `ms-auto`. Also, there the entire header breaks to a `flex-column` on mobile so we need the `me-auto` to center it in this case.

```
 JAR DETAILS LARGE SCREEN
+-------------------------------------------------------------------------+
|                                                                         |
|   +-----------------+ +------------------------+ +------------------+   |
|   |                 | |                        | |                  |   |
|   |                 | |                        | |   ms-auto  +---+ |   |
|   |                 | |                        | | <--------- | X | |   |
|   |                 | |                        | |            +---+ |   |
|   |                 | |                        | |                  |   |
|   +-----------------+ +------------------------+ +------------------+   |
|                                                                         |
+-------------------------------------------------------------------------+


 JAR DETAILS SMALL SCREEN
+---------------------------------------+
|                                       |
|   +-------------------------------+   |
|   |                               |   |
|   |                               |   |
|   +-------------------------------+   |
|                                       |
|   +-------------------------------+   |
|   |                               |   |
|   |                               |   |
|   +-------------------------------+   |
|                                       |
|   +-------------------------------+   |
|   |                               |   |
|   |     ms-auto  +---+  me-auto   |   |
|   |  <---------- | X | ---------> |   |
|   |              +---+            |   |
|   |                               |   |
|   +-------------------------------+   |
|                                       |
+---------------------------------------+



 EARNDREPORT / ORDERBOOK
+-------------------------------------------------------------------------+
|                                                                         |
|   +-----------------------------------------------------+ +---------+   |
|   |                                                     | |         |   |
|   |                                                     | |  +---+  |   |
|   |                                                     | |  | X |  |   |
|   |                                                     | |  +---+  |   |
|   |                                                     | |         |   |
|   +-----------------------------------------------------+ +---------+   |
|                                                                         |
+-------------------------------------------------------------------------+
```

src/components/Orderbook.tsx
```diff
@@ -1,196 +1,414 @@
...
 
 const withTooltip = (node: React.ReactElement, tooltip: string) => {
   return (
     <rb.OverlayTrigger overlay={(props) => <rb.Tooltip {...props}>{tooltip}</rb.Tooltip>}>{node}</rb.OverlayTrigger>
   )
 }
 
-const OrderbookTable = ({ orders }: OrderbookTableProps) => {
+// `TableNode` is known to have same properties as `ObwatchApi.Order`, hence prefer casting over object destructuring
+const toOrder = (tableNode: TableTypes.TableNode): ObwatchApi.Order => tableNode as unknown as ObwatchApi.Order
```
dnlggr, line 85:
```suggestion
const toOrder = (tableNode: TableTypes.TableNode) => tableNode as unknown as ObwatchApi.Order
```

Just wondering: Can the TS type system not infer the return type in this case?

src/components/Orderbook.tsx
```diff
@@ -1,196 +1,414 @@
...
+
+const renderOrderType = (val: string, t: TFunction<'translation', undefined>) => {
+  if (val === ObwatchApi.ABSOLUTE_ORDER_TYPE_VAL) {
+    return withTooltip(<rb.Badge bg="info">{t('orderbook.text_offer_type_absolute')}</rb.Badge>, val)
```
dnlggr, line 89: Tooltips are nice!

src/components/Orderbook.tsx
```diff
@@ -1,196 +1,414 @@
...
+            })}
+            {search && (
+              <>
+                <br />(
+                {t('orderbook.text_orderbook_filter_count', {
+                  filterCount: tableData.nodes.length,
+                })}
+                )
+              </>
+            )}
```
dnlggr, line 296: https://user-images.githubusercontent.com/10026790/182799404-471a9dcf-9ec2-4b98-9acd-c2cc7b9b3424.mov

Adding that text below looks a bit chunky as the header size changes and the texts don't look aligned since one is in parentheses. What about just replacing the `orderbook.text_orderbook_summary` text with the count of orders that's currently displayed? i.e.:

```
SEARCH: (none)

3 orders by 3 counterparties

ORDER: foobar
ORDER: foobaz
ORDER: bla
```

and

```
SEARCH: foo

2 orders by 2 counterparties

ORDER: foobar
ORDER: foobaz
```

src/components/Orderbook.tsx
```diff
@@ -1,196 +1,414 @@
...
-              ))}
-            </tbody>
-          </rb.Table>
         </div>
-      )}
-    </>
+        <div>
+          <rb.Form.Group controlId="search">
+            <rb.Form.Label className="m-0 pe-2 d-none">{t('orderbook.label_search')}</rb.Form.Label>
+            <rb.Form.Control
```
dnlggr, line 302: Any value in disabling search while the orders are refreshed?

src/components/Orderbook.tsx
```diff
@@ -1,196 +1,414 @@
...
+      placement="bottom"
+    >
+      <rb.Offcanvas.Header>
+        <rb.Container>
+          <div className="w-100 d-flex">
+            <div className="d-flex align-items-center flex-1">
+              <rb.Offcanvas.Title>{t('orderbook.title')}</rb.Offcanvas.Title>
+            </div>
+            <div>
+              <rb.Button variant="link" className="unstyled pe-0 ms-auto me-auto me-md-0" onClick={onHide}>
```
dnlggr, line 380:
```suggestion
              <rb.Button variant="link" className="unstyled pe-0" onClick={onHide}>
```

Same as for the earn report above.

dnlggr (commented):
Awesome, looks great! ✨ 

Just a couple of smaller thigns, nothing major. Overall looks awesome -- good job!

src/components/EarnReport.tsx
theborakompanioni: Thanks for the detailed explanation. Highly appreciated. :pray:

src/components/Orderbook.tsx
theborakompanioni: It can! : D
Sometimes I am just overly verbose.

Will apply!

src/components/Orderbook.tsx
theborakompanioni: Good point!

What do you think of adding a hint if the search is activated?
e.g. "2 orders by 2 counterparties (filtered)",  "found 2 orders by 2 counterparties", "2 orders by 2 counterparties match given search criteria"...

Or do you think it is totally unnecessary?

src/components/Orderbook.tsx
theborakompanioni: Done.

src/components/Orderbook.tsx
theborakompanioni: I changed it to have "(filtered)" at the end, as I feel that _some_ sort of visual feedback that the search value has actually been applied is a Good Thing™. However, if you think it is stupid and unncessary, I'll remove it.

dnlggr: I like "found 2 orders by 2 counterparties". It feels the most natural to read imo.

dnlggr: Pushed the wording update myself, hope that's ok. :)

dnlggr (approved):
LGTM ✅ ✨

PR #443, opened 2022-08-03
refactor: remove advanced wallet mode

### LINKED ISSUE #441: Remove advanced mode
Once https://github.com/joinmarket-webui/jam/pull/430 is merged, we're feature complete in a way that we can remove the "Developer/Advanced" mode we had until now.

dnlggr:
Resolves #441.

Removes the "advanced"/"developer" mode we had since all of its functionality is not available in the "normal" mode as well. No more two modes! 🎉


theborakompanioni (approved):
There are still some reference to an "advanced mode", e.g.
- In a comment in `utils.ts` for const `DEFAULT_MAX_TIMELOCK_YEARS`
- a translation key `current_wallet_advanced` when `current_wallet` exists

Should this be addressed now?
Do as you think is best - code looks good. Approved :heavy_check_mark:

dnlggr:
> There are still some reference to an "advanced mode"

Good catch. Removed those too!

PR #445, opened 2022-08-03
feat: remove feature flag and enable orderbook for all users

theborakompanioni:
Removes the feature flag and enables the orderbook for all users.
Can be merged once #434 is integrated.


dergigi (approved):
Looks good to me ✅

dnlggr (approved):

PR #448, opened 2022-08-04
ui: switch position of amount and recipient input on send page

theborakompanioni:
TBD.


dnlggr (approved):
Reasoning being that this is what many other wallets / banking apps use. Moreover, a recipient could theoretically contain an amount (e.g. via [BIP 21](https://en.bitcoin.it/wiki/BIP_0021) URI). Even though we don't support that (yet), it still makes sense to align or UX that way.

PR #451, opened 2022-08-04
feat: add sorting and filtering to earn report

theborakompanioni:
Adds ~~pagination,~~ sorting and filtering to Earnings Report overlay.

Was not quite sure how to best style the pagination. Not that happy with the result. Suggestions welcome!
Also, any ideas for improved wording of the table headings?

## :camera_flash: 
image

### DIFF
--- a/src/components/EarnReport.tsx
+++ b/src/components/EarnReport.tsx
@@ -1,4 +1,8 @@
-import React, { useEffect, useState } from 'react'
+import React, { useCallback, useEffect, useMemo, useState } from 'react'
+import { Table, Header, HeaderRow, HeaderCell, Body, Row, Cell } from '@table-library/react-table-library/table'
+import { useSort, HeaderCellSort, SortToggleType } from '@table-library/react-table-library/sort'
+import * as TableTypes from '@table-library/react-table-library/types/table'
+import { useTheme } from '@table-library/react-table-library/theme'
 import * as rb from 'react-bootstrap'
 import { useTranslation } from 'react-i18next'
 import * as Api from '../libs/JmWalletApi'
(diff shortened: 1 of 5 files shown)


dnlggr:
> Was not quite sure how to best style the pagination. Not that happy with the result. Suggestions welcome!

We can also omit it for now if you're not happy with it. I feel like it's not super essential to have.

MaxHillebrand (dismissed):
How about "my inputs" or "my amounts" as table header?

Maybe for pagination a simple "load more" button at the bottom of the table? Seems like less ui clutter.

src/components/EarnReport.tsx
```diff
@@ -8,172 +13,480 @@ import Balance from './Balance'
...
+  return new Date(Date.parse(`${val} GMT`))
+}
+
+const yieldgenReportLineToEarnReportEntry = (line: string): EarnReportEntry | null => {
+  if (!line.includes(',')) return null
+
+  const values = line.split(',')
+
+  // be defensive here - we cannot handle lines with unexpected values
+  if (values.length < 8) return null
```
dnlggr, line 100: Would `values.length > 8` be equally unexpected?

dnlggr:
Pagination is nice (maybe needs a bit of a cleaner UI to feel less cluttered) but I feel like it'd be more interesting for the orderbook than for the earn report since the orderbook will be hundreds of entries long by default.

src/components/EarnReport.tsx
theborakompanioni: Yes, equally unexpected, but having more values would not impair the component. Acceptable?

edit: It would have impacts if the order of the values change - but then again, that's also the case if a value is replaced, which would result in the same length. I will add some tests to ensure and document the data the component is expecting!

dnlggr: Sounds good!

theborakompanioni:
> Pagination is nice (maybe needs a bit of a cleaner UI to feel less cluttered) but I feel like it'd be more interesting for the orderbook than for the earn report since the orderbook will be hundreds of entries long by default.

Very valid point. Any immediate ideas or recommendations on the styling? Or should it be added and adapted in an upcoming version?

dnlggr:
The easiest would probably be to come up with a UI and add it in one go to both tables? 🤔

src/components/EarnReport.tsx
```diff
@@ -8,172 +13,480 @@ import Balance from './Balance'
...
+      display: flex;
+      justify-content: end;
+    }
+  `,
+  Cell: `
+    &:nth-of-type(2) {
+      text-align: right;
+    }
+    &:nth-of-type(3) {
+      text-align: center;
```
dnlggr, line 57: The input count looks a bit lonely centered like this. Does it look better left aligned? 🤔

theborakompanioni: You are right. I narrowed the heading with the suggestion by @MaxHillebrand.. that improved the situation a little bit.

Do you still think it should be left aligned? It looks a bit strange now imho..

image

as opposed to

image

dnlggr: Let's got with right aligned. Center is weird for a number column in a table and all the others are right aligned as well. See also [here](https://ux.stackexchange.com/a/24073).

theborakompanioni:
I'll be removing the pagination for now - it's not the best in terms of UI/UX. Will be readding it in a new PR to the Earn Report and Orderbook components.

src/components/EarnReport.tsx
```diff
@@ -8,172 +12,375 @@ import Balance from './Balance'
...
+                </HeaderCellSort>
+                <HeaderCell>{t('earn.report.heading_notes')}</HeaderCell>
+              </HeaderRow>
+            </Header>
+            <Body>
+              {tableList.map((item) => {
+                const entry = toEarnReportEntry(item)
+                return (
+                  <Row key={item.id} item={item}>
+                    <Cell>{entry.timestamp.toLocaleString()}</Cell>
```
dnlggr, line 191: I personally prefer ISO strings for timestamps. Might be e personal thing. Aay thoughts?

theborakompanioni:
- Pagination removed :white_check_mark: 
- Tests for parsing data added :white_check_mark: 

ping @dnlggr

src/components/EarnReport.tsx
```diff
@@ -8,172 +12,376 @@ import Balance from './Balance'
...
+const yieldgenReportLineToEarnReportEntry = (line: string): EarnReportEntry | null => {
+  if (!line.includes(',')) return null
+
+  const values = line.split(',')
+
+  // be defensive here - we cannot handle lines with unexpected values
+  if (values.length < 8) return null
+
+  return {
+    timestamp: parseYieldgenTimestamp(values[0]),
```
dnlggr, line 100: I guess here we'd need to use the same format as we display to the user. Otherwise it becomes impossible to search for a date when you don't know what the underlying representation, that's actually being searched for, is.

src/components/EarnReport.tsx
theborakompanioni: Tests added!

dnlggr (approved):

PR #458, opened 2022-08-08
fix: serialize values of `/maker/start` request body as strings

theborakompanioni:
Before this commit, values of `POST /maker/start` have been passed as int/decimal types.
After this commit, they well be passed [according to the api docs](https://joinmarket-org.github.io/joinmarket-clientserver/api/#operation/startmaker) as strings.

This prevents an exception on the backend `builtins.TypeError: a bytes-like object is required, not 'float'`. It seemed to still have worked fine before (at least for absolute offers).

### DIFF
--- a/src/libs/JmWalletApi.ts
+++ b/src/libs/JmWalletApi.ts
@@ -224,7 +224,7 @@ const postMakerStart = async ({ token, signal, walletName }: WalletRequestContex
     headers: { ...Authorization(token) },
     body: JSON.stringify({
       ...req,
-      // we enforce type-safety for the following properties, but their values must actually passed as string!
+      // We enforce type-safety for the following properties, but their values must actually be passed as string!
       cjfee_a: String(req.cjfee_a),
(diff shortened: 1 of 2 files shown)


src/libs/JmWalletApi.ts
```diff
@@ -222,7 +222,14 @@ const postMakerStart = async ({ token, signal, walletName }: WalletRequestContex
   return await fetch(`${basePath()}/v1/wallet/${encodeURIComponent(walletName)}/maker/start`, {
     method: 'POST',
     headers: { ...Authorization(token) },
-    body: JSON.stringify({ ...req, txfee: '0' }),
+    body: JSON.stringify({
+      ...req,
+      // we enforce type-safety for the following properties, but their values must actually passed as string!
```
dnlggr, line 227:
```suggestion
      // We enforce type-safety for the following properties, but their values must actually be passed as string!
```

dnlggr (approved):

PR #460, opened 2022-08-09
regtest: rename default wallet to Satoshi

theborakompanioni:
Renames the regtest default wallet from "funded" to "Satoshi".

Makes for nicer screenshots. :camera_flash:


dergigi (approved):
Wonderful! Thank you 🙏 

Looks good to me ✅

Issue #413, opened 2022-07-20
Earn: Show counterparty ID to the user


dergigi:
Raised by @Richie-BTC-Rich in #14:

> After the [earn] service is running a a link to the marketplace with your Counterparty ID would be helpful so you can verify your offer is listed [by looking up your ID in the orderbook]

---

Not sure if the [RPC API](https://joinmarket-org.github.io/joinmarket-clientserver/api/#operation/startmaker) returns the counterparty ID. Doesn't seem like it at a first glance.

dnlggr:
That's a good point. This ID would be the [nick](https://github.com/JoinMarket-Org/JoinMarket-Docs/blob/master/Joinmarket-messaging-protocol.md#pubkey-based-nicks) used for messaging. Not sure how to get to this nick, even on the CLI. Does anyone know?

theborakompanioni:
> That's a good point. This ID would be the [nick](https://github.com/JoinMarket-Org/JoinMarket-Docs/blob/master/Joinmarket-messaging-protocol.md#pubkey-based-nicks) used for messaging if I'm not mistaken. Not sure how to get to this nick, even on the CLI. Does anyone know?

Opened a ticket in the backend repo: https://github.com/JoinMarket-Org/joinmarket-clientserver/issues/1326

theborakompanioni:
e.g. as far as I can see, joininbox currently parses the nickname from the log files: https://github.com/openoms/joininbox/blob/6c177a71d8fcb87c93d6b3061ffee2393b926b3d/scripts/_functions.sh#L163

In order to do it the same way, a server side script running alongside `jmwalletd` that serve this data via an api would be needed.
Let's wait for a response in https://github.com/JoinMarket-Org/joinmarket-clientserver/issues/1326 and then decide what to do.

Having an own `jamd` running that adds some missing features would definitely be beneficial (however, it also introduces a lot of drawbacks..)..

dnlggr:
I'm moving this out of v0.0.10. I don't think this will become unblocked in the coming days.

PR #461, opened 2022-08-09
feat: show active offers

### LINKED ISSUE #399: Earn: show offer and fee settings when earn is active
There is no way to tell what fees you've set when earn is active. You would have to stop the maker service.

All we show currently is this:

Screenshot 2022-07-14 at 08 36 55

I imagine something like the following would be useful:

> Offering a minimum amount of 100,000 sats at a 0.03% fee.

or, when the offer is absolute

> Offering a minimum amount of 100,000 sats at a 250 sat fee.

---

Basically show whatever was set here:

Screenshot 2022-07-14 at 08 34 11

theborakompanioni:
Resolves #399. Also partly addresses #413 (but will be additionally tackled in a separate PR).

Shows the current active offers on the Earn page.

Please review and merge #456 before this PR.
These changes need JM built from the current `master` (at least including commit [6ec5c35c](https://github.com/JoinMarket-Org/joinmarket-clientserver/commit/6ec5c35c595364137620b5721a878b007b51dac9)).

 

Questions: 
- 'Nickname', 'Nick' or 'Nym'?
- Should the Nym be hidden if `showBalance` is disabled?
- Show some kind of loader till the data is available? Currently it "just appears" whenever the `/session` response arrives..


dnlggr:
This is great! Will test it later--I'm having issues building the regtest right now. 😅 

Regarding the questions:

> 'Nickname', 'Nick' or 'Nym'?

Honestly, I'd go for "Offer ID"--that's what it is in this context, right? I know this is the nickname on the IRC servers as well, but that's something the average user doesn't care about I guess. See also #413 where it's referred to as ID.

> Should the Nym be hidden if showBalance is disabled?

Good question! How about renaming the "show/hide balance" toggle to "show/hide sensitive data" and then also hiding the nick?

> Show some kind of loader till the data is available? Currently it "just appears" whenever the /session response arrives..

Without having tried it yet, what about showing the normal placeholder in the case where we can anticipate that an offer will be present in the `/session` response--that is when "Earn" is active?

src/i18n/locales/en/translation.json
```diff
@@ -247,6 +247,16 @@
     "text_stopping": "Stopping",
     "button_show_report": "Show earnings report",
     "button_show_orderbook": "Show orderbook",
+    "current": {
+      "title": "Active Offer #{{ id }}",
```
dnlggr, line 251: Does this ID mean much? I always get "Active Offer #0". Sounds like I can have multiple offers which I can't (right?) Maybe "Active Offer" is already enough? 🤔

src/i18n/locales/en/translation.json
```diff
@@ -247,6 +247,16 @@
...
+      "text_nickname": "Nym",
```
dnlggr, line 252: I'd go with Offer ID or just ID here, what do you think?

If needed we could also do `<nym>:<oid>` to make it a fully unique ID? 🤔

src/components/Earn.jsx
```diff
@@ -83,6 +84,99 @@ const factorToPercentage = (val, precision = 6) => {
...
+        <div className="d-flex align-items-center gap-1">{renderOrderType(offer.ordertype, t)}</div>
+      </div>
+      <rb.Container className="mt-2">
+        <rb.Row className="mb-2">
+          <rb.Col xs={12}>
+            <div className={styles.offerLabel}>{t('earn.current.text_nickname')}</div>
+            <div>{nickname}</div>
+          </rb.Col>
+        </rb.Row>
+        <rb.Row>
```
dnlggr, line 114:
```suggestion
        <rb.Row className="mb-1">
```

src/components/Earn.jsx
```diff
@@ -83,6 +84,99 @@ const factorToPercentage = (val, precision = 6) => {
...
+  const settings = useSettings()
+
+  return (
+    <div className={styles.offerContainer}>
+      <div className="d-flex justify-content-between align-items-center">
+        <div className={styles.offerTitle}>{t('earn.current.title', { id: offer.oid })}</div>
+        <div className="d-flex align-items-center gap-1">{renderOrderType(offer.ordertype, t)}</div>
+      </div>
+      <rb.Container className="mt-2">
+        <rb.Row className="mb-2">
```
dnlggr, line 108:
```suggestion
        <rb.Row className="mb-1">
```

src/components/Earn.jsx
```diff
@@ -294,6 +388,13 @@ export default function Earn() {
             !serviceInfo?.makerRunning &&
             !isWaitingMakerStart &&
             !isWaitingMakerStop && <p className="text-secondary mb-4">{t('earn.market_explainer')}</p>}
+          {serviceInfo?.makerRunning && serviceInfo?.offers && serviceInfo?.nickname && (
+            <>
```
dnlggr, line 392: I'd show one of our default placeholders here since we can already anticipate that an offer will load (since `makerRunning` is true).

src/components/Earn.module.css
```diff
@@ -46,3 +46,30 @@
...
+
+.offerContainer .offerLabel {
+  color: var(--bs-gray-600);
+  font-size: 0.8rem;
+}
+
+.offerContainer .offerContent {
+  font-size: 0.8rem;
+  word-break: break-all;
+}
```
dnlggr, line 75: That's the same styles as the existing FB box right?

theborakompanioni: Yes.

src/i18n/locales/en/translation.json
theborakompanioni: Opted for just showing `<nym>:<oid>` :+1:

src/i18n/locales/en/translation.json
theborakompanioni: Changed according to the comment below -> `<nym>:<oid>`.

src/components/Earn.jsx
theborakompanioni: Hmm.. the Earn page definitely needs some overhaul regarding loading, etc.
Will add a loader for the "active offer" box, but inclined to do some general refactoring in a follow-up PR.

dergigi:
Found a small edge case, I think.

When the string contains an `x` between two numbers, it will be rendered as a math symbol (which is a bootstrap thing, I believe). 

Screenshot 2022-08-18 at 13-24-22 Jam for JoinMarket

Might not be a big problem, because it's still an `x` afaik, but it looks a bit confusing (at least to me).

Here is how it looks if we wrap the whole string in a `<code>` tag:
Screenshot 2022-08-18 at 13-23-32 Jam for JoinMarket

theborakompanioni:
> Found a small edge case, I think.
> 
> When the string contains an `x` between two numbers, it will be rendered as a math symbol (which is a bootstrap thing, I believe).

Nice one! Adding css class `slashed-zeroes` (which is used throughout the app) will apply `font-feature-settings: 'calt' off` which fixes the problem. Again.. really good catch! Fixed.

dergigi (approved):

Issue #269, opened 2022-05-16
Ability to stop a running taker operation


theborakompanioni:
The only way to stop a running taker operation is to lock the wallet – that is not so intuitive.

At least in regtest, `Makers who didnt respond` can happen quite often. Should the possibility to stop a single taker operation be considered? In the meantime, the workaround works fine.

theborakompanioni:
Related https://github.com/joinmarket-webui/joinmarket-webui/issues/155

editwentyone:
something like this? https://www.figma.com/file/kfejZJFlwBywvLEnPEmJo1/?node-id=3283%3A85776

<img width="698" alt="image" src="https://user-images.githubusercontent.com/96262473/169146333-6c40313b-9739-4609-8f84-4beabaa9a13e.png">

dnlggr:
Looks good. I also like what you did with the tabs in that screenshot. 👌✨

dnlggr:
An issue we currently have is that we cannot differentiate on the client between the a single collaborative transaction and a schedule of collaborative transactions running. This makes it impossible to treat a single collaborative transaction (Send) and the scheduler (Jam) differently -- therefore anytime either of them is "active" UI-wise, the other one needs to be as well. So we can show a "Stop" button on the send page but it will also show when the scheduler is running, which is weird.

I will check with the backend and see if we can get separate "running" flags for the scheduler and a single collaborative transaction. That will make things a lot easier:

- Correctly showing the activity indicator. Currently the "Jam" tab is showing indicating activity when a single collaborative transaction is running
- Showing a stop button on the Send page only when a single collaborative transaction is running
- Blurring the Send page only when the scheduler is running
- ...

theborakompanioni:
>  So we can show a "Stop" button on the send page but it will also show when the scheduler is running, which is weird.

Can it be distinguished by checking if an API call to `/schedule` returns `404` or `200`?
- If `taker_running := true` and `/schedule` responds with `404`: Send is active.
- If `taker_running := true` and `/schedule` responds with `200`: Scheduler is active.

dnlggr:
That's a nice hack. 😉 I think we can get a flag into the `/session` endpoint quite easily though. I'm already on it. If that doesn't play out for some reason we can still fall back to what you described I'd say.

dnlggr:
[...]

PR #463, opened 2022-08-10
feat: click on active "joining" icon opens relevant screen

### LINKED ISSUE #320: Header: click on active "joining" icon should open relevant screen
@editwentyone mentioned this briefly in [Call 16](https://github.com/joinmarket-webui/joinmarket-webui/wiki/Community-Call-%2316), if a collaborative transaction is running clicking this icon (highlighted in green below) should lead the user to "Jam" or "Earn" respectively.

Screenshot 2022-06-06 at 13 51 17

theborakompanioni:
Resolves #320.

Before this commit, the "Joining" text+icon in the header nav was not clickable.
After this commit, the icon in the header nav is clickable and will either link to "Earn", "Jam" or "Send" screen.

Additional changes:
- Added `schedule` property to `ServiceInfo`
- Changed "Joining" icon color from gray to green (like in the ticket screenshot)
- Text "Joining" is only shown on `xs` and `sm` screens (side nav)
- Added Joining indicator to "Send" nav link when single collaborative tx is active (debatable!)

These changes need JM built from the current `master` (at least including commit [6ec5c35c](https://github.com/JoinMarket-Org/joinmarket-clientserver/commit/6ec5c35c595364137620b5721a878b007b51dac9)).

Relevant commits are also part of #461 and have been cherry-picked.

### DIFF
--- a/src/components/Navbar.jsx
+++ b/src/components/Navbar.jsx
@@ -1,10 +1,10 @@
-import React, { useMemo, useState } from 'react'
+import React, { useState } from 'react'
 import { Link, NavLink } from 'react-router-dom'
 import * as rb from 'react-bootstrap'
 import { useTranslation } from 'react-i18next'
 import Sprite from './Sprite'
 import Balance from './Balance'
-import { TabActivityIndicator, JoiningIndicator } from './ActivityIndicators'
+import { TabActivityIndicator } from './ActivityIndicators'
 import { useSettings } from '../context/SettingsContext'
 import { useCurrentWallet, useCurrentWalletInfo } from '../context/WalletContext'
 import { useServiceInfo, useSessionConnectionError } from '../context/ServiceInfoContext'
@@ -99,25 +99,11 @@ const CenterNav = ({ makerRunning, schedulerRunning, singleCollaborativeTransact
   )
 }
 
-const TrailingNav = ({ joiningRoute, onClick }) => {
+const TrailingNav = ({ onClick }) => {
   const { t } = useTranslation()
 
   return (
     <rb.Nav className="justify-content-center align-items-stretch">
-      {joiningRoute && (
-        <rb.Nav.Item className="d-flex align-items-center pe-2">
-          <div className="d-flex align-items-center px-0">
-            <NavLink to={joiningRoute} onClick={onClick} className="nav-link">
-              <rb.Navbar.Text className="d-md-none">{t('navbar.joining_in_progress')}</rb.Navbar.Text>
-              <JoiningIndicator
-                isOn={true}
-                className="navbar-text text-success"
-                title={t('navbar.joining_in_progress')}
-              />
-            </NavLink>
-          </div>
-        </rb.Nav.Item>
-      )}
       <rb.Nav.Item className="d-flex align-items-stretch">
         <NavLink
           to={routes.settings}
@@ -144,19 +130,6 @@ export default function Navbar() {
   const sessionConnectionError = useSessionConnectionError()
 
   const [isExpanded, setIsExpanded] = useState(false)
-
-  const joiningRoute = useMemo(() => {
-    if (!serviceInfo) return null
-
-    if (serviceInfo.coinjoinInProgress) {
-      return serviceInfo.schedule ? routes.jam : routes.send
-    }
-    if (serviceInfo.makerRunning) {
-      return routes.earn
-    }
-
-    return null
-  }, [serviceInfo])
 
   const height = '75px'
 
@@ -252,7 +225,7 @@ export default function Navbar() {
                       }
                       onClick={() => setIsExpanded(!isExpanded)}
                     />
-                    <TrailingNav joiningRoute={joiningRoute} onClick={() => setIsExpanded(!isExpanded)} />
+                    <TrailingNav onClick={() => setIsExpanded(!isExpanded)} />
                   </rb.Offcanvas.Body>
                 </rb.Navbar.Offcanvas>
                 <rb.Container className="d-none d-md-flex flex-1 flex-grow-0 align-items-stretch">
@@ -266,7 +239,7 @@ export default function Navbar() {
                 </rb.Container>
                 <rb.Container className="d-none d-md-flex flex-1 align-items-stretch">
                   <div className="ms-auto d-flex align-items-stretch">
-                    <TrailingNav joiningRoute={joiningRoute} />
+                    <TrailingNav />
                   </div>
                 </rb.Container>
               </>
--- a/src/components/Navbar.jsx
+++ b/src/components/Navbar.jsx
@@ -91,7 +91,7 @@ const CenterNav = ({ makerRunning, schedulerRunning, singleCollaborativeTransact
         >
           <div className="d-flex align-items-start">
             {t('navbar.tab_send')}
-            <TabActivityIndicator isOn={singleCollaborativeTransactionRunning} />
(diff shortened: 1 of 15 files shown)


src/components/Navbar.jsx
```diff
@@ -89,23 +89,32 @@ const CenterNav = ({ makerRunning, cjRunning, onClick }) => {
             'center-nav-link nav-link d-flex align-items-center justify-content-center' + (isActive ? ' active' : '')
           }
         >
-          {t('navbar.tab_send')}
+          <div className="d-flex align-items-start">
+            {t('navbar.tab_send')}
+            <TabActivityIndicator isOn={singleCollaborativeTransactionRunning} />
```
dnlggr, line 94: Can you give this a 3px left margin? Just to move it away from the "d" of "Send" a bit. Cf.:

Screenshot 2022-08-11 at 15 04 39
Screenshot 2022-08-11 at 15 04 43

dnlggr (commented):
With the activity indicator on the "Send" tab--do we even need the "Joining" indicator anymore? I almost feel like we can remove it entirely now, what do you think?

dnlggr:
With the `schedule` in the session, we can probably also remove the forced schedule polling in the Jam screen right? If so I'll create an issue for that.

theborakompanioni:
> With the activity indicator on the "Send" tab--do we even need the "Joining" indicator anymore? I almost feel like we can remove it entirely now, what do you think?

Either this, or remove the dots from the navbar links - please free me from the burden of making UI related decisions :sweat_smile: 

> With the `schedule` in the session, we can probably also remove the forced schedule polling in the Jam screen right? If so I'll create an issue for that.

Yes, that's true!

dnlggr:
> Either this, or remove the dots from the navbar links - please free me from the burden of making UI related decisions 😅

Let's defer the decision by dropping the "Joining"--I never liked the wording anyway. For now at least. The activity indicators do look a bit crowded up there but they're fine for now. A nice next step would be dropping those as well and replacing them with a nicely looking activity indicator (either in the header or in the footer) similar to the "Joining" thing but one that looks better and differentiates between the three different possible activities. Let's not wait around for designs though since this is a quite important PR (because of the schedule in the session response). What do you think?

dergigi:
> Let's defer the decision by dropping the "Joining"--I never liked the wording anyway.

+1

src/context/ServiceInfoContext.tsx
```diff
@@ -10,19 +10,58 @@ import * as Api from '../libs/JmWalletApi'
...
 type CoinjoinInProgressFlag = { coinjoinInProgress: boolean }
-type WalletName = { walletName: string | null }
 
-type ServiceInfo = SessionFlag & MakerRunningFlag & CoinjoinInProgressFlag & WalletName
+type ServiceInfo = SessionFlag &
+  MakerRunningFlag &
+  CoinjoinInProgressFlag & {
+    walletName: Api.WalletName | null
+    offers: Offer[] | null
+    nickname: string | null
```
dnlggr, line 63:
```suggestion
    nickname: string | null
    schedule: Schedule | null
```

Right? 🤔

src/context/ServiceInfoContext.tsx
```diff
@@ -74,17 +113,28 @@ const ServiceInfoProvider = ({ children }: React.PropsWithChildren<{}>) => {
...
             maker_running: makerRunning,
             coinjoin_in_process: coinjoinInProgress,
             wallet_name: walletNameOrNoneString,
+            offer_list: offers,
+            schedule,
+            nickname,
           } = data
           const activeWalletName = walletNameOrNoneString !== 'None' ? walletNameOrNoneString : null
-          return { sessionActive, makerRunning, coinjoinInProgress, walletName: activeWalletName } as ServiceInfo
+          return {
```
dnlggr, line 129: Should we add `singleCollaborativeTransactionRunning` and `schedulerRunning` directly to `ServiceInfo`? I feel like this could be useful throughout the app (e.g., #269). Maybe even something similar to what we have in `WalletContext` where there's a summary computed from the available raw data? What do you think?

src/context/ServiceInfoContext.tsx
theborakompanioni: Yes, of course! :grimacing:

src/context/ServiceInfoContext.tsx
theborakompanioni: Good idea!

theborakompanioni: Hmm.. thinking about it.., `makerRunning` and `coinjoinInProgress` can be updated via Websocket data.. `schedule` will currently only update from data of an http response.

I guess this deserves some more attention and docs.. will do it in a follow-up PR!

dnlggr: Ok sounds good!

src/components/Navbar.jsx
theborakompanioni: I added spacing to all tab indicators.. as it will look different in other languages.. acceptable?

src/index.css
```diff
@@ -373,7 +373,7 @@ main {
 .earn-indicator .activity-indicator {
   --size: 6px;
   position: absolute;
-  right: calc(var(--size) * -1);
+  right: calc(var(--size) * -1 - 3px);
```
dnlggr, line 376: Sorry for being imprecise, I meant adding the margin only to the "Send" indicator. Since that has the "d" at the end which makes it look crammed. "Earn" and "Jam" are fine.

theborakompanioni: I know what you mean, but this is depends on the language.. should we really do that?

src/components/Navbar.jsx
dnlggr: Oh ok that's a good point. Hmm. I'm tempted to say let's optimize for English. The other languages will look okish but I bet 95% will use English anyway.

theborakompanioni: Okay. Added class name "ms-1" instead of 3px to the Send link.. hope thats okay.

Maybe we can come up with a more generic indicator that looks good for all languages.

dnlggr: Yep, we'll come up with something different anyway: https://github.com/joinmarket-webui/jam/pull/463#issuecomment-1212111322

dnlggr: ms-1 is a bit much but let's leave it like that for now.

dnlggr (approved):

dergigi:
Ah, damn, just saw this now.

The way I understood it is keep the icon, but remove the "Joining" text 😬 

@theborakompanioni what do you say? I think having the icon (and the functionality added in this PR that leads you to the appropriate page when you click on it) makes a ton of sense.

theborakompanioni:
> @theborakompanioni what do you say? I think having the icon (and the functionality added in this PR that leads you to the appropriate page when you click on it) makes a ton of sense.

Can do! It is a matter of reverting https://github.com/joinmarket-webui/jam/pull/463/commits/de86014d030682baf947d9591311ef37fc6c55c2 to come back to the original functionality. The "Joining" text will only appear in the mobile menu, as it would look rather weird without text.

I think having both (activitiy inidicator and the icon on the right side) is fine. Will open a PR.

dergigi:
> Will open a PR.

Excellent, thank you!

theborakompanioni:
> > Will open a PR.
> 
> Excellent, thank you!

See https://github.com/joinmarket-webui/jam/pull/474

PR #464, opened 2022-08-10
ui: remove column 'fee' from Earn Report

theborakompanioni:
As suggested by @openoms in the [Community Call #21](https://bitcointv.com/w/tcebpxai765AnALer87fsU?start=17m16s) (at 17m16s), this commit removes the "Fee" column from the Earnings Report.

Since the Miner Fee Contribution is always zero, this value (`fee`) is the same as "earned" (`earnedAmount`).


dergigi (approved):
Looks good to me ✅ (utACK)

dnlggr (approved):

PR #467, opened 2022-08-11
dev: bullseye as base image for regtest containers

theborakompanioni:
Companion change of https://github.com/joinmarket-webui/jam-docker/pull/52 for the regtest environment.

The development environment should as closely as possible incorporate all changes of the prod image.
Neat side effect is, that the images get smaller, even if that is not a high priority objective for the dev setup.


dergigi:
> The development environment should as closely as possible incorporate all changes of the prod image.

100% agree.

> Neat side effect is, that the images get smaller, even if that is not a high priority objective for the dev setup.

Nice! Smaller is always good 👌

dergigi (approved):
Looks good to me ✅

PR #470, opened 2022-08-15
refactor: leverage session request for polling schedule

### LINKED ISSUE #468: Remove schedule polling on Jam page
The schedule (if one exists) is now part of the `/session` response. See #463. That means we can remove the schedule polling on the Jam page and use the schedule in the session for progress reports.

theborakompanioni:
Resolves #468.

Uses new `schedule` property of the `serviceInfo` object to display the scheduler progress instead of polling the `/schedule` endpoint.


dergigi (approved):
Looks good to me. tACK ✅

PR #472, opened 2022-08-15
feat: highlight own orders in orderbook

### LINKED ISSUE #413: Earn: Show counterparty ID to the user

theborakompanioni:
Closes #413.

This does not actually display the nickname (counterparty ID), but lets the user highlight his own orders, as the nickname is already shown on the Earn page (see #461).

First attempt was to just filter the orderbook for the users orders by using the nickname as search value.
However, opted for highlighting the row instead, as this provides the functionality to sort the data and provide more insights and better hints about the positions of your own orders within the orderbook. Let me know what you think.

### DIFF
--- a/src/components/Earn.jsx
+++ b/src/components/Earn.jsx
@@ -521,7 +521,11 @@ export default function Earn() {
       </rb.Row>
       <rb.Row className="mt-5 mb-3">
         <rb.Col className="d-flex justify-content-center">
-          <OrderbookOverlay show={isShowOrderbook} onHide={() => setIsShowOrderbook(false)} />
+          <OrderbookOverlay
+            show={isShowOrderbook}
+            onHide={() => setIsShowOrderbook(false)}
+            nickname={serviceInfo?.nickname}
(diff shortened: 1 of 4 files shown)


dergigi:
> opted for highlighting the row instead, as this provides the functionality to sort the data and provide more insights and better hints about the positions of your own orders within the orderbook.

Excellent, I think that makes a lot of sense. 🧐👍

dergigi (approved):
Well done, looks and feels great. This was much needed imho, makes things way more transparent and easier to track for the user. Code looks good & tested on regtest.

LGTM ✅

PR #486, opened 2022-08-25
fix: use orderbook.json instead of parsing html table

### LINKED ISSUE #454: Use `orderbook.json` instead of parsing html table
Currently, Jam parses the orders from the table of the orderbook html page.
There is an endpoint returning this data in json format, but this has not been used till now as it returned empty responses in our development environment. This issue is now located and reported here: https://github.com/JoinMarket-Org/joinmarket-clientserver/issues/1327
Switching to the json endpoint would do away with this (silly) workaround.

Either a) wait for the fix to be implemented (then the issue is blocked) or b) use parsing the html as fallback if the response is empty - then remove the fallback when the upstream issue is resolved.

theborakompanioni:
TBD.

Closes https://github.com/joinmarket-webui/jam/issues/454.

Not ready yet. Needs to be refactored.

### DIFF
--- a/src/components/Orderbook.tsx
+++ b/src/components/Orderbook.tsx
@@ -76,23 +76,82 @@ const TABLE_THEME = {
   `,
 }
 
+type OrderTypeProps = {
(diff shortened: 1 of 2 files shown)


theborakompanioni:
Superseded by https://github.com/joinmarket-webui/jam/pull/687.

PR #487, opened 2022-08-26
fix: create non-descriptor wallet

theborakompanioni:
Makes sure the created wallet is a non-descriptor wallet.

Changes:
- apply same changes as in https://github.com/joinmarket-webui/jam-docker/pull/58
- replace nbxplorer with own "regtest-initializer"
- update bitcoin core from v22 to v23
- add new npm script `npm run regtest:build` (make use of docker cache)

## How to test
If you just run `npm run regtest:up` you will see that the primary container will start successfully, but the secondary container will fail. This is because if you have not rebuilt your local dev environment (without changes from https://github.com/joinmarket-webui/jam-docker/pull/58), the secondary container defaults to creating a descriptor wallet and fails (which is exactly what is fixed in this PR).

Since https://github.com/joinmarket-webui/jam-docker/pull/58 is already merged, all you have to do is building the images and then start as usual.
1. `npm run regtest:build`
2. `npm run regtest:up`
3. `npm run regtest:init`
4. `npm start`


docker/regtest/docker-compose.yml
```diff
@@ -108,9 +108,9 @@ services:
   bitcoind:
     container_name: jm_regtest_bitcoind
     restart: unless-stopped
-    image: btcpayserver/bitcoin:22.0-1
+    image: btcpayserver/bitcoin:23.0
```
dergigi, line 111: 👍

docker/regtest/docker-compose.yml
```diff
@@ -99,7 +99,7 @@ services:
   irc:
     container_name: jm_regtest_irc
     restart: unless-stopped
-    image: ghcr.io/ergochat/ergo:stable
+    image: ghcr.io/ergochat/ergo:v2.10.0
```
dergigi, line 102: What's the benefit of switching to an explicit version here, instead of `stable` or `latest`?

docker/regtest/readme.md
```diff
@@ -4,6 +4,9 @@ This setup will help you set up a regtest environment quickly.
 It starts multiple JoinMarket containers, hence not only API calls but also actual CoinJoin transactions can be tested.
 Communication between these containers is done via Tor (if internet connection is available) and IRC (locally running container).
 
+Both containers will have a wallet named `Satoshi.jmdat` with password `test`.
+The second container has basic auth enabled (username `joinmarket` and password `joinmarket`).
```
dergigi, line 8: 📝 👀     👍

package.json
```diff
@@ -58,7 +58,8 @@
     "format": "prettier --write --no-error-on-unmatched-pattern 'src/**/*.{js,jsx,ts,tsx,json,css,md}'",
     "version": "node scripts/changelog.mjs && git checkout -b \"prepare-v${npm_package_version}-$(date +%s)\" && git add --all && git commit --message \"chore(release): prepare v${npm_package_version}\" && git push --set-upstream origin $(git branch --show-current)",
     "postversion": "which gh && gh pr create --title \"chore(release): prepare v${npm_package_version}\" --body \"Prepares the v${npm_package_version} release.\" --assignee @me --label release --repo joinmarket-webui/joinmarket-webui --draft",
-    "regtest:rebuild": "npm run regtest:clear && docker-compose --env-file docker/regtest/.env.example --file docker/regtest/docker-compose.yml build --pull --no-cache",
+    "regtest:build": "npm run regtest:clear && docker-compose --env-file docker/regtest/.env.example --file docker/regtest/docker-compose.yml build --pull",
+    "regtest:rebuild": "npm run regtest:build -- --no-cache",
```
dergigi, line 62: [TIL](https://stackoverflow.com/a/44743871)!

dergigi (approved):
tACK, LGTM ✅

docker/regtest/docker-compose.yml
theborakompanioni: I though it might prevent unintended side-effects of newer versions in the future. Kind of like "version pinning". Dont want to run in the situation where the regtest setup fails for users just to figure out the irc mechansim is not working - suppose it is rather hard to find if one did not regenerate the setup for a while. It not an important change and I can revert it, if you think we should stay on the `stable` image.

package.json
theborakompanioni: Yeah.. true that.. will fix!

theborakompanioni: Ahh.. I thought this was a change suggestion : D
Already applied..

PR #490, opened 2022-09-09
feat(navbar): align app flow 

### LINKED ISSUE #488: Align general application flow with scheduler changes
The reason why Jam is structured [the way it is](https://jamdocs.org/interface/00-cheatsheet/), is because we wanted to have an easy way to create `cj-out` UTXOs to create Fidelity Bonds for the **Earn** screen. That's why the default "app flow" is as follows:

```
Receive  »  Jam (Scheduler)  /  Earn  »  Send
```

With the recent Scheduler changes, this has to change. The Scheduler now always sweeps to external addresses, which means it is an automated way to *sweep the whole wallet* (all non-frozen funds) to addresses you control.

With #483 in mind, the updated flow could be simplified into `Receive  »  Earn`, since creating a `cj-out` UTXO for the FB will be part of the Earn flow. 

Thus, the updated flow should be:

```
Receive  »  Earn  »  Send  |  Sweep
```

...with `Sweep` being the scheduler. 

In the end, I think `Sweep` should be part of `Send`, because a sweep is just a special type of send, and the scheduler is just a special kind of sweep (with multiple targets & random splits). The main use-case of the scheduler will probably be to sweep to cold storage or to Lightning, since it will usually be a wallet you control.

I'd suggest we: 

- [x] update the flow (#490) as suggested above keeping most things as they are,
- [x] update the docs (https://github.com/joinmarket-webui/jamdocs/pull/12), and
- [x] update the cheatsheet (#496).

#483 can be another increment, and "merging" the scheduler with "send" can be another increment.

---

P.S: Not sure about terminology. "Sweep" is taken from JoinMarket—if we want to stick with the "jam" metaphors, something else might make more sense. "Flush" comes to mind, as it "cleans" all the jars.

dergigi:
Resolves #488 

Minimal changes, as discussed:

- Align flow in nav bar (`Receive  »  Earn  »  Send  |  Sweep`)
- Rename "Jam" to "Sweep"
- Rename "Jam Scheduler" to "Scheduled Sweep"

Screenshot 2022-09-09 at 13 52 32

### DIFF
--- a/src/components/Navbar.jsx
+++ b/src/components/Navbar.jsx
@@ -60,7 +60,7 @@ const CenterNav = ({ makerRunning, schedulerRunning, singleCollaborativeTransact
           }
         >
           <div className="d-flex align-items-start">
-            {t('Jam')}
+            {t('Sweep')}
             <TabActivityIndicator isOn={schedulerRunning} />
           </div>
         </NavLink>
--- a/src/components/Navbar.jsx
+++ b/src/components/Navbar.jsx
@@ -53,21 +53,6 @@ const CenterNav = ({ makerRunning, schedulerRunning, singleCollaborativeTransact
       <div className="d-none d-md-flex align-items-center center-nav-link-divider">»</div>
       <rb.Nav.Item className="d-flex align-items-stretch">
         <NavLink
-          to={routes.jam}
-          onClick={onClick}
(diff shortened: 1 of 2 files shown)


theborakompanioni (approved):
:heavy_check_mark: :+1:

PR #491, opened 2022-09-13
chore(i18n): add missing translations for fidelity bonds form

theborakompanioni:
Small i18n fixes:
- add missing translation for key `earn.fidelity_bond.review_inputs.text_primary_button_error`
- fix typo in key `.error.loading_address` -> `.error_loading_address`

This has come up when trying to create a fidelity bond with a wallet that does not support fidelity bonds. Jam, by default, will always generate fb-enabled wallets. If we want to prevent users with unsupported wallets from even seeing the form, Jam would have to make an additional http call prior to that - so it might not be worth it.
Question is: Should we append the phrase "Your wallet might not support fidelity bonds" or "Please make sure your wallet supports fidelity bonds", or something like that?

### DIFF
--- a/src/components/fb/CreateFidelityBond.jsx
+++ b/src/components/fb/CreateFidelityBond.jsx
@@ -265,7 +265,7 @@ const CreateFidelityBond = ({ otherFidelityBondExists, accountBalances, totalBal
         }
 
         if (timelockedAddress === null) {
-          return <div>{t('earn.fidelity_bond.error.loading_address')}</div>
+          return <div>{t('earn.fidelity_bond.error_loading_address')}</div>
         }
 
         return (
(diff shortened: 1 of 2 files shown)


dergigi:
> This has come up when trying to create a fidelity bond with a wallet that does not support fidelity bonds.

I don't understand. Can you elaborate?

theborakompanioni:
> I don't understand. Can you elaborate?

Fidelity Bonds have been introduced with JM v0.9.0 (released on 2021-07-31).
e.g. if you recover a wallet with `python3 scripts/wallet-tool.py recover` you'll get asked `Would you like this wallet to support fidelity bonds? write 'n' if you don't know what this is (y/n)`. If you opt-out, trying to create a timelocked FB address will yield an error.

dergigi:
Ah, I see. Thanks for the clarification.

> Question is: Should we append the phrase "Your wallet might not support fidelity bonds" or "Please make sure your wallet supports fidelity bonds", or something like that?

I think it's fine as it is. The error isn't perfect, but I'd consider this an edge case and we can elaborate in the docs.

dergigi (approved):
Looks good to me ✅

PR #492, opened 2022-09-13
fix: redirect to home if no wallet is active on route `/wallet`

theborakompanioni:
Before this commit, a user that logs into an active wallet from a second tab/browser, had the first tab/browser not redirect to the unlock wallet page if route `/wallet` was active.
After this commit, the will redirect to the "unlock wallet" page (like all other routes, e.g. `/jam`, `/send`, etc.).

This has really only been the case for route `/wallet`.

It can be tested by navigating to `/wallet` and unlocking the wallet from a second browser/tab.
It should then redirect properly. The secondary instance can be used as a regression test - it still has this bug included.

Additional fix coming out of this: Wallet name on unlock page should only be a link to `/wallet` if the wallet is active **and** unlocked by the current user.

### DIFF
--- a/src/components/App.jsx
+++ b/src/components/App.jsx
@@ -82,6 +82,7 @@ export default function App() {
                 />
                 {currentWallet && (
                   <>
+                    <Route path={routes.wallet} element={<MainWalletView />} />
                     <Route path={routes.jam} element={<Jam />} />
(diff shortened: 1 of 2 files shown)


dergigi (approved):
Looks good to me, utACK ✅

PR #493, opened 2022-09-13
ui: use 'zoom-in' cursor when hovering over jars on main page

theborakompanioni:
This PR changes the cursor when hovering over Jars to open the detailed view from 'pointer' to 'zoom-in'.

Saw it on Figma and thought it was easy to do.

## :camera_flash: Before/After
 

If you don't think this is a good idea, please feel free to just close the PR.


dergigi (approved):
Looks good to me, tACK ✅

Issue #489, opened 2022-09-09
Scheduler: Success Screen


dergigi:
Nothing happens on success, would be good to show a big fat ✅ or something, at least.

Since this is a long-running process I assume that users will leave it open/running and as it is now, currently, even if it is successful, they will be greeted with the "Scheduler" screen again, and possibly a warning (because of a lack of UTXOs or some such). For example:

![Screenshot 2022-09-09 at 12 52 12](https://user-images.githubusercontent.com/109058/189334619-e64757f5-1455-4395-9a04-cb5e19eab3c6.png)

In the future, it would be great to add a "fee report" or something similar. A summary, like a reverse-earning report. Just a thought.

theborakompanioni:
Good idea.

One difficulty comes to mind: At the moment we cannot really distinguish between a successful and a failed scheduler run (e.g. because of no retries left), other than by applying heuristics: e.g. either none or only frozen utxos left in the wallet after a scheduler run. Still thinking about how to communicate this to the user.

Related: https://github.com/joinmarket-webui/jam/issues/251

editwentyone:
i made a proposal in #516 , maybe that is enough for the time being

Issue #251, opened 2022-05-11
Update balance on Jam screen after stopping scheduled transactions


dnlggr:
See https://github.com/joinmarket-webui/joinmarket-webui/pull/242.

PR #494, opened 2022-09-13
fix: reload wallet info after stopping scheduler manually

theborakompanioni:
Tiny fix: Reloads wallet info after a scheduled sweep is stopped manually. 

Before this PR the displayed wallet balance was outdated and not updated when manually stopping a scheduled sweep.
After this PR the wallet balance will reflect the actual balance after manually stopping a scheduled sweep.

Additional: Removes a unneeded delay when stopping the scheduled sweep.

This does not (fully) address https://github.com/joinmarket-webui/jam/issues/251 or https://github.com/joinmarket-webui/jam/issues/489.


dergigi (approved):
Looks good to me, tACK ✅

PR #497, opened 2022-09-15
feat: abort collaborative transaction

### LINKED ISSUE #269: Ability to stop a running taker operation

theborakompanioni:
Closes #269.

Adds the ability to abort a running taker operation.

### How to test
1. Send a collaborative transaction
2. Try to abort the transaction
3. Verify that the transaction was aborted and not broadcasted

You might want to watch the logs while testing this feature with `npm run regtest:logs:jmwalletd`

Question: Currently, the text only says "Are you sure you want to abort the collaborative transaction?".. 
Stopping a taker operation might lead to end up wasting a "retry" (when commitment was already sourced).. Should it mention that it might lead to inconveniences if this is used "too much in a row"? How to communicate this best?


dergigi:
> Question: Currently, the text only says "Are you sure you want to abort the collaborative transaction?"..
Stopping a taker operation might lead to end up wasting a "retry" (when commitment was already sourced).. Should it mention that it might lead to inconveniences if this is used "too much in a row"? How to communicate this best?

Let's leave it as is and see how many users run into these kinds of issues in the wild.

dergigi (approved):
Looks good to me, tACK ✅

dergigi:
Notes to self: 

* "Abort" sounds quite drastic, maybe reword to "Cancel"?
* "Collaborative transaction started" toast seems unnecessary now; remove?

PR #498, opened 2022-09-16
fix: do not disable submit button on warnings on Send page

theborakompanioni:
Before this PR, the submit button on the Send page has been disabled, if the requirements for collaborative transactions were not met.
After this PR, the warnings are still displayed, but the button is enabled and a payment can be attempted.

Why this change? The functionality has been disabled, as single collaborative transactions have no retry mechanism in place (unlike the scheduled sweep) and users had no real hint on how to recover, in case a transaction gets "stuck". With #497 resolved, users can now abort/cancel single collaborative transactions, hence the button can safely be enabled.

However, if preconditions are not fulfilled, the transaction will likely either get "stuck" (e.g. when a utxo with <5 confs is used) or will be aborted by the server (20% rule, no retries left) - there is no good way of communicating this to the user other than the current warnings.

Additional Change:
- "Collaborative transaction started" toast has been removed, as per this comment https://github.com/joinmarket-webui/jam/pull/497#issuecomment-1249059814

Question: The warning texts have not been changed. Do you think they should be adapted and need rephrasing?

### DIFF
--- a/src/components/Send.jsx
+++ b/src/components/Send.jsx
@@ -240,7 +240,6 @@ export default function Send() {
 
   const [waitForUtxosToBeSpent, setWaitForUtxosToBeSpent] = useState([])
   const [paymentSuccessfulInfoAlert, setPaymentSuccessfulInfoAlert] = useState(null)
-  const [takerStartedInfoAlert, setTakerStartedInfoAlert] = useState(null)
 
   const isOperationDisabled = useMemo(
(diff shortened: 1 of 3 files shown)


dergigi:
Made a small adjustment to be more explicit in the button text:

Screenshot 2022-09-16 at 13 58 25

dergigi (approved):
Looks good to me, tACK ✅

PR #501, opened 2022-09-16
fix: consistent button text during screen initialization

theborakompanioni:
Consistent button labeling on Send page.

Additionally, change text from  "Ignore warning & send" to "Ignore warning & try send".


dergigi (approved):
Looks good to me, tACK ✅

PR #504, opened 2022-09-18
dev: convert app.jsx to app.tsx

cstenglein:
This PR converts `App.jsx` to `App.tsx`.

The `variant` in `Layout` was mandatory but not used, so I set a default value for it.

My IDE also reordered the imports, hope that is fine.


theborakompanioni:
> This PR converts `App.jsx` to `App.tsx`.

Nice!

> The `variant` in `Layout` was mandatory but not used, so I set a default value for it.
> 
> My IDE also reordered the imports, hope that is fine.

That's totally fine. Guess we should do the same. What would you suggest using? Is it possible to do this in a pre-commit hook?

I thought if using JSX, you always need `import React` - seems it is no longer necessary (?) - good!

Thanks for your work :orange_heart:

theborakompanioni (approved):

cstenglein:
Closed due to regulatory concerns. Thank you @theborakompanioni

PR #505, opened 2022-09-19
build: update dependencies

theborakompanioni:
Update dependencies.

```
 serve                                      13.0.2  →    14.0.1     
 @emotion/react                            ^11.9.3  →  ^11.10.4     
 @popperjs/core                            ^2.11.5  →   ^2.11.6     
 @table-library/react-table-library        ^4.0.10  →   ^4.0.18     
 @testing-library/jest-dom                 ^5.16.4  →   ^5.16.5     
 classnames                                 ^2.3.1  →    ^2.3.2     
 i18next                                   ^21.8.4  →   ^21.9.2     
 i18next-browser-languagedetector           ^6.1.4  →    ^6.1.5     
 jest-websocket-mock                        ^2.3.0  →    ^2.4.0     
 react-bootstrap                     ^2.5.0-beta.1  →    ^2.5.0       
 @types/jest                               ^27.5.1  →   ^29.0.3 
 lint-staged                               ^12.4.2  →   ^13.0.3
 prettier                                   ^2.6.2  →    ^2.7.1
 bootstrap                                  ^5.2.0  →    ^5.2.1
 react-i18next                            ^11.16.9  →  ^11.18.6
````

Relevant changes:
https://github.com/twbs/bootstrap/releases/tag/v5.2.1
https://github.com/soldair/node-qrcode/compare/v1.5.0...v1.5.1
https://github.com/i18next/react-i18next/blob/master/CHANGELOG.md

Did not update `react-router-dom` from v6.3.0 to v6.4.0 in this PR on purpose, as this would need some code changes. Will do that one in a follow-up PR.

PR #506, opened 2022-09-20
build(deps): remove web-vitals

theborakompanioni:
[`web-vitals`](https://github.com/GoogleChrome/web-vitals#overview) is a library to measure app metrics.
> [...] for measuring all the [Web Vitals](https://web.dev/vitals/) metrics on real users, in a way that accurately matches how they're measured by Chrome and reported to other Google tools

Jam does not make use of it and there is no reason why it should ever make use of it, hence this PR removes it.

This change is based on #505. Should be merged afterwards.


dergigi (approved):
tACK ✅

PR #507, opened 2022-09-20
fix(navbar): send before earn

dergigi:
I don't know how I got this wrong, but the order I did in #496 has earn before send:

```
Receive  »  Earn  »  Send  |  Sweep
``` 

Because of the new flow it should be **send before earn**, to get `cj-out` for FB:

```
Receive  »  Send  »  Earn  |  Sweep
```

I got it right in [the docs](http://localhost:8000/interface/00-cheatsheet/) and in the cheatsheet (#496) - but I messed it up in the previous PR (#490).

---

### Before 📸

Screenshot 2022-09-09 at 13 52 32

### After 📸

Screenshot 2022-09-20 at 19 55 54


theborakompanioni (approved):
> I'm tempted to do all the screenshots with this change and just act as if it was correct in v0.1 all along - otherwise we'd have to redo all the screenshots for the docs again sweat_smile

:+1: :heavy_check_mark:

PR #509, opened 2022-09-21
refactor: components from jsx to tsx

theborakompanioni:
Small refactoring turning some components from Javascript to TypeScript.

- `JarSelectorModal`
- `Alert`
- `ActivityIndicators`
- `App` and `App.test`
- `setupTests.js` to `setupTests.ts`

No functionality has been changed.


src/components/fb/CreateFidelityBond.jsx
```diff
@@ -509,7 +509,7 @@ const CreateFidelityBond = ({ otherFidelityBondExists, accountBalances, totalBal
 
   return (
     <div className={styles.container}>
-      {alert && <Alert {...alert} className="mt-0" onDismissed={() => setAlert(null)} />}
+      {alert && <Alert {...alert} className="mt-0" onClose={() => setAlert(null)} />}
```
dergigi, line 512: I assume there is no big difference between `onDismissed` and `onClose`?

dergigi (approved):
Looks good to me, tACK ✅

src/components/fb/CreateFidelityBond.jsx
theborakompanioni: This is just trying to be more consistent with the react-bootstrap components.
Argument is now the original arg + a string message: `export type SimpleMessageAlertProps = BsAlertProps & { message: string }`

Otherwise users would be able to pass both `onClose` and `onDismissed` with just the latter being actually used.

PR #510, opened 2022-09-21
feat(send): show txid on successful direct-send

### LINKED ISSUE #499: Send: show transaction id `txid` after successful direct send
Here is what is currently shown:

```
Payment successful: Sent 123123 sats to bcrt1q8z8alm0yljxuzm6a20xcqtdud83ny73v08zf6a.
```

The API does return `txid` for a [/directsend,](https://joinmarket-org.github.io/joinmarket-clientserver/api/#operation/directsend) which we should show in addition to the information above.

theborakompanioni:
Resolves #499.

Applying classes `small`, `slashed-zeroes` and `break-word` to make sure the text is readable on all screens (and not clipped).
In next iterations, this is message is better presented as toast message, which has been an idea for a while (not done in this PR).

### DIFF
--- a/src/components/Send.jsx
+++ b/src/components/Send.jsx
@@ -440,12 +440,16 @@ export default function Send() {
 
       if (res.ok) {
         const {
-          txinfo: { outputs, inputs },
+          txinfo: { outputs, inputs, txid },
         } = await res.json()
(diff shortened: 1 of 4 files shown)


dergigi (approved):
Looks good to me, tACK ✅

PR #514, opened 2022-09-22
feat: show sum of selected utxos

### LINKED ISSUE #440: Show sum of selected UTXOs in UTXO list
> Use case is e.g. to sweep an approximate amount to cold storage or spend the funds otherwise as payment. As a user, I do not want to sum it up manually. 

See https://github.com/joinmarket-webui/jam/pull/430#issuecomment-1198694180.

theborakompanioni:
Resolves #440.

Changes:
- show sum of selected utxos
- adapt margins/paddings to better match figma screens
- wording according to figma:
  - "Jar Details" -> "Details"
  - "Freeze/unfreeze selected" -> "Freeze/unfreeze"

### DIFF
--- a/src/components/jar_details/JarDetailsOverlay.module.css
+++ b/src/components/jar_details/JarDetailsOverlay.module.css
@@ -76,12 +76,13 @@
 .overlayContainer .tabContainer {
   display: flex;
   flex-direction: column;
-  gap: 2.5rem;
+  gap: 0.5rem;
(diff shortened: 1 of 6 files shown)


dergigi (approved):
Future improvement: If nothing is selected, we could show the total amount that is in the jar in the same place. Tracked as #517 

No selection: `12,773,047,386 sats in total`
With selection: `28,044 sats selected`

But this looks good as is. tACK ✅

PR #519, opened 2022-09-26
deps: update `react-router-dom` from v6.3.0 to v6.4.1

theborakompanioni:
Updates dependency `react-router-dom` from v6.3.0 to v6.4.1.

Originally wanted to leverage the [new `loader` feature](https://reactrouter.com/en/main/route/loader#loader) to separate data loading from view rendering.
This will be done in a follow-up PR as we often update context data which cannot be done in a satisfactory way yet (see https://github.com/remix-run/react-router/issues/9324).

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -1,7 +1,7 @@
 import { useCallback } from 'react'
 import * as rb from 'react-bootstrap'
 import { useTranslation } from 'react-i18next'
-import { Navigate, Route, Routes } from 'react-router-dom'
+import { createBrowserRouter, createRoutesFromElements, Navigate, Route, RouterProvider } from 'react-router-dom'
 import { routes } from '../constants/routes'
 import { useSessionConnectionError } from '../context/ServiceInfoContext'
 import { useSettings } from '../context/SettingsContext'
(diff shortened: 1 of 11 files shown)


src/components/App.tsx
```diff
@@ -40,6 +40,62 @@ export default function App() {
...
+              <>
+                <Route id="wallet" path={routes.wallet} element={<MainWalletView />} />
+                <Route id="jam" path={routes.jam} element={<Jam />} />
+                <Route id="send" path={routes.send} element={<Send />} />
+                <Route id="earn" path={routes.earn} element={<Earn />} />
+                <Route id="receive" path={routes.receive} element={<Receive />} />
+                <Route id="settings" path={routes.settings} element={<Settings stopWallet={stopWallet} />} />
+              </>
+            )}
+            <Route id="404" path="*" element={<Navigate to={routes.home} replace={true} />} />
```
dergigi, line 89: It seems like the 404 redirect does not work, here is what is shown to me when I navigate to a route that doesn't exist:

Screenshot 2022-09-28 at 14 46 20

dergigi (changes requested):
Looks good to me in general, but we should address the 404 error (unless it's something I'm doing wrong on my side).

src/components/App.tsx
theborakompanioni: Hmm... cannot reproduce currently. If I navigate to an non-existing route by typing it into the address bar, it will take me to `/` (as expected).
Sorry for my stupid question, but how did you navigate to the non-existing route?

I have only seen this meesage, when I abort `npm start`.. 
.. did you execute `npm install`?

dergigi: > (unless it's something I'm doing wrong on my side)

> did you execute `npm install`?

🤦‍♂️

dergigi (approved):
tACK, Looks good to me ✅

src/components/App.tsx
theborakompanioni: My bad. Reminder for myself: Always include a "How to test" section in PRs.

theborakompanioni: I have seen this screen now myself for a very brief moment when reloading a page. Will look into it!

PR #520, opened 2022-09-26
regtest: start additional maker service

theborakompanioni:
This PR adds a third JM regtest instance and starts an additional maker service during initialization with `npm run regtest:init`.
It is an attempt to make the development setup more resilient to maker timeouts. Setup script takes slightly longer and cpu usage during development will increase a bit.

Also adapted some regtest settings for faster/easier testing:
- actually set `maker_timeout_sec` on primary instance -> was `maker_timeout_sec` instead of `jm_maker_timeout_sec`
- set `taker_utxo_age` from 5 to 1 on primary instance (this revealed that coinjoin precondition warning message must be parameterized)
- do not set `tx_broadcast` - was `self` (use default value `random-peer`); you'll see more "tx already in chain" warnings, which is expected behaviour.

#### How to test
Start and initialize your dev environment as usual and try to send a collaborative transaction with 2 collaborators.

### DIFF
--- a/docker/regtest/common.sh
+++ b/docker/regtest/common.sh
@@ -189,7 +189,7 @@ fetch_available_wallets() {
 }
 
 # --------------------------
-# Verify no open session
+# Fetch Session Information
 # --------------------------
 ## API: GET /api/v1/session
 ## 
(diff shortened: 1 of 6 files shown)


dergigi (approved):
tACK, Looks good to me ✅

PR #521, opened 2022-09-26
refactor: jsx to tsx

theborakompanioni:
Small refactoring - most notably passing a `wallet` object to every component that needs an active wallet. This should indicate and make it more obvious that the components do not function without an active wallet.

No functional changes have been made.


dergigi (approved):
tACK, Looks good to me ✅

Issue #518, opened 2022-09-26
Send: fee transparency & ability to set fees


dergigi:
Two things:

- [x] It should be transparent how much a transaction is expected to cost in terms of fees #532 
- [x] Users should have the ability to manually set #522 

Keep in mind that, according to [waxwing](https://t.me/JoinMarketWebUI/4460),

> Joinmarket's design doesn't really allow the concept of 'try coinjoin with a small test amount' and then extrapolate to larger sizes. Such a large fee as you saw is unusual today, but very possible. With amounts in the thousands of dollars, percentage fees can instead be really negligible. 

See also:

- [Docs: Fees](https://jamdocs.org/market/fees/)
- [JoinMarket: A note on fees](https://github.com/JoinMarket-Org/joinmarket-clientserver/blob/master/docs/tumblerguide.md#a-note-on-fees)

PR #522, opened 2022-09-27
feat: basic fee settings

theborakompanioni:
Partly addresses #518.

This PR is a first attempt to let users adapt some (not all) fee related settings:
Namely:
- [`tx_fees`](https://github.com/joinmarket-webui/jam/blob/v0.1.0/docker/regtest/dockerfile-deps/joinmarket/latest/default.cfg#L132-L141)
- [`tx_fees_factor`](https://github.com/joinmarket-webui/jam/blob/v0.1.0/docker/regtest/dockerfile-deps/joinmarket/latest/default.cfg#L143-L147)
- [`max_cj_fee_abs`](https://github.com/joinmarket-webui/jam/blob/v0.1.0/docker/regtest/dockerfile-deps/joinmarket/latest/default.cfg#L175-L181)
- [`max_cj_fee_rel`](https://github.com/joinmarket-webui/jam/blob/v0.1.0/docker/regtest/dockerfile-deps/joinmarket/latest/default.cfg#L183-L189)

Please help with the fee settings description if you can. Most of it is taken from JM config file comments with just minor adaptions. Any improvement is highly appreciated.

Questions on arbitrarily chosen min/max values (mainly to prevent unintended inputs):
- Min value of `tx_fees_factor`: Current implementation has min. of 10%. Should values below be allowed? e.g. zero means, that no fee randomization is done. [Zero is also handled fine by JM](https://github.com/JoinMarket-Org/joinmarket-clientserver/blob/babad1963992965e933924b6c306ad9da89989e0/jmclient/jmclient/blockchaininterface.py#L465-L483
).
- Max value of `max_cj_fee_rel`: Currently max of `50%`. Increase/decrease?
- Max value of `tx_fees`: Currently max of `100 sats/vbyte`. Increase/decrease? No upper limit? There is a config value `absurd_fee_per_kb` which translates to `350 sats/vbyte` by default.. maybe take this settings value as an upper limit?

#### Tests
- setting tx fee rate manually to `1.12 sats/vbyte` with 12% rate variance:
```
2022-09-26 03:01:58,174 [INFO]  Using this manually set tx feerate (randomized for privacy): 1190 sat/vkB (1.1 sat/vB).
```
- setting tx fee rate manually to `42 blocks`:
```
2022-09-26 03:41:52,464 [INFO]  Using bitcoin network feerate for 42 block confirmation target (randomized for privacy): 10000 sat/vkB (10.0 sat/vB)
```
- setting `max_cj_fee_abs` to `249 sats`) (below the default order with abs fee 250sats) and `max_cj_fee_rel` to `0.0002%` to trigger an error when attempting a collaborative tx with `123456789 sats`:
```
2022-09-28 10:47:45,191 [INFO]  INFO:Received offers from joinmarket pit
2022-09-28 10:47:45,192 [WARNING]  ERROR not enough liquidity in the orderbook n=1 suitable-counterparties=0 amount=27300 totalorders=0
2022-09-28 10:47:45,193 [INFO]  Taker not continuing after receipt of orderbook
2022-09-28 10:47:45,193 [INFO]  Coinjoin did not complete successfully.
```
   - after increasing `max_cj_fee_abs` to `251sats`: Success!


dergigi:
Some questions that came up during my tests:

- ~~Why can't I set it to 1 sat/vByte?~~
- Why link the sats/vByte fee to the randomized percentage? When set to 20% I have to set the fee to `1.2`, which is very weird. It would be much more intuitive to just set it to `1` and the maximum of `1.2` is implicit in the 20% setting. Set the minimum, percentage defines the maximum.

Validation issues:

- `1.2` works, but `1,2` doesn't. We should make sure that inputs are i18n agnostic
- I can't set min abs fee to `0` but I can set it to`0.00000000001` which is actually `0`

JoinMarket issues:

> Both limits given in this section, relative and absolute, must be exceeded in order to not consider a certain offer.

I trust that this is true, but it is very weird behavior, isn't it? I would assume that if I set limits, breaking either limit would be a deal-breaker. Example: I don't ever want to pay more than a million sats, and I don't ever want to pay more than 10% in fees. If the order is large enough, a small percentage might be way more than a million sats, which makes one of the two rules I set obsolete.

Just thinking out loud here, not sure if I'm the weird one or if the way that JoinMarket works is the weird one.

dergigi (commented):
Some i18n inconsistencies in terms of number styles.

src/i18n/locales/en/translation.json
```diff
@@ -168,7 +169,35 @@
...
+      "description_tx_fees_satspervbyte": "The fee in satoshi per vbyte used to pay for transactions. A transaction without privacy improvements with one input and two outputs is about 250 bytes, a collaborative transaction can be a multiple of that depending on the number of collaborators and their inputs.",
+      "feedback_invalid_tx_fees_blocks": "Please provide a valid block target between {{ min }} and {{ max }}.",
+      "feedback_invalid_tx_fees_satspervbyte": "Please provide a valid transaction fee in sats/vByte between {{ min }} and {{ max }}.",
+      "label_tx_fees_factor": "Fee randomization",
+      "description_tx_fees_factor": "Rate variance randomization around the transaction fee. If you set your fee target to 10 sats/vbyte and the factor to 30%, it might use any value between 7 and 13 sats/vbyte for your transactions. Default is 20%.",
+      "feedback_invalid_tx_fees_factor": "Please provide a valid fee randomization value between {{ min }} and {{ max }}.",
+      "title_max_cj_fee_settings": "Collaborative transaction fee limits",
```
dergigi, line 187:
```suggestion
      "title_max_cj_fee_settings": "Maker fees",
```

src/i18n/locales/en/translation.json
```diff
@@ -154,6 +154,7 @@
     "hide_seed": "Hide seed phrase",
     "reveal_seed": "Reveal seed phrase",
     "show_logs": "Show logs",
+    "show_fee_config": "Show fee settings",
```
dergigi, line 157:
```suggestion
    "show_fee_config": "Set fee limits",
```

src/i18n/locales/en/translation.json
```diff
@@ -168,7 +169,35 @@
     "jam_twitter": "Jam Twitter",
     "confirm_locking_modal_title": "Lock Wallet",
     "confirm_locking_modal_body_earn": "Earn is active. Locking the wallet will stop the service. You can close the browser window to let it run in the background.",
-    "confirm_locking_modal_body_jam": "A collaborative transaction is being executed. Locking the wallet will stop the service. You can close the browser window to let it run in the background."
+    "confirm_locking_modal_body_jam": "A collaborative transaction is being executed. Locking the wallet will stop the service. You can close the browser window to let it run in the background.",
+    "fees": {
+      "title": "Fee settings",
```
dergigi, line 174:
```suggestion
      "title": "Fee limits",
```

src/i18n/locales/en/translation.json
```diff
@@ -168,7 +169,35 @@
...
+      "description": "Adapt the fee settings according to your needs. Please keep in mind, that these settings are only valid for the lifetime of the service. i.e. they will be reset to default values when the service shuts down or on a system reboot.",
```
dergigi, line 175:
```suggestion
      "description": "Adjust the fee limits according to your needs. These settings will be reset to default values when the JoinMarket service restarts, e.g. on a system reboot.",
```

dergigi:
> Current implementation has min. of 10%. Should values below be allowed? e.g. zero means, that no fee randomization is done. [Zero is also handled fine by JM](https://github.com/JoinMarket-Org/joinmarket-clientserver/blob/babad1963992965e933924b6c306ad9da89989e0/jmclient/jmclient/blockchaininterface.py#L465-L483).

Yes. We should allow 0%. Personally, I would want to be able to set everything to 1 sat/vByte if the mempool is empty.

theborakompanioni:
> Some questions that came up during my tests:
> 
>     * ~Why can't I set it to 1 sat/vByte?~
> 
>     * Why link the sats/vByte fee to the randomized percentage? When set to 20% I have to set the fee to `1.2`, which is very weird. It would be much more intuitive to just set it to `1` and the maximum of `1.2` is implicit in the 20% setting. Set the minimum, percentage defines the maximum.

Valid. Taken into account, that JM respects the `mempoolminfee` value of Bitcoin Core, even with variance applied, it will not go down below this value. Other than that, 1 sats/vByte or 1000 sats/kilo-vB (which is the actual value to set in the config), is still not possible, because a value between [1, 1000] will be interpreted as "block target". Everything above (>=1001) is treated as fee per kilo-vByte. The current approach stems [from suggestion here](https://github.com/JoinMarket-Org/joinmarket-clientserver/issues/1360#issuecomment-1259638764).

I guess 1 sats/vbyte can be transformed to 1.001 sats/vbyte as the one and only special case. It will surprise user, though, but your are right that it feels better than not allowing it.

> 
> Validation issues:
> 
>     * `1.2` works, but `1,2` doesn't. We should make sure that inputs are i18n agnostic

I have never seen any form behave like that - is this really an issue?

>     * I can't set min abs fee to `0` but I can set it to`0.00000000001` which is actually `0`

Thanks! Will be fixed.

> 
> **Edit:** I think I answered this myself: the absolute fee is per maker, the relative fee is in total. That makes sense, and with that the weird condition makes sense too. You don't want to overpay a single maker (hence the absolute limit) and you don't want to overpay in total (hence the relative limit).
> 
> **Edit 2:** But the conditional in the way you phrased it still doesn't make sense, because it would override one of the limits. I don't want to overpay, period! Hence it should be:
> 
> > If an offer exceeds either of these limits it will be declined.
> 
> At least in my head that makes sense. Maybe I'm still stupid.

It is taken directly from the [JM config](https://github.com/JoinMarket-Org/joinmarket-clientserver/blob/v0.9.8/jmclient/jmclient/configure.py#L314-L318):
> Both the limits given in max_cj_fee_abs and max_cj_fee_rel must be exceeded in order to not consider a certain offer.

But yes, it is phrased in a confusing way and should definitely be rephrased.

Thanks for all the inputs. Will fix everything addressed asap.

PulpCattel:
@dergigi 

>  I think I answered this myself: the absolute fee is per maker, the relative fee is in total. That makes sense, and with that the weird condition makes sense too. You don't want to overpay a single maker (hence the absolute limit) and you don't want to overpay in total (hence the relative limit).

Where did you get this idea? Both limits are checked for each individual offer.

The code is here https://github.com/JoinMarket-Org/joinmarket-clientserver/blob/master/jmclient/jmclient/support.py#L243-L268

> Example: I don't ever want to pay more than a million sats, and I don't ever want to pay more than 10% in fees. If the order is large enough, a small percentage might be way more than a million sats, which makes one of the two rules I set obsolete.

I think the way to read these settings is: I'm fine paying (to each maker) either at most X sats or at most X% of the CJ amount. In your example, if I'm mixing a big amount where the abs fee is > than 1kk sats, but the % is < than 10%, then I'm telling the software it's fine, it's inside the allowed range.

dergigi:
> Where did you get this idea? Both limits are checked for each individual offer.
>
> The code is here https://github.com/JoinMarket-Org/joinmarket-clientserver/blob/master/jmclient/jmclient/support.py#L243-L268

Thanks, should've checked the code right away. Was a long day.

> I think the way to read these settings is: I'm fine paying (to each maker) either at most X sats or at most X% of the CJ amount.

That makes sense, thank you again.

PulpCattel:
BTW, not saying it's the best possible approach, only that is the current implement one AFAICT. I guess both approaches can have merits.

dergigi:
> a value between [1, 1000] will be interpreted as "block target"

Oh my 😅

> I guess 1 sats/vbyte can be transformed to 1001 sats/kilo-vbyte as the one and only special case. It will surprise user, though, but your are right that it feels better than not allowing it.

I'd be for that. To me it would be more surprising that I can't set it to 1 sat/vB.

> I have never seen any form behave like that - is this really an issue?

Every single calculator app will parse whatever comma-style you use as a comma, no matter if you press `.` or `,`. It's not a critical issue, just annoying. At least we want to be consistent, I tried to input what was suggested as an example (`1,2`) and it failed, because it should've been `1.2`.

I'll have another go at improving the wording a bit once the other issues are fixed.

theborakompanioni:
> I'll have another go at improving the wording a bit once the other issues are fixed.

@dergigi Should be ready for a re-review now. Please adapt the phrases to whatever you think fits best.

> > I guess 1 sats/vbyte can be transformed to 1001 sats/kilo-vbyte as the one and only special case. It will surprise user, though, but your are right that it feels better than not allowing it.
> 
> I'd be for that. To me it would be more surprising that I can't set it to 1 sat/vB.

Adapted! Thx!  Still testing though (and waiting for feedback on https://github.com/JoinMarket-Org/joinmarket-clientserver/issues/1360#issuecomment-1261809147) - but seems good so far.

> > I have never seen any form behave like that - is this really an issue?
> 
> Every single calculator app will parse whatever comma-style you use as a comma, no matter if you press `.` or `,`. It's not a critical issue, just annoying. At least we want to be consistent, I tried to input what was suggested as an example (`1,2`) and it failed, because it should've been `1.2`.
> 

Sorry.. you know.. "don't trust, verify" and all that.. [Gnome Calculator](https://wiki.gnome.org/Apps/Calculator) certainly does not do it. I have not addressed this as of yet, as this is the standard behaviour of html input elements with `type=number`.

dergigi:
I realize that JoinMarket uses the same parameter for "block target" and "sats/vByte", but it feels weird that one overrides the other in the interface.

Example: if I set the block target to 3 blocks, switch to sats/vByte, it will be set to 3 sats/vByte. If I change that now to 10 sats/vByte and switch back to the block target setting, the block target is now set to 10 blocks.

All that being said, I'm tempted to merge this soon and make these kinds of improvements in follow-up PRs.

dergigi:
Note to self: we should mention the defaults of all values in the interface.

src/components/settings/FeeConfigModal.tsx
```diff
@@ -0,0 +1,562 @@
...
+const TX_FEES_BLOCKS_MAX = 1_000
+const TX_FEES_SATSPERKILOVBYTE_MIN: SatsPerKiloVByte = 1_000 // 1 sat/vbyte
+// 350 sats/vbyte - no enforcement by JM - this should be a "sane" max value (taken default value of "absurd_fee_per_kb")
+const TX_FEES_SATSPERKILOVBYTE_MAX: SatsPerKiloVByte = 350_000
+const TX_FEES_SATSPERKILOVBYTE_ADJUSTED_MIN = 1_001 // actual min of `tx_fees` if unit is sats/kilo-vbyte
+const TX_FEES_FACTOR_DEFAULT_VAL = 0.2 // 20%
+const TX_FEES_FACTOR_MIN = 0 // 0%
+const TX_FEES_FACTOR_MAX = 1 // 100%
+const CJ_FEE_ABS_MIN = 1
+const CJ_FEE_ABS_MAX = 100_000_000 // 1 BTC - no enforcement by JM - this should be a "sane" max value
```
dergigi, line 24: More like insane max value 😅

src/components/settings/FeeConfigModal.tsx
```diff
@@ -0,0 +1,562 @@
...
+
+  const submit = async (feeValues: FeeValues, txFeesUnit: TxFeeValueUnit) => {
+    const allValuesPresent = Object.values(feeValues).every((it) => it !== undefined)
+    if (!allValuesPresent) return
+
+    let adjustedTxFees = feeValues.tx_fees!
+    if (txFeesUnit === 'sats/kilo-vbyte') {
+      // There is one special case for value `tx_fees`:
+      // Users are allowed to specify the value in "sats/vbyte", but this might
+      // be interpreted by JM as "targetd blocks". This adaption makes sure
```
dergigi, line 375:
```suggestion
      // be interpreted by JM as "targeted blocks". This adaption makes sure
```

dergigi (approved):
Looks good to me ✅

Great job @theborakompanioni 👏

src/components/settings/FeeConfigModal.tsx
theborakompanioni: Valid statement. Using 1k BTC would be 0.1% fee - seems high. Don't want to restrict it too much, though. Really don't know an acceptable upper value here. 1mil? 100k? Reduce to 10k even?
Should the current orderbook be taken into account?

theborakompanioni: Lowered to 1mil sats.

dergigi (approved):
Final changes look good to me too. I'll go ahead and merge this now.

PR #529, opened 2022-10-03
fix(sweep): wait for scheduler start/stop

theborakompanioni:
Closes #267.

Before this PR, the Sweep screen did not wait for an operation to start/stop and assumed the operation will succeed.
After this PR, the screen will show a loading indicator till the service has been successfully started/stopped.

PR #526, opened 2022-10-03
chore(github): add issue templates

theborakompanioni:
Adds issue templates for "bugs" and "features".

---

## Bug template
**Expected behavior**

**Actual behavior**

A clear and concise description of what the bug is and what actually happens.

**Steps to reproduce the problem**
  1.
  2.
  3.

**Specifications**
  - Version:
  - Platform:
  - Browser:

**Additional context**

Add any other context about the problem here to help explain your problem.
e.g. error logs or screenshots if applicable.

---

## Feature template

**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

PR #530, opened 2022-10-03
fix(sweep): reload wallet info after scheduled sweep

### LINKED ISSUE #251: Update balance on Jam screen after stopping scheduled transactions

theborakompanioni:
Fixes #251.

~~This PR is based on #526 - please merge #526 first.~~
This PR is based on #529 - please merge #529 first.

Before this PR, when a scheduled sweep operation ended successfully, the balance has not been updated.
After this PR, the balance will be updated as soon as the scheduler stops (either naturally or aborted).

## How to test?
Invoke a scheduled sweep to external addresses and wait for completion.
After the scheduler finishes, the balance should be updated and show _zero_ coins left (compare this to the old behaviour which would still display the balance you started with).
You can also observe the api requests in dev tools (F12) and verify that utxos are reloaded when the operation ends.


dergigi:
I assume you mean based on #529, right?

dergigi (approved):
Finally managed to test this successfully. I guess my regtest UTXOs were all messed up and the scheduler didn't like that.

Code looks good to me, tACK ✅

PR #536, opened 2022-10-07
fix(fees): mitigate construction of non-forwardable transactions

theborakompanioni:
When the fee target is low, JM sometimes constructs transactions, which are declined from being relayed. In order to mitigate such situations, the minimum fee target (when provided in sats/vbyte) has been increased from 1 sat/vByte to 3 sats/vByte, and fee randomization lowered from 0 - 100% to 0 - 50%.

This can only mitigate the problem and should be seen as a temporary solution till the situation has been addressed upstream.

See https://github.com/JoinMarket-Org/joinmarket-clientserver/issues/1360#issuecomment-1259638764 and https://github.com/JoinMarket-Org/joinmarket-clientserver/issues/1360#issuecomment-1262295463

### DIFF
--- a/src/components/settings/FeeConfigModal.tsx
+++ b/src/components/settings/FeeConfigModal.tsx
@@ -1,6 +1,6 @@
 import { forwardRef, useRef, useCallback, useEffect, useState } from 'react'
 import * as rb from 'react-bootstrap'
-import { useTranslation } from 'react-i18next'
+import { Trans, useTranslation } from 'react-i18next'
 import { Formik, FormikErrors } from 'formik'
 import classNames from 'classnames'
 import { FEE_CONFIG_KEYS, TxFeeValueUnit, toTxFeeValueUnit, FeeValues, useLoadFeeConfigValues } from '../../hooks/Fees'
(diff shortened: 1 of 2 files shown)


src/components/settings/FeeConfigModal.tsx
```diff
@@ -14,12 +14,28 @@ type SatsPerKiloVByte = number
...
+const TX_FEES_SATSPERKILOVBYTE_MIN: SatsPerKiloVByte = 3_000 // 3 sat/vbyte
 // 350 sats/vbyte - no enforcement by JM - this should be a "sane" max value (taken default value of "absurd_fee_per_kb")
 const TX_FEES_SATSPERKILOVBYTE_MAX: SatsPerKiloVByte = 350_000
 const TX_FEES_SATSPERKILOVBYTE_ADJUSTED_MIN = 1_001 // actual min of `tx_fees` if unit is sats/kilo-vbyte
 const TX_FEES_FACTOR_MIN = 0 // 0%
-const TX_FEES_FACTOR_MAX = 1 // 100%
+/**
+ * For the same reasons as stated above (comment for `TX_FEES_SATSPERKILOVBYTE_MIN`),
+ * the maximum randomization factor must not be too high.
+ * Settling on 50% as a reasonable compromise this the problem is addressed.
```
dergigi, line 35:
```suggestion
 * Settling on 50% as a reasonable compromise until this problem is addressed.
```

dergigi (approved):
tACK, Looks good to me ✅

PR #540, opened 2022-10-10
fix: display error message if backend is unreachable

### LINKED ISSUE #538: Verify 404 routes work properly
It seems like the 404 redirect does not work, here is what is shown to me when I navigate to a route that doesn't exist:

Screenshot 2022-09-28 at 14 46 20

_Originally posted by @dergigi in https://github.com/joinmarket-webui/jam/pull/519#discussion_r982361015_

The 404 route is only applied if the backend is up and running. If there is a problem, it should be indicated with a custom error page instead of the default error page of `react-router-dom`.

theborakompanioni:
Resolves #538.

Changeset is based on #521 (please review/merge #521 first).

Before this PR the app showed a generic error message when the backend was unreachable.
After this PR the app displays an error message with more informative reason, e.g. "No connection to backend: Gateway Timeout.".

At first I tried implementing this behaviour via the recommended `react-router-dom`'s `errorElement` property (as [described in the docs](https://reactrouter.com/en/main/start/tutorial#not-found-data)). However, this did not work for reasons not yet known. I will try it again with a future release, as this is a more appropriate error handling mechanism and also enables better separation of data and views, which would be nice to have.

#### How to test
Checkout the `master` branch and start the app (with `npm start`) without starting the regtest environment - verify that you see a generic error message. Switch to this branch (`git checkout error-page`) and verify that you see a more informative message.


dergigi (approved):
tACK, Looks good to me ✅

PR #549, opened 2022-10-20
fix: do not enable debug features on `npm start`

### LINKED ISSUE #548: fix: default to "mainnet" settings on `npm start`

When a user starts Jam with `npm start` as described in ["Connecting to a local instance"](https://jamdocs.org/software/installation/#connecting-to-a-local-joinmarket-instance), Jam will start in `development` mode and debug features are enabled, e.g. "skipping seed verification" or "insecure scheduler options". 

Instead of adapting the documentation or changing the mechanism for mainnet users, Jam should by default use "mainnet" settings and developers should change their way of starting Jam to enable debug features. This can be done by switching from starting Jam locally with `npm start` to e.g. `npm run dev:start`.

theborakompanioni:
Resolves #548.

Before this PR, starting Jam with `npm start` enabled debug features.
After this PR, `npm start` will behave like if a user built the app with `npm build` and debug features are disabled.

This means a developer must switch to using `npm run dev:start` in order to enable debug features.
If a dev still uses `npm start` the app will behave "as if on mainnet" - debug features are disabled.

This is a compromise and making a dev change his behaviour feels better than making an ordinary user start Jam in a different way.


src/constants/debugFeatures.ts
```diff
@@ -1,22 +1,20 @@
 interface DebugFeatures {
   insecureScheduleTesting: boolean
   allowCreatingExpiredFidelityBond: boolean
+  skipWalletBackupConfirmation: boolean
 }
 
-const devMode = process.env.NODE_ENV === 'development'
+const devMode = process.env.NODE_ENV === 'development' && process.env.REACT_APP_JAM_DEV_MODE === 'true'
```
dergigi, line 7: 👍

package.json
```diff
@@ -48,14 +48,15 @@
     "serve": "14.0.1"
   },
   "scripts": {
+    "dev:start": "REACT_APP_JAM_DEV_MODE=true npm start",
```
dergigi, line 51: Nice and simple, I like it 👍

dergigi (approved):
Clean and simple solution, well done! 👏 

tACK ✅

Issue #547, opened 2022-10-20
feat: spend fidelity bond


theborakompanioni:
**Is your feature request related to a problem? Please describe.**
If a wallet contains an **expired** fidelity bond, as soon as it is **unfrozen**, no other action than direct-sends (maker/taker operation) are possible anymore from Jar A (where the FB resides).

**Describe the solution you'd like**
A concrete way to ~~renew or~~ spend the FB without unfreezing it manually via the UI.
Prevent FBs from being unfrozen after they expire, but let it be used via distinct actions, e.g. "Move to jar"~~/"Renew"~~ -> See screenshots.

**Describe alternatives you've considered**
All alternatives that come to mind involve several manual steps by a user, which feels unnecessarily complicated.

**Additional context**
<img src="https://user-images.githubusercontent.com/3358649/196885718-5be11740-02f6-4496-90f0-aed006438032.png" width=400 />

dergigi:
I guess this can be closed as #556 got merged? 

If "renew" is still a priority I'd say we create a new issue.

theborakompanioni:
Resolved by #556.

PR #550, opened 2022-10-20
refactor: prepare spending fidelity bond

theborakompanioni:
This is a refactoring in preperation of #547 

Why? Because the follow-up PR would contain too much changes to be properly reviewed.

The change set may seem a little bit overwhelming, but this PR is in fact just renaming and moving stuff around.

e.g.
- more generic components  (especially regarding passing translations strings) that will be used in PR for #547
- add missing i18n translation keys
- add property`utxosByJar` to `WalletInfo` (because it is used quite frequently and should only be created once)
- renaming `account` -> `jar` wherever possible

Also, it adds a `globals.d.ts` file to contain all app-wide used TypeScript types (currently only holding `JarIndex`).

One small bugfix is included: Prevent an error while creating a Fidelity Bond from displaying a loading indicator indefinitely.
Before this fix, when creating a FB failed, a user could not restart the procedure. After this PR is applied, a user can properly restart the flow in order to try creating a FB again.

### How to test?
Try creating a fidelity bond with a coinbase transaction that has less than 100 confirmations (execute `./docker/regtest/mine-block 1 ${address_from_jar_A}`) without this fix applied (e.g. from the current master branch) and notice that the resulting error prevents users from restarting the "create FB"-flow. Then apply this fix (by checking out this branch `git checkout prepare-spend-fb`) and try again. This time the error should be displayed and a button to "Start over" is presented.


editwentyone:
utACK

editwentyone (approved):
utACK

Issue #528, opened 2022-10-03
ui: improve header of JarDetailsOverlay


theborakompanioni:
Header of `JarDetailsOverlay` component can be visually improved and should be aligned with the Figma copy.
Figma: https://www.figma.com/file/kfejZJFlwBywvLEnPEmJo1/JoinMarket-UI

## :camera_flash: Before/Current

<img src="https://user-images.githubusercontent.com/3358649/193567398-962ea0ee-9982-41bf-8d23-289900ddb8c6.png" width=600 />

## :camera_flash: After
<img src="https://user-images.githubusercontent.com/3358649/193567447-264fd3b4-a927-4df0-a552-7b97f1b67441.png" width=600 />

Jared-Dahlke:
@theborakompanioni how does this look?

light mode:
![image](https://user-images.githubusercontent.com/32936978/210836220-1f14b736-2568-47e5-83f6-842e48e42a7c.png)

dark mode:
![image](https://user-images.githubusercontent.com/32936978/210836116-e338d269-05d8-44c1-bf80-c92d8996f07a.png)

I think the tabs need a new styling for dark mode when they are on the lighter color bg, but I don't see a figma for that. Was thinking about adding a new prop to <SegmentedTabs/>  to accamodate this

theborakompanioni:
> @theborakompanioni how does this look?

Great :orange_heart: 
 
> I think the tabs need a new styling for dark mode when they are on the lighter color bg, but I don't see a figma for that. Was thinking about adding a new prop to to accamodate this

Yeah, what about adapting the background color in dark mode e.g. matching the darker color of the table? We can further discuss in the PR. Just try and take what you feel fits best for now.

PR #551, opened 2022-10-26
feat: Show jar total amount in detail view

### LINKED ISSUE #517: Jar detail view: show total amount in jar

httpiga:
Closes #517.
 I added total balance next to the tab selector so that it will be always visible.

Screenshot 2022-10-26 at 12 10 57

### DIFF
--- a/src/components/jar_details/JarDetailsOverlay.module.css
+++ b/src/components/jar_details/JarDetailsOverlay.module.css
@@ -75,26 +75,6 @@
   background-color: var(--bs-gray-800);
 }
 
-.overlayContainer .tabContainer .utxoListTitleBar .operationsContainer {
-  display: flex;
-  width: 100%;
-  justify-content: space-between;
-  flex-direction: column;
-  align-items: flex-start;
-  gap: 0.5rem;
-}
-
-.selectedUtxosSumContainer {
-  align-self: end;
-}
-
-@media only screen and (min-width: 576px) {
-  .overlayContainer .tabContainer .utxoListTitleBar .operationsContainer {
-    flex-direction: row;
-    align-items: center;
-  }
-}
-
 .overlayContainer .tabContainer .utxoListTitleBar .freezeUnfreezeButtonsContainer {
   display: flex;
   flex-direction: row;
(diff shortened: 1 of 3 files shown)


theborakompanioni:
Hey @httpiga, thanks for your contribution  - nice!
There is already a place where the balance is shown if you select individual utxos. Would it make sense to move the total balance to the same position (if no utxo is selected)?



I am asking because there is also a task related to improving the header. See https://github.com/joinmarket-webui/jam/issues/528.

editwentyone:
I like the input, edited my screen

image

theborakompanioni:
> I really like @editwentyone design, what do you think if we merge it with my consideration (assuming you agree with this) and we put total amount info below jar name?

:+1: from me!

ping @editwentyone

editwentyone:
@httpiga I would like to keep the header clean, the total amount is now under "details" and "utxos" visible

please check the Figma file, you can have a better understanding where we are heading: [link](https://www.figma.com/file/kfejZJFlwBywvLEnPEmJo1/JoinMarket-UI?node-id=4139%3A78178&viewport=-2993%2C-784%2C0.51)

editwentyone:
hm you are right, its not including the internal ones. will need to rethink it a little bit more, thanks for the hint

theborakompanioni:
> Do you think that's better pausing this PR while waiting for an updated design, or merge this _just to have it_ and refactor later?

Would you be okay with moving it to the place the current figma files suggest, i.e. having it only on the "Jar" tab for now, and then adding it to "Details" tab in a follow-up PR? 

I think it is fine either way, we can also merge it now if you prefer that!

theborakompanioni:
> Okay, next week I'm going to move the total amount to the "UTXOs" tab like the Figma suggest!

Hey @httpiga, still motivated to get this feature in? :orange_heart:

theborakompanioni:
> Hey, sorry for having disappeared but had un unexpected busy period - I'll be glad to implement this as soon as I can!

Don't worry. Glad you are still interested. Please know, that there is absolutely no urgency involved. Take your time!

theborakompanioni:
> Thank you for your patience, what about this?

Nice! :orange_heart: 

> 
> Bitcoin amount is not bold since the font weight in `BalanceComponent` is forced at 400 by `slashed-zeroes` class name, which is used also by other components. Do we want to refactor `BalanceComponent` to use Bootstrap classes?

What do you think about merging it and doing the refactoring in a follow-up PR?
Of course, only if you are motivated to do so!

Issue #252, opened 2022-05-11
More detailed error messages on Jam screen


dnlggr:
See https://github.com/joinmarket-webui/joinmarket-webui/pull/242.

theborakompanioni:
Assigned myself to this.

Currently, whenever an error is returned by the api, something like this is done:
```ts
await Api.getWalletUtxos(requestContext)
      .then((res) =>
        res.ok ? res.json() : Api.Helper.throwError(res, t('global.errors.error_reloading_wallet_failed'))
      )
[...]
```

However, the second param to `throwError` is just a fallback reason. When JM provides a reason, it sometimes is just a generic phrase like `Processing failed`. Hence, an improvement would be, if the returned reason is wrapped with a more specific error, which shold be know because of the context. Here, it might be something like `Error while reloading wallet. Reason: Processing failed`.

e.g.
```ts
await Api.getWalletUtxos(requestContext)
      .then((res) =>
        res.ok ? res.json() : Api.Helper.throwError(res, {
          context: t('global.errors.error_reloading_wallet_failed'),
          fallback: t('global.errors.reason_unknown')
        }))
      )
[...]
```

PR #556, opened 2022-11-01
feat: spend fidelity bond

theborakompanioni:
Partly addresses #547

First iteration of moving a fidelity bond after it expired.
Feedback is highly welcome, especially in regard to wording.

A follow-up PR for "renewing" the FB is in the making (also part of #547).

First attempts of this feature were based on the "Create FB" flow:
- Select target jar
- Freeze non-participating UTXOs manually
- Unfreeze the Fidelity Bond UTXO
- Display payment info and confirm
- Send FB UTXO via direct-send
- Restore state of previously frozen UTXOs

While this method gives users a better understanding of what is going on, it was annoying, resulted in so much code and just felt bloated.

Now it is just:
- Select target jar
- Display payment info and confirm

With the current approach, the necessary intermediate steps are hidden from the user. This makes for a much nicer UX. Any input on this is highly appreciated.

Also, the resulting method has been generalized: Sending any amount of specific UTXOs to a generic destination (internal or external). This can be reused in the upcoming PR when renewing the FB and also elsewhere.

## Other stuff
- disallow unfreezing of expired fidelity bonds from `JarDetailsOverlay` component
  - this is necessary as unfrozen, expired FB cannot take part in taker operations (when starting a maker operation they are auto-frozen by JM, but unfortunately not for taker operations)
  - Fidelity Bond UTXOs will always show the lock icon
- externalized `PaymentConfirmModal` to be used from other components than `Send`
- prepare and lay ground for #252 to improve error messages coming from the backend
- order fidelity bond UTXOs by value and locked state
- prevent `Balance` component from unnecessary re-renders (+handle possible `NaN` values)

**Note**: Spending an expired fidelity bond UTXO can still fail if the median time of the chain is still lower than the locktime of the bond. There is currently no way to verify if it is successfully spendable and a bond will always be marked as "expired" if its locktime is in the past.

## How to test?

Create an expired fidelity bond (only possible when starting via `npm run dev:start`) and try moving it into a jar.
Also note, that only the actual fidelity bond UTXO should be spent, all other UTXOs should be frozen and not be moved. Once the payment is done (regardless of success), UTXOs should have their original state again.


editwentyone:
Nice work so far! 

Is the source jar selectable to move the fb utxo inside there? If not, Maybe it should be grayed out? 

„Payment without privacy improvements“ I think will stick stay with „moving without privacy improvements“ because we are not paying anything in the users eye 

I like it. Where is this „moving“ icon inside the button from?

theborakompanioni:
> Nice work so far!
> 
> Is the source jar selectable to move the fb utxo inside there? If not, Maybe it should be grayed out?

Yes. You can move the expired FB anywhere.

> 
> „Payment without privacy improvements“ I think will stick stay with „moving without privacy improvements“ because we are not paying anything in the users eye

I was thinking of completely removing this phrase, as users might be confused and it is not really needed here. What do you think?

> 
> I like it. Where is this „moving“ icon inside the button from?

"Transfer" icon on https://bitcoinicons.com
I know that there is no icon in your template. Can also be removed, if you say so!

editwentyone:
All good so far. Will have a look asap from my desktop. But overall you nailed it. Maybe we should rephrase it to something like: make fb usable again… bla bla

so the user gets the context that before that it’s not possible to access the funds

theborakompanioni:
> All good so far. Will have a look asap from my desktop. But overall you nailed it. Maybe we should rephrase it to something like: make fb usable again… bla bla
> 
> so the user gets the context that before that it’s not possible to access the funds

That's true.. maybe "Move to Jar" is misleading as it already is in Jar A, it is just not able to take part in a taker/maker operation.
Any input is highly appreciated. Maybe just something plain and simple "Spend"?

editwentyone (approved):

dergigi:
Oh wow, that's big. Will do my best to review this fully until EOW.

dergigi:
Conflicts introduced due to merge of #566 - please resolve @theborakompanioni 🙏

theborakompanioni:
> Conflicts introduced due to merge of #566 - please resolve @theborakompanioni pray

Resolved. :pray:

dergigi:
What about: "Unlock expired fidelity bond?"

That's what you want to do after all, no? In some sense the FB is locked away, both in time and space. Once the timelock expires you can unlock it completely, by moving it to any jar.

Thoughts on this terminology @theborakompanioni @editwentyone?

dergigi:
Another thought: Why give the user the option to select a jar in the first place? 

I'd default to **Jar A**, or at least pre-select A and hide the selection below a fold. Why make the user think? It's not terribly important, is it? The user just wants to press "Unlock" and wants to use the funds again.

Here is how I would simplify the flow & the language:

Screenshot 2022-12-09 at 13-25-42 Jam for JoinMarket

Screenshot 2022-12-09 at 13-26-27 Jam for JoinMarket

Two clicks, the second one being a confirmation.

theborakompanioni:
> Another thought: Why give the user the option to select a jar in the first place?
> 
> I'd default to **Jar A**, or at least pre-select A and hide the selection below a fold. Why make the user think? It's not terribly important, is it? The user just wants to press "Unlock" and wants to use the funds again.

Done.

> 
> Two clicks, the second one being a confirmation.

Not too convinced, as this is an action not performed very often and the described approach provides less freedom users.
It is now changed to skip the selected automatically and default to Jar A - the code to selected a Jar is still present for now and can be enabled with a flag - hope that is okay (in case it is decided to enable this flow or provide a toggle)

> (I'd also remove the red warning that tells the user that this is a send without privacy improvement. The user can't change it at this point, there is no need to show a warning. It's just scary and confusing.)

Done.

dergigi:
Nice! I like it. Much more streamlined.

> It is now changed to skip the jar selection and default to Jar A automatically - the code to select a Jar is still present for now and can be enabled with a flag - hope that is okay (in case it is decided to enable this flow or provide a toggle).

If we want to keep it in, we could add an `Unlock Options` to the Unlock modal, just like we have on Receive?

Add this "fold" ...

Screenshot 2022-12-13 at 11-53-26 Jam for JoinMarket

...to this screen:

Screenshot 2022-12-13 at 11-54-49 Jam for JoinMarket

---

Would still default to A (unless you go through the trouble of changing it) and would hide the complexity nicely, while still providing the option.

What do you think?

theborakompanioni:
> What do you think?

Full ACK.. Much better, imho. : -)
I'll merge this as it's an improvement to the current state and it feels "good enough" for now.
It is more important that users are able to spend their expired fidelity bonds. Currently, only acting as maker would freeze expired bonds automatically (in case it was unfrozen), but not when acting as a taker. This can potentially lead to confusion and/or needing additional support.

An issue for this has been created: https://github.com/joinmarket-webui/jam/issues/574

PR #564, opened 2022-11-05
refactor: prepare removing `/display` request on every page load

theborakompanioni:
TBD.


dergigi (approved):
Looks good to me, tACK ✅

Good job cleaning up some of the naming as well, it's much clearer now! ✨ 👏

Issue #356, opened 2022-06-28
Question: How to prevent loading `/display` and `/utxos` on most pages


theborakompanioni:
This issue is related to _performance_, so not a high priority as of yet.

Some (most) screens require the actual balance and utxo states to work properly.
Currently, this information is always loaded before the components enables it's functionality.
However, this information does not change very frequently. Hence, most of these polling requests are rather unnecessary and could be omitted.

Is there a way these requests can be triggered only when necessary?
Some thoughts come to mind:
- Using the websocket data to determine if a reload is necessary
- Fallback to auto-reload on component initialization if websocket is not available
- Ability to manually reload needed data (?)
- Should some data be cached, and if so, what is a suitable approach

Any insight on this is highly appreciated. Obviously, this data should be stored only as long as a session is active, optionally encrypted if that makes any sense.

PR #566, opened 2022-11-07
refactor: speed up initial page load

theborakompanioni:
Resolves https://github.com/joinmarket-webui/jam/issues/356.
Based on https://github.com/joinmarket-webui/jam/pull/564.

**Note**: This is based on #564 - you can either review them separately or just review this one. Be aware that the change set is rather large. Unfortunately, this could not be avoided due to the deep impact of the new behaviour on nearly all components.

Before this commit, a request to API endpoint `/display` was always triggered on almost every page load.
After this commit, the request to `/display` is made when necessary.

Users have reported that on `mainnet`, this request alone can take up to ~20 seconds!
The speed increase is also very noticeable on `regtest` in your dev environment.

### Behaviour before
Requests to `/display` were done when:
- After initial loading of wallet
- navigating to `MainWalletView`
- navigating to `Send`
- navigating to `Earn`
- navigating to `Jam`
- freezing/unfreezing of UTXOs

### Behaviour now
Requests to `/display` are done when:
- After initial loading of wallet
- opening JarDetails Overlay
- attempting to create a Fidelity Bond

---

However, there are trade-offs! Now, one must know which data is needed on the current screen as opposed to just reloading all data. i.e. if `addressSummary` is needed, data from `/display` still has the be loaded.

Also, it can be that the user is not warned when an address is reused. This is the case after sending to an own address. Before, if a second attempt to send to the same address is made, a user has seen a warning. Now, this warning is not displayed anymore. It can be mitigated by also scanning the UTXO addresses (not done in this PR!).

### How to test
You can compare new and past behaviour by clicking through the UI on `localhost:3000` (new behaviour) and on `localhost:29080` (past v0.1.2 behaviour). If you open these in two windows next to each other, the speed increase becomes apparent.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,18 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+### [0.1.3](https://github.com/joinmarket-webui/jam/compare/v0.1.2...v0.1.3) (2022-11-10)
+
+#### Fixed
+
+* **docker**: wait for bitcoind to accept RPC calls ([#559](https://github.com/joinmarket-webui/jam/pull/559)) ([6e2ee47](https://github.com/joinmarket-webui/jam/commit/6e2ee47538fe225f7b84eb2de245993a90dfd042))
+* **pagination:** colors of option element in dark mode ([#554](https://github.com/joinmarket-webui/jam/issues/554)) ([86dc2c5](https://github.com/joinmarket-webui/jam/commit/86dc2c5d545037e2a1a593049ac6b96e31910d07))
+
+#### Added
+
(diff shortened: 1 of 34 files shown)


dergigi:
@theborakompanioni I assumed that #564 could be merged and that there would be no conflicts with this PR. Turns out the assumption was wrong. To avoid this in the future, we could change the base branch of cascading PRs? 

In any case, I will review this again after conflicts are resolved.

theborakompanioni:
> @theborakompanioni I assumed that #564 could be merged and that there would be no conflicts with this PR. Turns out the assumption was wrong. To avoid this in the future, we could change the base branch of cascading PRs?

I thought so too. My fault. Yes, will change base branches of cascading PRs from now on. Sorry for the delay.

> In any case, I will review this again after conflicts are resolved.

Conflicts are resolved.

dergigi (approved):
Finally managed to review and test this PR, apologies for the delay.

Looks good to me, and I didn't run into any issues while testing. Performance improvement is substantial! Well done 🙌 

tACK ✅

PR #567, opened 2022-11-07
build(deps): update dependencies

theborakompanioni:
Removes unused dependency `serve`, `@emotion/react` and `@popperjs/core` and updates the following ones:    
     
     bootstrap                         ^5.2.1  →  ^5.2.2
     i18next-browser-languagedetector  ^7.0.0  →  ^7.0.1
     react-bootstrap                   ^2.5.0  →  ^2.6.0
     react-router-dom                  ^6.4.2  →  ^6.4.3
     typescript                        ^4.7.2  →  ^4.8.4

Also, separate between `dependencies` and `devDependencies`. This is normally not needed for a non-library project (and `create-react-app` defaults to put everything in `devDependencies`), but it is nicer for devs to know which dependencies are needed during runtime and which are for testing/developing only.


editwentyone (approved):
utACK

PR #572, opened 2022-11-30
fix: no. of collaborators as integer

theborakompanioni:
Before this PR, Jam displayed and sent a fractional number as collaborators to the backend (which rounded down to the nearest int). After this commit is applied the number of collaborators as always parsed and displayed as integer.

Before:
image

After:
image


dergigi (approved):
Looks good to me, tACK ✅

Small NIT: imho it would be better if the input field wouldn't allow nonsense in the first place, but that's just me. This PR is definitely an improvement, so I'll merge it as is.

PR #576, opened 2022-12-14
chore(ui): add padding to seed backup success message

theborakompanioni:
Small cosmetic change. Adds padding to the success message after seed phrase confirmation.


dergigi (approved):
Looks good to me, utACK ✅

PR #584, opened 2023-01-03
fix: correctly render existing bonds in Safari

theborakompanioni:
It seems Safari is not able to construct `Date` objects from strings formatted as `yyyy-MM-dd 00:00:00` (see [stackoverflow#4310953](https://stackoverflow.com/questions/4310953/invalid-date-in-safari)).
Before this commit, as soon as a user on Safari created a fidelity bond, an error was thrown and the page became Earn page did not render.
After this commit, the fidelity bond is shown correctly and the Earn page renders successfully.

This is an _untested_ change, as every attempt to run Safari failed on my workstation. It be great if someone can verify that it works as expected. However, the root problem should be addressed, as all date objects are instantiated with a timestamp (type `number`) which every browser supports.

### DIFF
--- a/src/components/fb/CreateFidelityBond.jsx
+++ b/src/components/fb/CreateFidelityBond.jsx
@@ -555,7 +555,7 @@ const CreateFidelityBond = ({ otherFidelityBondExists, wallet, walletInfo, onDon
           }}
         >
           {t('earn.fidelity_bond.confirm_modal.body', {
-            date: new Date(lockDate).toUTCString(),
+            date: new Date(fb.lockdate.toTimestamp(lockDate)).toUTCString(),
             humanReadableDuration: fb.time.humanReadableDuration({
               to: fb.lockdate.toTimestamp(lockDate),
               locale: i18n.resolvedLanguage || i18n.language,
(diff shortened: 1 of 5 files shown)


dergigi:
I was able to create a FB without issues on current `master` commit 9d1575f running Safari Version 16.2 (18614.3.7.1.5).

Thus running the changeset of this PR has no effect for me, it seems?

theborakompanioni:
> I was able to create a FB without issues on current `master` commit [9d1575f](https://github.com/joinmarket-webui/jam/commit/9d1575ffec4764f834fa28b9faefee97ae88ee5b) running Safari Version 16.2 (18614.3.7.1.5).

Just fyi, this has been reported by two distinct users on Safari. After creating a FB, Jam "crashed" (showing the error page). Switching to a different browser has been a successful workaround. Unfortunately, I do not know which version of Safari they used. 

Additionally, it is still safer to use the unix timestamp appended to the `path` of the UTXO, than to pass the raw `locktime` property to the `Date` constructor. e.g.
```json
{
  [...]
  "path": "m/84'/1'/0'/2/40:1682899200",
  "locktime": "2023-05-01 00:00:00",
}
```

That's why it might be good to be merged.
What do you think?

PR #586, opened 2023-01-03
feat: add dedicated error page

theborakompanioni:
Before this commit, in case there was an error, a generic error page had been shown.
After this commit, a dedicated error page, including the navbar, the footer and with a link to create an issue on GitHub, will be displayed.

Whenever you browse the app, this page should never render in the first place, as all error are to be handled by the actual component. However, some errors slip through (e.g. with the date parsing bug in safari #584) and then, the user should be presented with a better view.

The page is really just the bare minimum. The error is unexpected and in most cases, cannot be resolved by the user without support.

In order for this to work properly, the dependency `react-router-dom` has been upgraded from `v6.4.3` to `v6.6.1`.

## How to test

In dev mode (when starting with `npm run dev:start`), a special route with name `error-example` is present.
Visit http://localhost:3000/error-example to view this page.

## :camera_flash: Before
image

## :camera_flash: After
image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -23,7 +23,7 @@
         "react-dom": "^17.0.2",
         "react-i18next": "^12.0.0",
         "react-router-bootstrap": "^0.26.2",
-        "react-router-dom": "^6.4.3"
+        "react-router-dom": "^6.6.1"
       },
       "devDependencies": {
         "@testing-library/jest-dom": "^5.16.5",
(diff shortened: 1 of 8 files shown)


dergigi (approved):
Looks good to me, tACK ✅ 

Suggesting some small changes.

dergigi (approved):

dergigi:
Looks like this now: 

Screenshot 2023-01-24 at 00-04-30 Jam for JoinMarket

PR #591, opened 2023-01-09
feat: quick freeze/unfreeze utxos

### LINKED ISSUE #480: Quick freeze/unfreeze in UTXO list
- [x] Red/blue color depending on state
- [x] Quick freeze/unfreeze on hover
- [x] Various UI touch-ups

---

See [this comment](https://github.com/joinmarket-webui/jam/pull/430#issuecomment-1213045038) and [Call 22](https://github.com/joinmarket-webui/jam/wiki/Community-Call-%2322).

https://www.figma.com/file/kfejZJFlwBywvLEnPEmJo1/JoinMarket-UI?node-id=5071%3A87607

image

https://user-images.githubusercontent.com/96262473/184351051-60867ad0-af57-4861-a041-68231bff357f.mov

### LINKED ISSUE #525: ui: shorten addresses in JarDetailsOverlay
could you shorten the address? "bc1x xxxx ... xxxx" there are examples in Figma. what do you think?

_Originally posted by @editwentyone in https://github.com/joinmarket-webui/jam/issues/495#issuecomment-1264432605_

theborakompanioni:
Closes #480.
Closes #525.

This PR introduces a way to quickly freeze or unfreeze single UTXOs.

Some decision have been made that differ from the Figma design:
- using slightly different colors for red and blue, that look good on both, dark and light mode
- remove "frozen" tag, as snowflake icon indicating "frozen" state is already present
- "bond" tag renamed to "locked"
- ordering of columns: quick freeze/unfreeze placed at the front
- confirmations column is kept, as it is important info that should be quickly available
- instead of a "temperature indicator" (Font Awesome, CC BY 4.0 License) a "sun" icon (Bitcoin Icons, MIT) is used, in order to avoid license issues


theborakompanioni:
Q: In order to save some space on mobile devices, should the "quick freeze/unfreeze" button move from it's own column to the position of the "utxo icon" when hovering the row?

ping @editwentyone

editwentyone:
@theborakompanioni because there is no hover mode on mobile, we don't need the column with the action icon. we just need the status column with the snowflake. if a user wants to perform an action, he needs to select the row and unveil the action buttons.

dergigi (approved):
Nice! I like it.

It feels a little unintuitive to me, as I would expect the "freeze" action icon to be the same as the freeze indicator icon, as it is the case with "like" or "favorite" and so on. (Compare with "star" in Gmail's list view, or "read/unread" in ProtonMail's list view, and so on.)

Having a 2nd icon for the action is a bit weird, and having a one column for the state (frozen, not frozen) and one column for the action (freeze, unfreeze) is also unusual. Apologies on my side for bringing this up now, and not earlier in Figma.

That being said, it's a small feature for power-users, and it's good as it is. Tested on desktop and mobile. Let's see what the users say, if anything.

Looks good to me, tACK ✅

PR #593, opened 2023-01-11
ui(Jars): move jar info next to icon on small screens

theborakompanioni:
Small PR to align the main wallet view with Figma screens.

Places the jar info (name and balance) next to the jar icon on small screens.


editwentyone (approved):
ACK - THNX

Issue #508, opened 2022-09-21
Send: Hide "number of collaborators" behind send options


dergigi:
As of now, we have "receive options" for the receive screen, but no "send options" for the send screen.

Two things can be set:
* Collaborative transaction (default = true)
* Number of collaborators (default = 9)

I guess these defaults are fine for most people, and could be hidden behind collapsible options, just like we have it on the receive screen.

editwentyone:
[🎨 Figma](https://www.figma.com/file/kfejZJFlwBywvLEnPEmJo1/JoinMarket-UI?node-id=4433%3A81973&viewport=-1567%2C-3208%2C0.59)

<img width="502" alt="image" src="https://user-images.githubusercontent.com/96262473/197946488-0e9b4460-57fa-4f13-8a02-15b4eb229543.png">

theborakompanioni:
Closed by #595

PR #595, opened 2023-01-14
Send: Hide "number of collaborators" behind send options

httpiga:
This solves [#508](https://github.com/joinmarket-webui/jam/issues/508).

I decided to not over-engineering the new `Accordion` component but we may take into consideration possible future side effects of unmounting children components (which will likely be form fields) when user closes the accordion: for example, if we will use it with some Formik registered components, we may (or may not) get some unwanted behaviour about validation or field values being correctly send in the form submit (take my consideration with a grain of salt, I haven't verified it).
Any feedback is appreciated 🧡 

#### Closed
image

#### Opened
image

[Development soundtrack](https://open.spotify.com/album/5MYbCAee86KlqqCNbouhhG?si=i5Q65Bn_SfG4tjCGOvYckw) 🎧

### DIFF
--- a/src/components/Accordion.tsx
+++ b/src/components/Accordion.tsx
@@ -28,8 +28,8 @@ const Accordion = ({ children, defaultOpen, title }: AccordionProps) => {
         {title}
         <Sprite symbol={`caret-${isOpen ? 'up' : 'down'}`} className="ms-1" width="20" height="20" />
       </rb.Button>
+      <hr className="m-0 text-secondary" />
       {isOpen && <div className="py-4">{children}</div>}
-      <hr className="m-0" />
     </div>
   )
 }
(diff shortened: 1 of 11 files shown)


editwentyone:
thanks, could you please tune it more to this screen? would be awesome 

https://www.figma.com/file/kfejZJFlwBywvLEnPEmJo1/JoinMarket-UI?node-id=4433%3A81973&t=jL3pT3ELgD9n83IM-11

the horizontal line should stick to the sending options for example

theborakompanioni:
Added some minor adaptions:
- remove "form" in "form" tag to prevent `validateDOMNesting(...): <form> cannot appear as a descendant of <form>` warning (`CollaboratorsSelector` component)
- hide (instead of removing) children node in `Accordion` component
- make `title` property mandatory in `Accordion` component
- make use of `Accordion` on Earn page, also fixing validation messages

theborakompanioni (approved):

PR #599, opened 2023-01-23
feat(Jam): add success message

### LINKED ISSUE #489: Scheduler: Success Screen

theborakompanioni:
Resolves #489.

This PR adds a success message after a scheduler run completes.

Before this PR, the page reloaded greeting the user with a message ala "The scheduler requires your UTXOs to have 5 or more confirmations", which is confusing to say the least. 
After this PR, the user will see a screen indicating success with a short summary, i.e. "Successfully completed n scheduled transactions".

Please note that due to polling the scheduler data, this might miss the state change of the last scheduled item. To mitigate this situation, it is checked whether only frozen UTXOs remain in the wallet, which should always be the case on mainnet. However, in "dev mode" (where – on purpose – not all jars a swept in order to decrease the time the scheduler needs to run; an on top – the funds are sent to Jar A of the same wallet) this might still occur. 
https://github.com/joinmarket-webui/jam/blob/8a0be65a185180f0561699bbed2f1fb0a4d0ae2f/src/components/Jam.tsx#L249-L260
If you have an idea on how to upgrade this from a workaround to a fully reliable fix, please say so!

As always, any feedback is greatly appreciated. As well as any improvement of  wording/phrases/etc.! 

## How to test
Start the scheduler while mining some blocks. Wait for the scheduler run to be completed.
Since polling is done every 10 seconds, and mining in the regtest environment is also done every 10 seconds, the situation described above might apply and you won't see the success message. Either increase the time between regtest blocks by running `watch -n 60 ./docker/regtest/mine-block.sh` (blocks every 60 secs) instead of `npm run regtest:mine` (blocks every 10 secs), OR send the funds to an external test wallet (e.g. to one of the regtest makers, `http://localhost:29080`). As mentioned, this should only be an issue on `regtest`, not in production.


dergigi:
Awesome! That's very much needed. 

> Since polling is done every 10 seconds, and mining in the regtest environment is also done every 10 seconds,

What about using different intervals as defaults, possibly primes, so that the chance of this race condition to occur is minimized? `11` and `13` would work, but any two primes should work.

dergigi:
On another note: it's difficult to review the changeset in `Jam.*` because of the switch from `.jsx` to `.tsx` - now there's no diff to look at, as it's a complete rewrite.

Not a huge issue, but doing this kind of refactoring in a separate PR would be better.

dergigi:
Any thoughts on the "prime number" hack as a fix for the issue you raised, @theborakompanioni?

> What about using different intervals as defaults, possibly primes, so that the chance of this race condition to occur is minimized? `11` and `13` would work, but any two primes should work.

src/components/Jam.tsx
```diff
@@ -0,0 +1,548 @@
...
+      body.tumbler_options = {
+        addrcount: addressCount,
+        minmakercount: 1,
+        makercountrange: [1, 0],
+        mixdepthcount: addressCount,
+        mintxcount: 1,
+        txcountparams: [1, 0],
+        timelambda: 0.025, // 0.025 minutes := 1.5 seconds
+        stage1_timelambda_increase: 1.0,
+        liquiditywait: 10,
```
dergigi:
```suggestion
        liquiditywait: 13,
```

dergigi (approved):
I've changed the values to primes (`11` and `13`) so the chance of the race-condition you described happening will be reduced. I've tested it a bunch of times and it always worked for me on `regtest` (and I don't think these changes will break anything - famous last words, ha!). I'll merge this now and will prepare a release after that. Feel free to revert my changes if you think they could be problematic. 

Great job on this, it's way prettier and intuitive now. 💖 

tACK, LGTM ✅

PR #600, opened 2023-01-25
ui(Send): replace dropdown input with JarSelector

### LINKED ISSUE #598: Display spendable balance on send page
From the chat:

> When a jar has less sats than you are trying to spend it let's you attempt the spend and then quietly fails.  It took me a few tries to figure out what was going on.  Normally this would be obvious, but in the case of jar with the FB it shows the amount of sats including the FB.  Leading me to think they were all available to spend.  

> I think a warning to let you know that you don't have sufficient funds(unlocked/unfrozen) for the proposed spend would fix this.

This has been talked about before, but there is no ticket yet.
It is confusing to show the total balance of a jar, when in fact some UTXOs are frozen or time-locked.
How can this best be displayed and communicated to the user?

theborakompanioni:
Closes #598.

This  replaces the "Source Jar" dropdown with the `SelectableJars` component.
Additionally, the spendable balance per Jar on shown instead of the total amount.


editwentyone (approved):
ACK 🚀

dergigi:
Nice! Happy to test this once conflicts are resolved 🙏

dergigi (approved):
Looks good to me, tACK ✅

PR #601, opened 2023-01-25
fix(fb): correctly display success screen after unlocking fb

theborakompanioni:
There has been a bug when unlocking an expired fidelity bond: After the tx had been broadcasted, the success screen disappeared, as the modal was hidden as soon as no FB was present anymore.

With this fix, the success screen is shown correctly after an expired FB has been unlocked.

### How to test?
On master, verify you can reproduce the bug: After you spend an expired Fidelity Bond, no success screen is displayed.
Then checkout this branch and repeat the procedure. This time the success screen should be displayed.


dergigi (approved):
Looks good to me, tACK ✅

PR #585, opened 2023-01-03
fix: install python3-venv in regtest environment

theborakompanioni:
Fixes https://github.com/joinmarket-webui/jam/issues/581

Quick workaround to make the regtest environment compatible with the current state of master ([`ea7a6e`](https://github.com/JoinMarket-Org/joinmarket-clientserver/commit/ea7a6e1a5b9dde5eb0af2b22d831e843f4800020)) again. This can be removed [once the install script is fixed upstream](https://github.com/JoinMarket-Org/joinmarket-clientserver/pull/1396#issuecomment-1369975749).

Also, adapts the payjoin server port from `8080` to `8082` according to [`2749da1`](https://github.com/JoinMarket-Org/joinmarket-clientserver/commit/2749da174ac0cf7c5186b88af15b2970df5dce49) - included in JM `v0.9.9`.

PR #604, opened 2023-01-31
fix(fees): allow min tx fee of 1sat/vbyte

theborakompanioni:
Requires JM `v0.9.9` or greater -> needs #585 to be merged and a recreated dev environment by running `npm run regtest:rebuild`!

Before this commit, minimum in Jam was 3 sats/vbyte as prior ot JM `v0.9.9`, this resulted in an error, as JM used to little fees and the tx was rejected from the mempool.
When this commit is applied, setting the tx fee to 1 sat/vbyte is allowed.

## How to test
Set your tx fees to 1 sat/vbyte (with 0% randomization) and try to spend an expired fidelity bond.


dergigi (approved):
tACK ✅

PR #532, opened 2022-10-04
feat: display fee settings on Send page

theborakompanioni:
Partly addresses #518.
Display values based on fee settings in the confirmation modal of the Send page.

"Mining fee target" will be displayed either as "targeted blocks" or "sats/vByte" (randomized or exact):
e.g.
- "Next block" (`tx_fees := 1`)
- "In 6 blocks" (`tx_fees := 6`)
- "In 1,000 blocks" (`tx_fees := 1000`)
- "1.001 sats/vByte" (`tx_fees := 1001`, `tx_fees_factor := 0`)
- "9.1 - 10.9 sats/vByte" (`tx_fees := 10000`, `tx_fees_factor := 0.9`)

"Estimated Max. Collaborator Fee" will be displayed for collaborative transactions.
Calculation can be found here: https://github.com/joinmarket-webui/jam/blob/feat/display-fee-settings-in-send/src/hooks/Fees.ts#L69-L70 (Tests: https://github.com/joinmarket-webui/jam/blob/feat/display-fee-settings-in-send/src/hooks/Fees.test.ts#L22-L69)

PR #606, opened 2023-02-04
feat(Send): Fee breakdown table

httpiga:
Resolves https://github.com/joinmarket-webui/jam/issues/632.

Hello people! 🧡 
I'm trying to implement the fee breakdown table in the send page, as designed in the Figma project. 
UI seems ready (see screenshot below), but befor implementing the logic I need to better understand how fees work in Jam 😄 
- Where can I find the "Absolute fee limit" and "Relative fee limit" values?
- What do you mean by "Absolute/Relative fee limit, **chosen randomly**"? The **limit** should be fixed, do you mean that the fee is randomly selected below the absolute/relative limit?
- What value do we want to show in the personal absolute/relative fee limit, if the user didn't set any? 
- How to calculate the estimated total fees? The Figma design shows a single value, what about calculating the minimum and maximum possible fees for the transaction, and displaying the range? 

image

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -29,7 +29,7 @@ jobs:
         run: npm ci
       # Checks
       - name: Lint
-        run: npm run lint -- --version && npm run lint
+        run: npm run lint
       # Test
       - name: Test
         run: npm test


dergigi:
> What do you mean by "Absolute/Relative fee limit, chosen randomly"? The limit should be fixed, do you mean that the fee is randomly selected below the absolute/relative limit?

Yes, this is poorly worded. The limit is fixed and the fee is chosen randomly below the limit.

theborakompanioni:
> Hello people! orange_heart I'm trying to implement the fee breakdown table in the send page, as designed in the Figma project. UI seems ready (see screenshot below), but befor implementing the logic I need to better understand how fees work in Jam smile

Hey @httpiga, sorry for the late reply.

> 
>     * Where can I find the "Absolute fee limit" and "Relative fee limit" values?

These can be loaded with function `useLoadFeeConfigValues` from [`hooks/Fees.ts` ](https://github.com/joinmarket-webui/jam/blob/master/src/hooks/Fees.ts). See usage in `Send`, `FeeConfigModal`, etc.

>     * What do you mean by "Absolute/Relative fee limit, **chosen randomly**"? The **limit** should be fixed, do you mean that the fee is randomly selected below the absolute/relative limit?

Already addressed by @dergigi above. Agree, wording is not clear and should be improved.

>     * What value do we want to show in the personal absolute/relative fee limit, if the user didn't set any?

These values _must_ be set in order for collaborative transactions to work. JM[ does not provide default values and will use `x`](https://github.com/JoinMarket-Org/joinmarket-clientserver/blob/v0.9.9/jmclient/jmclient/configure.py#L314-L324) in the standard config. The `standalone` image of Jam [will populate these settings with random values](https://github.com/joinmarket-webui/jam-docker/blob/v0.1.5-clientserver-v0.9.9/standalone/jam-entrypoint.sh#L54-L58).

Currently, Jam will only prompt a user to set them manually, after an attempt to invoke a collaborative transaction fails ("Config variables 'max_cj_fee_rel' and 'max_cj_fee_abs' must be set in your joinmarket.cfg in order to send collaborative transactions. Consider adding them to your config manually."). This can also be improved, as the fee settings are now loaded immediately and the fee values should already be present.


>     * How to calculate the estimated total fees? The Figma design shows a single value, what about calculating the minimum and maximum possible fees for the transaction, and displaying the range?
> 

e.g. [See the code in `PaymentConfirmModal`](https://github.com/joinmarket-webui/jam/blob/v0.1.5/src/components/PaymentConfirmModal.tsx) for the current estimate, which is shown before a user attempts a transaction. Code can probably be externalized to be used in multiple components. Note, that it only estimates the "maker fees" (paid to other JM makers participating in the transaction), not the miner fee for the transaction, as this is not possible for two reasons: 
- the amount of inputs and outputs is not know beforehand, so it is not possible to know the transaction size, hence the transaction fees cannot be estimated
- it is also impossible if the user has a "block target" as mining fee target (read: "Include my TX in the next `n` blocks", not a sats/vbyte value), as Jam cannot know information about the current mempool conditions. See https://github.com/joinmarket-webui/jam/pull/532#issue-1396083901 for more info on these values.


Hope I could give a better insight. Tricky topic. Please do not hesitate to ask further questions if something is not clear.

src/i18n/locales/en/translation.json
```diff
@@ -259,6 +261,16 @@
...
+    "fee_breakdown": {
+      "absolute_limit": "Absolute fee limit (<balance/>) * {{ num }} collaborators",
+      "or_relative_limit": "OR relative fee limit ({{ percentage }}) * {{ num }} collaborators",
+      "not_set": "Not set",
+      "too_low": "Too low",
+      "total_estimate": "Max collaborators fee",
+      "plus_mining_fee": "Plus mining fees",
+      "why_cant_estimate_mining_fee": "Why can't we estimate mining fees?",
+      "cant_estimate_mining_fee_info": "The amount of inputs and outputs is not know beforehand, so it is not possible to know the transaction size, hence the transaction fees cannot be estimated.<br>In addition to this, if you have set a \"Block target\" as mining fee (read: \"Include my tx in the next n blocks\", instead of a sats/vByte value), we cannot estimate the fee needed for your transaction since Jam does not have access to the current mempool conditions."
```
theborakompanioni, line 272:
```suggestion
      "cant_estimate_mining_fee_info": "The amount of inputs and outputs is not know beforehand, so it is not possible to know the transaction size, hence the transaction fees cannot be estimated.<br>In addition to this, if you have set a \"Block target\" as mining fee (read: \"Include my tx in the next n blocks\", instead of a sats/vByte value), the fee needed for your transaction cannot be calculated since Jam does not have access to the current mempool conditions."
```

:+1: for the text and explanation in general. Nice!
Small nit: I'd suggest to not use "we", but rather passive voice. Okay for you?

MaxHillebrand (dismissed):
that hover tooltip is in light mode, and I'd prefer to avoid paying for the user's medical bill from the resulting bleeding eyes...

Dark mode should be consistently dark, maybe a different shade of black for the tooltip box?

editwentyone:
i like the idea to tell the user based on their amaount what fee option to use and calculacte / predict it as precise as possible. 

any ideas how that could look like? or is this a task for me ;)

editwentyone:
@httpiga so, after finally looking and thinking about the fee table, we can make it way more easy but precise to communicate what costs are coming up, I will have a call with tbk and propose something asap. 

right now neither my initial nor yours makes a lot of sense (sorry to realize that so late) without modifying it further. hope its ok for you to wait a bit and then tackle it again.

editwentyone:
hey, 

so I came up with something different, hope it is clearer and if not, we really need to consider to remove it all together. looking forward to your feedback. also I did more than just the fee breakdown, you will see in the video. 

- new fee breakdown, preview
- new send button behavior to get rid of the confirmation dialog
- new summary page layout while joining


https://github.com/joinmarket-webui/jam/assets/96262473/28a6e962-581d-4fe9-9cdc-cc483c8e3019

theborakompanioni:
> hey,
> 
> so I came up with something different, hope it is clearer and if not, we really need to consider to remove it all together. looking forward to your feedback. also I did more than just the fee breakdown, you will see in the video.
> 
>     * new fee breakdown as a preview for the user
> 
>     * new send button behavior to get rid of the confirmation dialog
> 
>     * new summary page layout while joining

I like everything except the button behaviour.. I am not the one to actually suggest UI stuff, but it seems a little bit more complex (especially on mobile - where there is no "hovering"). The current approach seems simpler.. :thinking: 

What do you think @httpiga ?

(However, I like the new fee summary better than the current "table"!)

editwentyone:
> I really like this last version! It is definitely more clear than the previous table, it only shows the two relevant values to the user and allows them to immediately take action to update the fee limits. The only detail that confuses me is that by highlighting the higher fee amount (in your video, the absolute fee limit) it seems that its box is somehow "selected" while they aren't selectable. In my opinion, the two boxes shouldn't have any differences in terms of "visual weight".
> 

hm, we can of course remove the "visual weight". it can confuse, it can also function as a highlighter. something we need to test: lets do it **with** and learn going forward 

> I find it funny the button behaviour, even if I agree can bring some problems in mobile, not only about the hovering but also because in mobile browsers a long tap makes you select the underlying text/component. We can keep the current confirmation modal, as suggested by tbk, or we can replace the "Send" button with a

lets forget my funny button solution for the time being, lets focus here just on the the fee table.  

> Btw, congratulations on this nice solution!

thanks, hope it helps with users confusion

theborakompanioni:
> Hi guys, a little update on this: my last commit addresses @editwentyone suggestion. The only thing that's missing is opening the fee modal when the user clicks on the links: now links bring you to the settings page. Until I fix it, you can take a look and, if you want, share some feedback: 

Nice, just tried this and I think it is way better than the previous table-like info element. Well done, @httpiga. :pray: :rocket: :rock: :rock: :rock: 

Some feedback/suggestions/notes:
- We should make it more clear, that these are the collaborator fees only, e.g. title "Fee Estimation" to "Max Collaboration Fee (for this payment)"/"Maximum Collaborator Fee"
- Bolder font style for "This does not include regular mining fees."?
- It does not work correctly for "sweeps" yet
- Fees always denominated in sats -> nice, I think that is fine
- As you already mentioned, opening the Fee Modal with a single click would be great

Generally one thing is, that it "looks like" the user can interact with this element. Maybe we can make it more clear:
- Gray out the non-effective fee (e.g. apply css style `text-muted` if `FeeCard` is not `highlight`ed)

I can apply some suggestions myself, if you want.  May I push some of these changes?

theborakompanioni:
Note: It does not play too nicley with missing `max_cj_fee_abs` or `max_cj_fee_rel`. But let's fix that in a follow-up PR.
There should probably be a special warning (with the sending form disabled) if one of them is missing.

And since the fee values are now loaded in the component, we finally can refactor/remove the [`enhanceTakerErrorMessageIfNecessary` function](https://github.com/joinmarket-webui/jam/blob/c64a39371f8bef7559f1b10c98bad37a01f11626/src/components/Send/helpers.ts#L45). :tada:

theborakompanioni (approved):
tACK. Handling missing fee config values needs to be done in a follow-up PR, other than that, I feel this is ready to be merged :rocket:

theborakompanioni:
Great work @httpiga . Thanks for your endurance and for sticking to it, even when requirements changed during development :orange_heart:

Issue #608, opened 2023-02-06
No connection to backend: Gateway Timeout - ECONNREFUSED


SpartanBit:
Hi all, I have followed the JAM installation guide for Raspibolt described here:
https://raspibolt.org/guide/bonus/bitcoin/Jam.html

however the webui on the local network shows the following message: No connection to backend: Gateway Timeout.
This seems a very similar issue, but on Raspiblitz: https://github.com/joinmarket-webui/jam/issues/88

If I have a look at the logs, every time I try to reach JAM I get the following:

Error occurred while proxying request 192.168.1.18:4020 to undefined [ECONNREFUSED] (https://nodejs.org/api/errors.html#errors_common_system_errors)
Error occurred while proxying request 192.168.1.18:4020/api/v1/session to https://localhost:28183/ [ECONNREFUSED] (https://nodejs.org/api/errors.html#errors_common_system_errors)
Error occurred while proxying request 192.168.1.18:4020/api/v1/wallet/all to https://localhost:28183/ [ECONNREFUSED] (https://nodejs.org/api/errors.html#errors_common_system_errors)

Not sure what's causing it, but i tried to get to the root cause of this, and the only thing I've found is that some of the other guides do specify in the .env file not only the port but also the host/server at 127.0.0.1:...
see https://raspibolt.org/guide/bonus/lightning/thunderhub.html#configuration or https://raspibolt.org/guide/bitcoin/blockchain-explorer.html#configuration

Could that be the cause behind this problem, and if so, how can i resolve it?

I would be truly grateful if someone could help.
Thanks!

theborakompanioni:
Hey @SpartanBit, sorry for the late reply. Still figuring out what might be the problem.
It seems the JM API is refusing connections, so it might be that the daemon is not running in the first place?

It can be that JM creates a config file at first startup and subsequently exits. Have you tried restarting `jmwalletd`, e.g. with `sudo systemctl restart jmwalletd`?

`jmwalletd` listens on port `28183` by default.
After restarting, can you try to request data from the JM API directly?
e.g.
```sh
curl "https://localhost:28183/api/v1/session" --show-error --insecure
```

Sorry for shooting in the dark.

SpartanBit:
Thanks for your feedback @theborakompanioni 
I waited a bit to reply as I wanted to test the new version and see if that fixed it, but nothing changed :(

Also tried your curl request before and after updating to the version 1.5, but obtained the same output, not sure if that could help:

[...]

PR #611, opened 2023-03-06
chore: use `127.0.0.1` instead of `localhost` in setupProxy.js

theborakompanioni:
Users reported problems on [Raspibolt following this guide](https://raspibolt.org/guide/bonus/bitcoin/Jam.html).
See https://github.com/joinmarket-webui/jam/issues/608#issuecomment-1451831421
This seems to fix their problem while also having the development environment in functional state.

### DIFF
--- a/src/setupProxy.js
+++ b/src/setupProxy.js
@@ -6,7 +6,7 @@ const SUPPORTED_BACKENDS = [BACKEND_NATIVE, BACKEND_STANDALONE]
 
 const {
   PUBLIC_URL = '',
-  JAM_BACKEND = 'native',
+  JAM_BACKEND = BACKEND_NATIVE,
   JMWALLETD_API_PORT = '28183',
   JMWALLETD_WEBSOCKET_PORT = '28283',
   JMOBWATCH_PORT = '62601',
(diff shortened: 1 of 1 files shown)


kristapsk:
Concept ACK. I think the problem is that localhost defaults to IPv6 ::1, but processes might only listen to IPv4 127.0.0.1.

theborakompanioni:
> Concept ACK. I think the problem is that localhost defaults to IPv6 ::1, but processes might only listen to IPv4 127.0.0.1.

Makes sense. Thanks for clarifying. :raised_hands: 
Sounds like the problem should rather be fixed in their setup, and "localhost" should be kept.

kristapsk:
> "localhost" should be kept.

Don't see any benefits here. AFAIK you don't gain anything by using IPv6 for localhost over IPv4. Just using 127.0.0.1 everywhere makes things simpler and there will potentially be less problems. Only situation where this might not work is some hypothetical future machine where IPv4 support is abandoned in favour of IPv6.

openoms (approved):
utACK - makes sense

kristapsk (approved):
ACK 78f12007b8927e7ebeb7d67f537550ec2c0f6825 (tested by applying this patch on top of v0.1.5 on RaspiBolt)

PR #621, opened 2023-06-05
feat: import wallet

theborakompanioni:
Closes #137.

Adds the ability to import an existing wallet in Jam.

- [x] Customize options (plain input of gaplimit and blockheight)
- [x] Reset `gaplimit` to original value
- [x] Add "back" button
- [x] Prevent importing if rescan is active
- [x] Call `/display` after rescanning finished
- [x] Test on `mainnet`

## How to test
After checking out the branch, execute `npm run regtest:rebuild` in order to rebuild your dev environment to get the lastest upstream changes from JM (the "import wallet" endpoint is only on `master` and not released yet). Otherwise you will get `"Error while importing the wallet. Reason: Method Not Allowed"` when trying to import a wallet.

PR #635, opened 2023-08-13
feat(balance): display frozen balance on jars

### LINKED ISSUE #634: feat: show frozen/locked balance for Jars

As a user, I want to know which UTXOs are forzen/locked/spendable in my wallet quickly.
The state of an UTXO (frozen, locked, etc.) is not totally obvious in the Jars component which is present on the send, receive and wallet overview page.

 

Ability to know quickly, which of the funds are spendable/forzen/locked when looking at the jars.

e.g. this is just a suggestion (!)

theborakompanioni:
Resolves #634.

Displays the frozen or locked balance in the Jars component.

### DIFF
--- a/src/components/JarSelectorModal.tsx
+++ b/src/components/JarSelectorModal.tsx
@@ -77,7 +77,8 @@ export default function JarSelectorModal({
               <SelectableJar
                 key={account.accountIndex}
                 index={account.accountIndex}
-                balance={account.calculatedTotalBalanceInSats}
+                balance={account.calculatedAvailableBalanceInSats}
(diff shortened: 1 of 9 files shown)


editwentyone (approved):
ACK

theborakompanioni:
This is really just a suggestion. Any suggestion on how to better display frozen funds is highly appreciated.
Maybe one can display a few ice cubes in Jar with frozen UTXOs.. :ice_cube: :ice_cube: :ice_cube: ?

Heads up: The build errors are fixed in #621

editwentyone:
for the time beeing its already usefull and really quickly implemented, plus its already motivating the user to click on it to "look inside" 

never the less i will think about the other idea and play around with it

Issue #597, opened 2023-01-19
Created a Fidelity Bond but "Start Earning" returns "Maker could not start, no confirmed coins."


KainerW:
I Created a Fidelity Bond but the button at the bottom "start earning" returns "Maker could not start, no confirmed coins." at the top of the Earn page.

If there is a problem - a way to fix it or at least some infos around the problem would be better here. 

**Actual behavior**

**Steps to reproduce the problem**
  1. Send bitcoin to wallet
  2. Use Earn - Created a Fidelity Bond
  3. Press "start earning"

**Specifications**
  - Version: raspiblitz 1.9.0
  - Platform: raspberry pi 
  - Browser: firefox

**Additional context**

Error-Message on page (after pressing button "start earning) "Maker could not start, no confirmed coins." does not give any information on how to fix. There are def confirmed coins in my wallet and the coins are even confirmed in a Fidelity Bond as shown on same page.

Trying to get more infos (e.g. looking into the order book if my fidelity bond is active/existing there) returned nothing (could not find anything related to my bond in the order book...)

Is it a bug? Not really I guess but it is just confusing to present "Maker could not start, no confirmed coins." while my fidelity bond is active and shown on same page.

After creating the fidelity bond... do I have to press "start earning" or is this button only for users without fidelity bond?

PulpCattel:
> There are def confirmed coins in my wallet and the coins are even confirmed in a Fidelity Bond as shown on same page.

Do you have other coins other than the bond?
Fidelity bonds are timelocked on the Bitcoin blockchain, and **cannot** be spent in any way until the timelock expires. You need some other coins, other then the fidelity bond.

See the documentation:
* https://jamdocs.org/interface/fidelity-bonds/
* https://github.com/JoinMarket-Org/joinmarket-clientserver/blob/master/docs/fidelity-bonds.md

> It is impossible to move or spend funds that are locked in a fidelity bond before the bond expires. They cannot be used in collaborative transactions as fidelity bonds are time-locked by the Bitcoin protocol.

KainerW:
Thank you PC, I did in fact just understand that wrong.

theborakompanioni:
This is indeed a UX problem that must be addressed. Thanks for reporting.

editwentyone:
<img width="563" alt="image" src="https://user-images.githubusercontent.com/96262473/215142613-a5475811-ce67-43de-96a6-64953e8d21d2.png">

[...]

PR #649, opened 2023-08-24
ui(fb): display warning if all funds are used for fidelity bond

theborakompanioni:
Mitigate cases described in #597.

Adds a warning to the Fidelity Bond Form, when all UTXOs are selected:
> Keep in mind: As you are using all available funds for the creation of this fidelity bond, you will not have any UTXOs left. A fidelity bond will not participate in collaborative transactions and you have to fund your wallet again to start sending or earning.

Additionally:
- Fidelity Bond component transformed from `jsx` to `tsx` -> this still needs additional refactoring, but as a redesign is already planned, no further changes were made (issue https://github.com/joinmarket-webui/jam/issues/650)


kristapsk:
Concept ACK

MaxHillebrand (approved):
very helpful!

PR #655, opened 2023-09-09
fix(fee-randomization): fix fee range in PaymentConfirmModal

### LINKED ISSUE #654: fix: transaction fees are only randomized upwards
Newest JM release randomized tx fees only upwards.

> We currently randomize transaction fees both ways. I think it's not the right approach. When user specifies transaction fee rate he has some expectations of confirmation time. So probably we should never use actual fee rate below configured / specified one, only randomize upwards.

https://github.com/JoinMarket-Org/joinmarket-clientserver/issues/1429

Jam currently still displays the following description:
> Random fees improve privacy. The percentage is to be understood as a +/- around the base fee. Example: If you set the base fee to 10 sats/vByte and the randomization to 30%, a value between 7 and 13 sats/vByte will be used. Default: 20%.

And displays the randomized fee range in the Payment Confirmation Modal.

theborakompanioni:
Resolves #654.

Adapts the texts and fee calculation for backends >=v0.9.10.

- [x] Adapt fee range in `PaymentConfirmModal` 
- [x] Adapt description in Fee Settings

### DIFF
--- a/src/components/PaymentConfirmModal.tsx
+++ b/src/components/PaymentConfirmModal.tsx
@@ -11,6 +11,13 @@ import { AmountSats } from '../libs/JmWalletApi'
 import { jarInitial } from './jars/Jar'
 import { isValidNumber } from '../utils'
 
+const feeRange: (feeValues: FeeValues) => [number, number] = (feeValues) => {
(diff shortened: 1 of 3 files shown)


editwentyone (approved):
cACK

PR #658, opened 2023-09-11
chore(import): display warning on large gaplimit values

theborakompanioni:
Following changes should improve user experience and prevent some mistakes while importing wallets by adding visual feedback:
- Adds a max blockheight value: `10_000_000` - just to prevent errors during import  (should be lower than the current blockheight anyway)
- Adds a max gaplimit value: `10_000` - rather high , but users should not be prevented from doing it, if they really want to
- Warn on "large" gaplimit values: any value greater than `250` (arbitrarily chosen) will display a warning

### DIFF
--- a/src/components/Accordion.tsx
+++ b/src/components/Accordion.tsx
@@ -1,15 +1,23 @@
-import React, { PropsWithChildren, useState } from 'react'
+import { ReactNode, PropsWithChildren, useState } from 'react'
+import classNames from 'classnames'
 import { useSettings } from '../context/SettingsContext'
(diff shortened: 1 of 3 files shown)


editwentyone:
ACK

editwentyone (approved):

theborakompanioni:
Will change wording from "an extreme decline of performance and responsiveness is to be expected" to something like "can lead to a decline of performance and responsiveness".

editwentyone (commented):
perfect

PR #659, opened 2023-09-13
feat(i18n): update german translation (de)

Overload3910:
added the strings for the wallet import


Overload3910:
I am not sure if these two strings can even be translated 

@theborakompanioni Can you check? :)


SCR-20230919-rs7
SCR-20230919-rsg

PR #660, opened 2023-09-13
feat(i18n): update to Chinese translations

Overload3910:
imported the update from Transifex... still not finished yet


Overload3910:
A few more words are missing on the Chinese translation I imported. 

Want to merge this anyway because it's better than nothing?

PR #665, opened 2023-09-22
 feat: new authentication mechanism (access/refresh token)

theborakompanioni:
Resolves #663.

This PR adds handling for the new JWT auth mechanism.

From [JSON-RPC-API-using-jmwalletd.md#rules-about-making-requests](https://github.com/JoinMarket-Org/joinmarket-clientserver/blob/master/docs/JSON-RPC-API-using-jmwalletd.md#rules-about-making-requests):

> On initially creating, unlocking or recovering a wallet, a new access and refresh token will be sent in response, the former is valid for only 30 minutes and must be used for any authenticated call, the former is valid for 4 hours and can be used to request a new access token, ideally before access token expiration to avoid unauthorized response from authenticated endpoint and in any case, before refresh token expiration.

The token is refreshed every 22.5 minutes (30min * 0.75). In dev mode more often (every 60 seconds).

There is currently a problem upstream with handling wallets that contain spaces in their filename, should be fixed with https://github.com/JoinMarket-Org/joinmarket-clientserver/pull/1562. Nonetheless, the code can already be reviewed and should not be affected by the changes.

### How to test
Rebuild the dev docker environment `npm run regtest:rebuild` and verify:
- Create wallet still works
- Import wallet still works
- Lock/Unlock still works
- Token can be refreshed
- Requests to RPC api are still successful after token refresh
- Websocket communication still functioning after token refresh (e.g. start/stop a maker and watch the websocket data)

### Misc
- Method `Api.Helper.parseAuthProps` is temporary and can be removed in #670.
- Changes backend version in regtest env to `master` again.

PR #671, opened 2023-10-04
refactor: Consistent use of `walletName` vs `walletFileName`

### LINKED ISSUE #640: refactor: Consistent use of `walletName` vs `walletFileName`
It is not totally clear, when a wallet name is needed, if the filename (including extension `.jmdat`) or the "display name" (without extension) is needed. Hence, everytime the wallet name is rendered, it is first given to util function `displayWalletName`. This can be made more obvious and the wallet filename, as well as the wallet "display name" should be made statically available in the `WalletContext`.

Hint: There is already a type `Api.WalletName` which  actually means the "wallet filename". Change/introduce types to make this more obvious.

theborakompanioni:
Resolves #640.

Blocked by #665 (needs rebase afterwards).

**This is a refactoring PR - no new features, no functionality changes.**

While trying to consistently use `displayName` and `walletFileName`, it was easier/necessary to also migrate all components to typescript. So this turned from resolving #640 into a "jsx to tsx" PR rather quickly. Now, only one non-typescript component is left (`ShareButton`). 

- refactor: Wallets component from jsx to tsx
- refactor: Settings component from jsx to tsx
- refactor: ScheduleProgress component from jsx to tsx
- refactor: Onboarding component from jsx to tsx
- refactor: Earn component from jsx to tsx
- refactor: Receive component from jsx to tsx
- refactor: CreateWallet component from jsx to tsx
- refactor: BitcoinQR component from jsx to tsx


editwentyone (commented):
cACK

PR #675, opened 2023-10-10
chore: debounce polling of session info

theborakompanioni:
Before these changes, the `/session` endpoint has been polled with `setInterval` - i.e. rescheduled to run every 10 seconds, regardless of the response duration. On slow connection, e.g. over Tor, this can lead to requests queuing up.

After this commit, a new request is made 10 seconds _after_ the previous response arrived, which mitigates the behaviour described above.


editwentyone (approved):
tACK

PR #676, opened 2023-10-11
chore: do not preselect jar on send page

theborakompanioni:
Before this commit, the first funded jar has been preselected as source jar.
After this commit, no jar will be preselected.

This was a request by @editwentyone.

Additionally, it fixes a strange bug: Scrolling to the top of the page did not work properly in all browsers.


editwentyone (approved):
running as expected, even when you trick yourself and select target A before source A ^^

PR #677, opened 2023-10-13
ui(footer): improve label and colors of websocket indicator

theborakompanioni:
Before this commit, the label on the websocket indicator just stated "connected" or "disconnected", which caused some confusion. 

After this commit, it will be labelled  "Websocket conntected" or "Websocket disconnected". Also, if the websocket is disconnected, the icon is coloured gray instead of red, as Jam will work normally without a stable websocket connection (UI responsiveness will be slower, but is functioning).

### :camera_flash: After

<br />

### DIFF
--- a/src/components/App.test.tsx
+++ b/src/components/App.test.tsx
@@ -74,12 +74,12 @@ describe('<App />', () => {
       render(<App />)
     })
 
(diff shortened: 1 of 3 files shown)


editwentyone (approved):
as expected, noize

PR #680, opened 2023-10-18
chore: remove create wallet button from settings page

theborakompanioni:
This is just a proposal, feel free to close it, if you think it is not a good idea.

I do not quite  see the use of a "create wallet" button from the settings page.
A wallet can already be created from the main view when no wallet is active.
This PR will remove the button from the settings page.

What do you think?


editwentyone (approved):
sure, just the right thing to do

PR #681, opened 2023-10-18
dev: always show logs button in dev mode

theborakompanioni:
The feature to show the log file is not available on all platforms.
Before this commit, the "show logs" button has only been displayed if a special endpoint indicates it is enabled.

However, this is somewhat confusing, as the endpoint is not available in the regtest environment.
So, when this commit is applied, the button will always be shown while developing.

With `npm run dev:start`, the log endpoint will return an error (as it is not implemented in the dev environment, since this is interacting with a "native" JM instance) and a message is displayed to run `npm run dev:start:secondary` (which interacts with an instance in the `standalone` image, that has the log feature enabled!).

Additionally, this moves the menu item from section "Markets" to "Wallet".

### DIFF
--- a/src/components/LogOverlay.module.css
+++ b/src/components/LogOverlay.module.css
@@ -1,5 +1,5 @@
 .logContentPlaceholder {
-  height: 1rem;
+  height: 2.625rem;
   margin: 1px 0;
 }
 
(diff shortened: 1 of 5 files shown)


editwentyone (approved):
cACK

PR #684, opened 2023-10-20
chore: validate wallet name

### LINKED ISSUE #683: / char should not be allowed in wallet name
**Expected behavior**

entering a `/` char in the wallet name should be disallowed by the javascript.

**Actual behavior**

Wallet creation in bitcoin core will fail because / cannot be in a directory name

  1. Create a new wallet
  2. Put a / in the name
  3. Click create

**Specifications**
  - Version: 016
  - Platform: Linux
  - Browser: Firefox

**Additional context**

image

theborakompanioni:
Resolves #683.

Before this commit, the wallet name has been validated only on the backend.
After this commit is applied, it will also be validated by the frontend.

The validation is a bit more strict than "necessary", only numbers, letters and `_` (underscore) will be allowed from now on.

### DIFF
--- a/src/components/CreateWallet.test.tsx
+++ b/src/components/CreateWallet.test.tsx
@@ -19,7 +19,8 @@ jest.mock('../libs/JmWalletApi', () => ({
 const NOOP = () => {}
 
 describe('<CreateWallet />', () => {
-  const testWalletName = 'wallet'
+  const testWalletName = 'test_wallet21'
+  const invalidTestWalletName = 'invalid_wallet_name!'
   const testWalletPassword = 'correct horse battery staple'
 
   const setup = ({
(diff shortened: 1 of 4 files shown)


kristapsk:
Why not also allow hyphen-minus (`-`)? I personally use that for JM wallet names instead of underscore.

theborakompanioni:
> Why not also allow hyphen-minus (`-`)? I personally use that for JM wallet names instead of underscore.

Now also allows hyphens. Error feedback message has been changed to 
```
Please choose a valid wallet name: Use only letters, numbers, underscores or hyphens.
```

Additionally, the length of the wallet name has been limited to `234` chars.
(Notice that, while JM allows longer names, it will create a wallet but later fail loading it.)

Thanks for the feedback @kristapsk :pray:

kristapsk:
> (Notice that, while JM allows longer names, it will create a wallet but later fail loading it.)

Interesting, something to test and likely fix in JM.

kristapsk:
> Additionally, the length of the wallet name has been limited to `234` chars.
> (Notice that, while JM allows longer names, it will create a wallet but later fail loading it.)

With which OS, filesystem and Python version do you experienced this? I cannot reproduce under Linux, ext4 and Python 3.11.4, at least not with cli tools. You had this both with scripts and RPC API or only RPC API?

theborakompanioni:
> > Additionally, the length of the wallet name has been limited to `234` chars.
> > (Notice that, while JM allows longer names, it will create a wallet but later fail loading it.)
> 
> With which OS, filesystem and Python version do you experienced this? I cannot reproduce under Linux, ext4 and Python 3.11.4, at least not with cli tools. You had this both with scripts and RPC API or only RPC API?

Hey @kristapsk! (Mostly) _everything_ I do and test is with the RPC API!
I am using the regtest docker image based on `Debian GNU/Linux 12 (bookworm)`, fs `ext4`, python  `v3.11.6`.

editwentyone (approved):

Issue #561, opened 2022-11-03
Normalise Bond values or color the non-zero values?


alexwaltz:
_1st of all I am not sure if this is a good idea to change unless JoinMarket does it 1st._

But I find myself squinting at the screen every time I am trying to see what are other bonds available, and it's really hard to see among the sear of 0s.

The normalization factor should be easy to figure out and should be somehow proportional to the r - the interest rate as this is what introduces all those 0s after the decimal point.

The formula for valuing a bond is: `V^2 (exp(rT) - 1)^2`
Where:
- `V`: the amount of Bitcoins to be locked
- `r`: the interest rate - by default, this is set to 0.015
- `exp`: the exponent - by default, this is set to 1.3
- `T`: the time to be locked

So the formula practically is: 
<img width="138" alt="Screenshot 2022-11-03 at 17 43 16" src="https://user-images.githubusercontent.com/92048497/199795789-71e2bab8-7251-4303-a2b4-458b496a10b3.png">

Another option that would maintain consistency with JoinMarket would be to highlight the non-zero values with a certain color label 
<img width="170" alt="Screenshot 2022-11-03 at 17 38 55" src="https://user-images.githubusercontent.com/92048497/199794863-9c02c77c-08fa-4222-92eb-7374d66ecd3f.png">

PR #687, opened 2023-10-26
fix: use orderbook.json instead of parsing html table

theborakompanioni:
Closes https://github.com/joinmarket-webui/jam/issues/454.
Closes https://github.com/joinmarket-webui/jam/issues/561.

- Fetches `orderbook.json` instead of parsing the html table to get data.
- Uses normalized fidelity bond value
- Removes column "Miner Fee Contribution"


editwentyone (approved):
cACK

PR #688, opened 2023-10-29
chore: do not display xpubs

theborakompanioni:
..as those derive public keys from an unexpected derivation depth.

See upstream issue https://github.com/JoinMarket-Org/joinmarket-clientserver/issues/1586

So, best thing for now is to temporarily do not display any xpub, instead of a "wrong" one. What do you think?


editwentyone:
good idea, wrong informations and frustrations are worse

editwentyone:
please also hide empty "internal addresses" - thanks

theborakompanioni:
> please also hide empty "internal addresses" - thanks

Done :heavy_check_mark:

editwentyone (approved):

PR #691, opened 2023-11-04
ui: improve balance spacing, coloring and grouping

### LINKED ISSUE #690: Balance Component - Spacing, Coloring and Grouping
it would be nice to display all the balances grouped, colored and spaced "correctly" - like the bitcoin design community is suggesting it:
Bildschirmfoto 2023-10-31 um 09 46 03

- grouped by 0,00 000 000
- all the pre 0 coloured differently, like some alpha percentage
- spacing is a thing space https://www.compart.com/en/unicode/U+2009
- use also slashed Zeros if possible everywhere, where the user finds basically number-informations only (like table cells)
- use a monospace variant in tables or basically everywhere

this helps to see the sats :D see reference: https://bitcoin.design/guide/designing-products/units-and-symbols/#formatting-units

theborakompanioni:
Resolves #690.

Styles the "insignificant" digits of the balance component making the value more readable and easy on the eyes.


editwentyone:
👏 awesome, thanks for the fast development 

next logical step is to get rid of "Display amounts in sats" under settings and thus we also don't need the kebab sats logo anymore/ anywhere. 

what do you think? do you like it?

editwentyone (approved):

theborakompanioni:
> 👏 awesome, thanks for the fast development
> 
> next logical step is to get rid of "Display amounts in sats" under settings and thus we also don't need the kebab sats logo anymore/ anywhere.

Yeah, however, let's ask users first if they even use the option to display amounts in sats. Currently, we also have all input fields representing amounts in sats, not in BTC. Okay for you?

> what do you think? do you like it?

I really like the visual representation, yes. I do not like the current implementation as it uses [`opacity`](https://developer.mozilla.org/en-US/docs/Web/CSS/opacity) instead of the more accurate [`mix-blend-mode`](https://developer.mozilla.org/en-US/docs/Web/CSS/mix-blend-mode).. I did not had good results with `mix-blend-mode` yet - but will keep trying to improve it.

theborakompanioni:
Newest commit styles the amount with css only. I'd say, that is "good enough" - at least better than before. Would you check again if it satisfies your expectations @editwentyone? Colors can be still be adapted, of course! Please check both, dark and light mode! Thanks :pray:

theborakompanioni:
Yo, @editwentyone, this is now ready for review again.

One remark: Instead of color `#536475` it uses `#212529` in light mode for highlighted balance digits. Otherwise it's quite hard to distinguish from de-emphasized digits.

editwentyone:
> Yo, @editwentyone, this is now ready for review again.
> 
> One remark: Instead of color `#536475` it uses `#212529` in light mode for highlighted balance digits. Otherwise it's quite hard to distinguish from de-emphasized digits.

do you have an example? because it should work, its the same color as the Jar Name "**A**pricot"
other than that, everything is fine

theborakompanioni:
> lets do: `#344353` instead of `#212529` if its ok for you

This also feels rather unfavorable visually. Too little contrast, imho,  don't you think? What about all the people with color weakness, like me? :grimacing:

editwentyone:
ok how about '#0B3A69'?

theborakompanioni:
> ok how about `#0B3A69`?

This is quite bluish.. but contrast ist good, imho. 
Should it be applied? :thinking: 

image

editwentyone (approved):

Issue #692, opened 2023-11-04
Align all Amount Input Fields to ₿


editwentyone:
it would be awesome to consolidate all input fields where the user needs to input an amount of sats / btc to the same component and styling:

- Receive » options
- Send
- Earn » options

Either B (icon) and Sats (kebab icon) switcher, or everything to B only. maybe even with a proper Placeholder Text with the same styling "0.00 000 000" but **when the user copy/paste values its of course **without** the spacing**

(added to 0.2.0, you can raise your hand and have a veto if you want to postpone that ;) ) 

Screenshots:
![Bildschirmfoto 2023-11-04 um 19 16 39](https://github.com/joinmarket-webui/jam/assets/96262473/32ea9d4d-23fa-46fd-affb-84d9f27442a2)
![Bildschirmfoto 2023-11-04 um 19 16 32](https://github.com/joinmarket-webui/jam/assets/96262473/791ba22b-6bd1-4cd6-ac76-43e4c04d0f46)
![Bildschirmfoto 2023-11-04 um 19 16 21](https://github.com/joinmarket-webui/jam/assets/96262473/1c19e02d-c326-47f9-9f95-3f71538dc998)

theborakompanioni:
> (added to 0.2.0, you can raise your hand and have a veto if you want to postpone that ;) )

It's okay! However, please do not auto-assign me. You are destroying all my hopes that other fellow devs will ever contribute to the project :wink: 
Tagging it with "good-first-issue" would be nice (if the ticket is actually small and concise).

editwentyone:
you still have hopes and dreams :)

PR #708, opened 2023-12-12
chore(send): align unit in collaborators fee breakdown

### LINKED ISSUE #707: Show Sats in the same way everywhere
@theborakompanioni could you please please please change the view for all Sats to the new way – thanks

image

theborakompanioni:
Resolves #707.

Shows the expected collaborator fees in the unit of the users choice.
Jam can still display values in sats (instead of BTC) - so the toggle will influence the displayed unit (as opposed to always show in BTC). Getting rid of the conversion to sats has been thought about, but should be done in a separate issues if needed.

### DIFF
--- a/src/components/Send/DestinationInputField.tsx
+++ b/src/components/Send/DestinationInputField.tsx
@@ -146,12 +146,7 @@ export const DestinationInputField = ({
                     disabled={disabled || !walletInfo}
                   >
                     <div className="d-flex justify-content-center align-items-center">
-                      <Sprite
-                        symbol="jar-closed-empty"
-                        width="28px"
-                        height="28px"
-                        style={{ paddingBottom: '0.2rem' }}
-                      />
+                      <Sprite symbol="jar-closed-empty" width="28px" height="28px" style={{ marginTop: '-3px' }} />
                     </div>
                   </rb.Button>
                   <rb.Form.Control.Feedback type="invalid">
--- a/src/components/Send/FeeBreakdown.tsx
+++ b/src/components/Send/FeeBreakdown.tsx
@@ -4,7 +4,7 @@ import classNames from 'classnames'
 import Balance from '../Balance'
 import * as rb from 'react-bootstrap'
 import { useSettings } from '../../context/SettingsContext'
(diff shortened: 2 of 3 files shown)


MarnixCroes (approved):
ack fbc852a306f8c2f3ed2897e8fa3dea2ab3207bf8

> Jam can still display values in sats (instead of BTC) - so the toggle will influence the displayed unit (as opposed to always show in BTC). Getting rid of the conversion to sats has been thought about, but should be done in a separate issues if needed.

do you mean getting rid of the BTC/sats option in settings in general?
for me that would be a nack, it's nice.

the thing that could be improved about this is having also the input field in BTC when unit is BTC. now those are still sats which can be annoying/confusing.

editwentyone:
what about these?
290032138-abf46f0e-faa0-49f3-a6cc-fbc7f3c3f4f5

editwentyone:
> the thing that could be improved about this is having also the input field in BTC when unit is BTC. now those are still sats which can be annoying/confusing.

already on it : https://github.com/joinmarket-webui/jam/issues/692

editwentyone:
tiny stuff in details view: 

Bildschirmfoto 2023-12-13 um 11 24 14

theborakompanioni:
> what about these? 290032138-abf46f0e-faa0-49f3-a6cc-fbc7f3c3f4f5

- First circle: Can be fixed - just overlooked - thank you!
- Second circle: Will try.. this is a i18n string.. let's see how to use the balance component here..

theborakompanioni:
> do you mean getting rid of the BTC/sats option in settings in general? for me that would be a nack, it's nice.

Yes. The current behaviour of the BTC unit should (theoretically) combine best of both worlds.. which would make the differentiation obsolete.. would be less code while having nearly same functionality. However, I can live with keeping it.

> the thing that could be improved about this is having also the input field in BTC when unit is BTC. now those are still sats which can be annoying/confusing.

Yep, that is planned to be worked on soon and hopefully lands in v0.2.0. :raised_hands:

theborakompanioni:
> what about these? [...]

@editwentyone

editwentyone (approved):

PR #710, opened 2023-12-13
dev(regtest): update bitcoin core to v26

theborakompanioni:
This updates regtest Bitcoin Core from v25 to v26.

Should serve as a reminder for everyone using JM/Jam wanting to upgrade to Bitcoin Core v26: You need to add arg `-deprecatedrpc=create_bdb` (at start or in conf file).

See https://github.com/JoinMarket-Org/joinmarket-clientserver/issues/1571 for more information.

Meta question: How will this work for users with Umbrel/Citadel/RaspiBlitz/etc.?

### DIFF
--- a/docker/regtest/docker-compose.yml
+++ b/docker/regtest/docker-compose.yml
@@ -93,7 +93,7 @@ services:
   bitcoind:
     container_name: jm_regtest_bitcoind
     restart: unless-stopped
-    image: btcpayserver/bitcoin:25.0
+    image: btcpayserver/bitcoin:26.0
     environment:
       BITCOIN_NETWORK: regtest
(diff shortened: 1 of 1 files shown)


kristapsk:
> Meta question: How will this work for users with Umbrel/Citadel/RaspiBlitz/etc.?

They should add `deprecatedrpc=create_bdb` for now. People using these should review / add comments and make sure it gets added to config when upgrading to v26. Probably JM will have proper descriptor wallet support on next release, but not sure how fast we will get there.

kristapsk (approved):
utACK db73b4e3be9e7003a582c352f90b8e25be54f315

Issue #636, opened 2023-08-13
bug: Check for existing wallet before importing/creating wallets


theborakompanioni:
**Expected behavior**

A user will be warned early if a wallet name is already taken.

**Actual behavior**

On the create or import page, if a duplicate wallet name is given, the user is not warned and will error only later in the process (e.g. after providing a mnemonic phrase).

**Steps to reproduce the problem**
  1. Imit the regtest environment
  2. Try to create or import a wallet names "Satoshi" (which already exists)
  3. Notice that an error occurs when submitting the last step - which could be known earlier

Thanks to @editwentyone for reporting this.

theborakompanioni:
Resolved in #720. Thank you @zmjohnso :pray:

PR #719, opened 2024-01-29
636 check for existing wallet

zmjohnso:
This PR is a fix for #636. When creating a new wallet, the wallet name will now be validated immediately for duplicates, instead of on form submission when the user clicks "Create". 

https://github.com/joinmarket-webui/jam/assets/25010707/6b6f75c7-2293-4e41-9b33-45b08482fe6c

Additionally, this PR refactors the `getWalletAll` api calls to use react router loaders. This should facilitate async data loading as well as reduce extraneous useEffect and useState calls. Using react router loaders/actions while making more use of `errorElement`'s instead of extra `useEffect`/`useState` calls feels like a good pattern that could be explored more in the future...

I did do manual testing, but I am not 100% confident I was able to capture all existing functionality, especially in `src/components/Wallets.tsx`. This could use some extra eyes.

Many thanks to @theborakompanioni for drawing my attention to using react router loaders to fix this bug 🙌

### DIFF
--- a/src/components/App.test.tsx
+++ b/src/components/App.test.tsx
@@ -1,8 +1,14 @@
 import { render, screen, act } from '../testUtils'
 import user from '@testing-library/user-event'
 import * as apiMock from '../libs/JmWalletApi'
+import * as loadersMock from './loaders/DataLoaders'
 
 import App from './App'
+
+jest.mock('./loaders/DataLoaders', () => ({
+  ...jest.requireActual('./loaders/DataLoaders'),
+  allWalletsLoader: jest.fn(),
+}))
 
 jest.mock('../libs/JmWalletApi', () => ({
   ...jest.requireActual('../libs/JmWalletApi'),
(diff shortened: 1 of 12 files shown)


theborakompanioni:
Hey @zmjohnso! Thanks so much!

Seems the "import wallet" flow is currently not working (`react_router_dom__WEBPACK_IMPORTED_MODULE_6__.useLoaderData() is undefined`).

zmjohnso:
> Hey @zmjohnso! Thanks so much!
> 
> Seems the "import wallet" flow is currently not working (`react_router_dom__WEBPACK_IMPORTED_MODULE_6__.useLoaderData() is undefined`).

My bad, missed that file when changing the return type of `useLoaderData()`. Should be good now!

theborakompanioni:
So, I have looked over the changes and think it is really nice. If this works well, we should use it more extensively and finally get rid of custom loading indicators per page component.

However, what I noticed is that the requests are often repeatedly invoked. e.g. if you just use the theme toggle (top right corner), you'll see that everytime the button is pressed, multiple "get all wallets" requests are made. I did not investigate too much but maybe you have an idea why that is?

Additionally, if we can rectify this behaviour, "import wallet" should also check for already existing wallet names. :raised_hands: 
Hope we can somehow find the root cause of this. :raised_hands: 
Do you have any ideas?

PR #720, opened 2024-02-08
636 check for existing wallet

zmjohnso:
This PR is a fix for https://github.com/joinmarket-webui/jam/issues/636. When creating/importing a new wallet, the wallet name will now be validated immediately for duplicates, instead of on form submission when the user clicks "Create".

https://github.com/joinmarket-webui/jam/assets/25010707/6443dcfb-875e-4293-8b92-ae8125c9cd2c

An attempted fix was made using react router loaders, but this caused many redundant api calls to be made on component mount and session refresh. For now, this fix follows existing patterns, but utilizing a loader is something that could/should be investigated in future refactors.

### DIFF
--- a/src/components/WalletCreationForm.tsx
+++ b/src/components/WalletCreationForm.tsx
@@ -6,7 +6,6 @@ import Sprite from './Sprite'
 import { JM_WALLET_FILE_EXTENSION, sanitizeWalletName } from '../utils'
 import styles from './WalletCreationForm.module.css'
 import * as Api from '../libs/JmWalletApi'
-import { useServiceInfo } from '../context/ServiceInfoContext'
 
 export interface CreateWalletFormValues {
   walletName: string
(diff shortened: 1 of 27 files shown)


src/components/WalletCreationForm.tsx
```diff
@@ -36,14 +37,29 @@ const WalletCreationForm = ({
...
 
+  useEffect(() => {
+    const abortCtrl = new AbortController()
+
+    Api.getWalletAll({ signal: abortCtrl.signal })
+      .then((res) => (res.ok ? res.json() : Api.Helper.throwError(res, t('wallets.error_loading_failed'))))
+      .then((data) => {
+        if (abortCtrl.signal.aborted) return
+        setWalletList(data.wallets)
+      })
```
theborakompanioni, line 51:
```suggestion
      })
      .catch(() => {
        // do nothing on purpose
      })

    return () => abortCtrl.abort()
```

kristapsk:
Concept ACK

theborakompanioni (approved):

PR #735, opened 2024-03-29
ui(orderbook): add delay hint to highlighted offer toggle

theborakompanioni:
Adds a small hint that it might take some time for an offer to appear in the local order book, e.g. when the maker has just been started recently. This hint will disappear if the current nickname is present in the offer list.

Edit: Screenshots outdated - see below :pray:

### DIFF
--- a/src/components/Orderbook.tsx
+++ b/src/components/Orderbook.tsx
@@ -14,9 +14,11 @@ import Balance from './Balance'
 import Sprite from './Sprite'
 import TablePagination from './TablePagination'
 import { factorToPercentage, isAbsoluteOffer, isRelativeOffer } from '../utils'
-import { isDevMode } from '../constants/debugFeatures'
-import styles from './Orderbook.module.css'
+import { isDebugFeatureEnabled, isDevMode } from '../constants/debugFeatures'
 import ToggleSwitch from './ToggleSwitch'
+import { pseudoRandomNumber } from './Send/helpers'
+import { JM_DUST_THRESHOLD } from '../constants/config'
+import styles from './Orderbook.module.css'
 
 const TABLE_THEME = {
   Table: `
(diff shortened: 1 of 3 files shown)


editwentyone:
can you pull the view on toggle down to the green highlighted row? people would understand faster

editwentyone (approved):

theborakompanioni:
> can you pull the view on toggle down to the green highlighted row? people would understand faster

I am sorry, I do not understand what you mean exactly. Can you please explain this in more detail?

editwentyone:
sure, sorry. 
- on toggle on the hightlight 
- the user is automatically scrolled or moved to the right page
- and the user sees its own green highlighted offer

is this possible?

theborakompanioni:
> sure, sorry.
> 
>     * on toggle on the hightlight
> 
>     * the user is automatically scrolled or moved to the right page
> 
>     * and the user sees its own green highlighted offer
> 
> 
> is this possible?

Of course that would be possible. However, isn't it kind of unexpected to scroll down on click on a toggle?
What about showing a users offer separately, e.g. above the table, in addition to highlighting the table row?

theborakompanioni:
Notes from discussion:
- [ ] Orderbook should get a sticky header
- [ ] Toggle is disabled till the users order in the local orderbook
- [ ] On toggle, pagination is set to all and offer should scroll into view

theborakompanioni:
> Notes from discussion:
> 
>     * [ ]  Orderbook should get a sticky header
> 
>     * [ ]  Toggle is disabled till the users order in the local orderbook
> 
>     * [ ]  On toggle, pagination is set to "display all" and offer should scroll into view

Thought about this some more.
1. Sticky header is difficult, as the table would need a min-height. Does not look nice. :x: 
2. Disable toggle till the offer is in the list: Done :white_check_mark: 
3. Scrolling to the offer while setting pagination to "display all" feels confusing 

What feels like a good solution to point #3 is:
Ability to pin the offer to the top of the list. 
It seems to be the easiest and achieves what should be done: Being able to compare ones own offer to other offers.

### :camera_flash: 

#### Offer absent (toggle disabled + subtitle hint)


#### Offer present


#### Offer present, highlighted


#### Offer present, pinned to top



What do you think @editwentyone ?

editwentyone:
@theborakompanioni  i'm fine with the changes :) you can merge

PR #739, opened 2024-04-04
641 Validate Mnemonic Words on Wallet Import

### LINKED ISSUE #641: feat(import): validate mnemonic phrase against BIP39 wordlist
Currently, the jm backend will do the validation and fail, if the mnemonic phrase is invalid.
There should also be a visual hint, that the given word is contained in the BIP39 wordlist.

If a some words are not part of the BIP39 wordlist, make the user aware, but still continue the flow!

From: https://www.figma.com/file/kfejZJFlwBywvLEnPEmJo1/JoinMarket-UI?node-id=2850%3A67638

zmjohnso:
Fixes #641.

To check for each word, I simply added the [bip39 javascript](https://github.com/bitcoinjs/bip39) package to the project to get the full wordlists, then check that the mnemonic word input is contained within the list. I've defaulted the check against the English word list since it seems that the joinmarket api also defaults to the English words. There might be a better way to do this? I'm not exactly sure how the jm api is doing the check.

After:
image
image


kristapsk:
> I've defaulted the check against the English word list since it seems that the joinmarket api also defaults to the English words. There might be a better way to do this? I'm not exactly sure how the jm api is doing the check.

JoinMarket is hardcoded to English wordlist, it uses old code from Electrum. https://github.com/JoinMarket-Org/joinmarket-clientserver/blob/239ac4689a61e769fa17f8e9ef2688b835d0748b/src/jmclient/old_mnemonic.py

zmjohnso:
Thanks for that reference @kristapsk! But I am still a bit confused how jm is doing the check, since that word list does not seem to contain everything in the bip39 wordlist, i.e., "abandon". I will try to understand this better. Thanks again.

@theborakompanioni Yes, I completely agree with your first point. I have added a hardcoded version of the bip39 wordlist and removed the bip39 package. On your second point: currently the user can still continue through the workflow to import a wallet even if the UI shows the word as "invalid". Then after reviewing the words and clicking the final "import" button, the jm api will throw an error if it is not a valid mnemonic phrase. I think this workflow currently makes sense as is. We should allow the user to continue and ultimately this just serves as a visual check if the user has a typo, but the final validation happens on the api.
So the api determines validity and I'm not sure a different/separate visual indicator is necessary. But definitely open to more thoughts on this.

kristapsk:
> Thanks for that reference @kristapsk! But I am still a bit confused how jm is doing the check, since that word list does not seem to contain everything in the bip39 wordlist, i.e., "abandon". I will try to understand this better. Thanks again.

From API perspective you either give completely correct world list or get error back.

theborakompanioni:
> @theborakompanioni Yes, I completely agree with your first point. I have added a hardcoded version of the bip39 wordlist and removed the bip39 package. On your second point: currently the user can still continue through the workflow to import a wallet even if the UI shows the word as "invalid". Then after reviewing the words and clicking the final "import" button, the jm api will throw an error if it is not a valid mnemonic phrase. I think this workflow currently makes sense as is. We should allow the user to continue and ultimately this just serves as a visual check if the user has a typo, but the final validation happens on the api. So the api determines validity and I'm not sure a different/separate visual indicator is necessary. But definitely open to more thoughts on this.

Very reasonable! :+1: 

Only little nit regarding the code is that now the `isValid` property of `MnemonicWordInput` does not work "as expected" and might surprise callers (i.e. validation is currently done outside the component and e.g. passing `isValid := true` can lead to the input ending up being displayed as "invalid"). What about adding an additional component `Bip39MnemonicWordInput` that decorates the `MnemonicWordInput` with validation - do you think that would make it clearer?

Otherwise, let's merge this soon! Thank you so much @zmjohnso Great work. Simple and clean.

zmjohnso:
@theborakompanioni So I added the `Bip39MnemonicWordInput` decorator component which wraps `MnemonicWordInput` and just does the `isValid` check for if the value is in the bip39 word list. I think this is cleaner and allows the `MnemonicWordInput` to remain unchanged and thus callers of this component can expect the same behavior as before. I think this does make things a bit more clear, but please let me know if this isn't what you meant.

theborakompanioni:
> @theborakompanioni So I added the `Bip39MnemonicWordInput` decorator component which wraps `MnemonicWordInput` and just does the `isValid` check for if the value is in the bip39 word list. I think this is cleaner and allows the `MnemonicWordInput` to remain unchanged and thus callers of this component can expect the same behavior as before. I think this does make things a bit more clear, but please let me know if this isn't what you meant.

Made some small adaptions (taking into account the `isValid` given by the caller). Looks good to me! Will merge soon.

theborakompanioni (approved):

PR #745, opened 2024-04-29
fix(Earn): validate offer minsize

### LINKED ISSUE #740: "Earn" Feature fails without explanation if "Minimum amount" size is too high
**Expected behavior**

When using the earn feature, user should set parameters, click "Start Earning!" and the service should work or explain why it isn't working. 

**Actual behavior**
Earn Feature Fails to start without explanation

Offer Options>Minimum amount:

If this field is larger than the sum of the unfrozen outputs in any jar the service will fail to start.

Offer Options>Minimum amount>input value greater than the sum of largest jar>click "Start Earning!"
  

**Specifications**
  - Version: 0.2.0
  - Platform: StartOS
  - Browser: Firefox

**Additional context**

This is confusing to beginners because the total jam balance leads you to believe that you have plenty of capacity to demand a large min input size.  It is of course Jar capacity, not total wallet capacity that limits the min input size.

theborakompanioni:
Fixes #740.
However, this is a temporary fix to prevent the backend from dying.

There should be a follow-up PR that adds better explanation more prominently.
e.g. similar to the  `CoinjoinRequirementSummary`/`CoinjoinRequirementViolation` alerts on "Jam" and "Send".

### DIFF
--- a/src/components/Earn.tsx
+++ b/src/components/Earn.tsx
@@ -4,7 +4,7 @@ import * as rb from 'react-bootstrap'
 import { useTranslation } from 'react-i18next'
 import { TFunction } from 'i18next'
 import { useSettings } from '../context/SettingsContext'
-import { CurrentWallet, useCurrentWalletInfo, useReloadCurrentWalletInfo } from '../context/WalletContext'
+import { CurrentWallet, useCurrentWalletInfo, useReloadCurrentWalletInfo, WalletInfo } from '../context/WalletContext'
 import { useServiceInfo, useReloadServiceInfo, Offer } from '../context/ServiceInfoContext'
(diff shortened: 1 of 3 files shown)


editwentyone:
any hint why the lowest is 27300 sats? but other than that, approved

editwentyone (approved):

theborakompanioni:
> any hint why the lowest is 27300 sats? but other than that, approved

27300 is the "dusty minsize" and apparently needed to be substracted from the minsize. I am totally sure as to why that is the case. However, the backend does not like larger values and dies a while later.

Have you reproduced the problem and verified that this PR prevents the issue yourself?

editwentyone:
double checked, reproduced, now double approved

theborakompanioni:
> double checked, reproduced, now double approved

Thank you :pray:

PR #759, opened 2024-05-16
chore(ui): disable lock button while locking wallet

theborakompanioni:
Locking the wallet can take some time (sometimes a few seconds, e.g. via Tor). This avoids additional requests being made if the button is clicked while a request is in-flight.

(It is not totally obvious in the screenshot, but the button is disabled)


amitx13 (approved):
Yup! There are no additional requests; everything is working fine.

PR #762, opened 2024-05-22
modal instead of dropdown for new FB creation form

### LINKED ISSUE #760: Modal instead of dropdown for new fidelity Bond creation form
Currently to create a new fidelity bond, we have a dropdown which extends down the page. which is kind of disturbing for user.
Screenshot from 2024-05-21 12-48-06
Screenshot from 2024-05-21 12-48-25

Its better to have a separate area like a modal to create a new bond. So instead of a dropdown on the same page, we would have a popup there itself when user clicks "Configure Fidelity Bond".

barrytra:
this PR fixes #760 
When user clicks "configure fidelity bond" modal appears like this:
Screenshot from 2024-05-22 12-17-39
Whole flow to create new bond works fine on my setup.
Open to any kind of suggestions.

PS: Cross Button at top-right functions same as cancel button. I've kept it as it will be useful for later UI design.

### DIFF
--- a/src/components/fb/CreateFidelityBond.module.css
+++ b/src/components/fb/CreateFidelityBond.module.css
@@ -50,24 +50,3 @@
   color: 'white';
   border-radius: '50%';
 }
-
-.form {
-  padding: 0.5rem;
-}
-
-.formHeader {
-  display: flex;
-  justify-content: space-between;
-  align-items: center;
-  padding: 0.5rem 1rem 0 1rem;
-}
-
-.formHeader .title {
-  width: 100%;
-  font-size: 1.2rem;
-  color: var(--bs-body-color);
-}
-
-.formHeader .svg {
-  color: var(--bs-body-color);
-}
--- a/src/components/fb/CreateFidelityBond.module.css
+++ b/src/components/fb/CreateFidelityBond.module.css
@@ -50,3 +50,24 @@
   color: 'white';
   border-radius: '50%';
 }
+
+.form {
+  padding: 0.5rem;
+}
+
+.formHeader {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  padding: 0.5rem 1rem 0 1rem;
+}
+
+.formHeader .title {
+  width: 100%;
+  font-size: 1.2rem;
+  color: var(--bs-body-color);
+}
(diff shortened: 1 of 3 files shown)


theborakompanioni:
@barrytra Nice!

Works exactly as expected. Great work! :rocket: 

- UI might need small adaptions to have the header consistent with other modals (renew bond, fee settings, seed modal, etc.):

## :camera_flash: 




i.e. by using:
```html
<rb.Modal show={isExpanded} animation={true} backdrop="static" centered={true} onHide={() => setIsExpanded(false)}>
<rb.Modal.Header closeButton>
  <rb.Modal.Title>{t('earn.fidelity_bond.create_fidelity_bond.title')}</rb.Modal.Title>
</rb.Modal.Header>
<rb.Modal.Body>
[...]
</rb.Modal.Body>
</rb.Modal>
```
See file `SpendFidelityBondModal.tsx` for examples! :pray: 

- I think the custom CSS styles are also not needed! :muscle: 

- Can we find a better name for variable `isExpanded`? e.g. `showCreateFidelityBondModal`

- The sprite does not need to change from "plus" to "caret-up" anymore, when `isExpanded` is true

What do you think?

barrytra:
Yes.. i agree to suggestions and they have been implemented.
My only concern to use customised CSS was because even in light mode the header appears to be dark.
Screenshot from 2024-05-22 15-01-35
But its okay if other modals have the same flow. 
please have a check

theborakompanioni:
> Yes.. i agree to suggestions and they have been implemented. My only concern to use customised CSS was because even in light mode the header appears to be dark. But its okay if other modals have the same flow. please have a check

Good thinking! Imho, it is okay and better to stay consistent. Header color in light mode can be changed in a different PR, if needed!

tACK. Thanks @barrytra :muscle:

theborakompanioni (approved):

amitx13 (approved):
yup!  It's working fine. Good job, @barrytra

barrytra:
Ahh!! I get it. It was actually placed according to dropdown. This should be fixed.
What do you think shall it be shown inside the modal itself?

barrytra:
Also can you suggest a way to produce this error??

PR #766, opened 2024-05-28
feat(orderbook): show fidelity bond value and locktime

### LINKED ISSUE #754: feat(orderbook): show fidelity bond value and locktime

The bond value currently shown in the orderbook is good for sorting the FB column.
However, it is also interesting to users to know how many coins it represents and how long they are locked up in the bond.

Show the `value` and `locktime` in the orderbook.
These can be shown in a popover or a modal.
Additionally, other infos can be also shown, e.g. `txid:vout`.
Would be nice if the locktime is displayed as "expires on YYYY-MM-DD (in X days/weeks/years)"

**Additional context**

Fidelity bond data is contained in the `orderbook.json` response:
e.g.
```shell
curl 'http://localhost:3000/obwatch/orderbook.json'
```
```json
{
  "offers": [
    {
      "counterparty": "J59wR9dzj6iJKbXw",
      "oid": 0,
      "ordertype": "sw0absoffer",
      "minsize": 27300,
      "maxsize": 124876516161,
      "txfee": 0,
      "cjfee": 250,
      "fidelity_bond_value": 0
    },
    {
      "counterparty": "J57GUX3hqXJ5kDyg",
      "oid": 0,
      "ordertype": "sw0reloffer",
      "minsize": 100001,
      "maxsize": 237376516161,
      "txfee": 0,
      "cjfee": "0.0003",
      "fidelity_bond_value": 31549804.289008386
    }
  ],
  "fidelitybonds": [
    {
      "counterparty": "J57GUX3hqXJ5kDyg",
      "utxo": {
        "txid": "fdcdb333efa66bb71d90528459800c495ce0b2d47d4a943e19c24734b9c5ca3c",
        "vout": 0
      },
      "bond_value": 31549804.289008386,
      "locktime": 1725148800,
      "amount": 123454197,
      "script": "00209c3c5e5d08acb494360e8ea3ae11bea4f7cc9c0e480916c79d7f7dc6e450804e",
      "utxo_confirmations": 13,
      "utxo_confirmation_timestamp": 1715173479,
      "utxo_pub": "02bde9f04df2b44b935e18365acfcafd2479aa3acf812a2a8408360473824f8d33",
      "cert_expiry": 1
    }
  ]
}
```

theborakompanioni:
Resolves #754.

Displays amount and expiration date of fidelity bonds in the orderbook overlay.

## How to test

Create a fidelity bond for one of the dev makers (e.g. on localhost:29080 - you need to stop the maker, create a fidelity bond, and restart the maker again). Then, reload the orderbook on the main dev instance and verify that the correct Fidelity Bond information is displayed.

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -87,8 +87,6 @@ export default function App() {
     [reloadCurrentWalletInfo],
   )
 
-  debugger
-
   const router = createBrowserRouter(
     createRoutesFromElements(
       <Route
--- a/src/components/Orderbook.tsx
+++ b/src/components/Orderbook.tsx
@@ -321,14 +321,14 @@ const offerToTableEntry = (offer: ObwatchApi.Offer, t: TFunction): OrderTableEnt
 interface OrderbookProps {
   entries: OrderTableEntry[]
   refresh: (signal: AbortSignal) => Promise<void>
+  isLoading: boolean
   nickname?: string
 }
 
-export function Orderbook({ entries, refresh, nickname }: OrderbookProps) {
+export function Orderbook({ entries, refresh, isLoading: isLoadingRefresh, nickname }: OrderbookProps) {
   const { t } = useTranslation()
   const settings = useSettings()
   const [search, setSearch] = useState('')
-  const [isLoadingRefresh, setIsLoadingRefresh] = useState(false)
   const [isHighlightOwnOffers, setIsHighlightOwnOffers] = useState(false)
   const [isPinToTopOwnOffers, setIsPinToTopOwnOffers] = useState(false)
   const [highlightedOrders, setHighlightedOrders] = useState<OrderTableEntry[]>([])
@@ -389,12 +389,9 @@ export function Orderbook({ entries, refresh, nickname }: OrderbookProps) {
             onClick={() => {
               if (isLoadingRefresh) return
 
-              setIsLoadingRefresh(true)
-
               const abortCtrl = new AbortController()
               refresh(abortCtrl.signal).finally(() => {
-                // as refreshing is fast most of the time, add a short delay to avoid flickering
-                setTimeout(() => setIsLoadingRefresh(false), 250)
+                console.log('Finished reloading orderbook.')
               })
             }}
           >
@@ -511,6 +508,7 @@ export function OrderbookOverlay({ nickname, show, onHide }: OrderbookOverlayPro
 
   const refresh = useCallback(
     (signal: AbortSignal) => {
+      setIsLoading(true)
       return ObwatchApi.refreshOrderbook({ signal, redirect: 'manual' })
         .then((res) => {
           if (!res.ok && res.type !== 'opaqueredirect') {
@@ -523,6 +521,7 @@ export function OrderbookOverlay({ nickname, show, onHide }: OrderbookOverlayPro
         .then((orderbook) => {
           if (signal.aborted) return
 
+          setIsLoading(false)
           setAlert(undefined)
           setOffers(orderbook.offers || [])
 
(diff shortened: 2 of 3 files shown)


src/components/App.tsx
```diff
@@ -87,8 +87,6 @@ export default function App() {
     [reloadCurrentWalletInfo],
   )
 
-  debugger
```
0xSaksham, line 90: Locally I have removed the debugger too. If it doesn't serve any purpose, we should remove it.

src/components/Orderbook.tsx
```diff
@@ -102,6 +103,10 @@ interface OrderTableEntry {
   bondValue: {
     value: number
     displayValue: string // example: "0" (no fb) or "114557102085.28133"
+    locktime?: number
```
0xSaksham, line 106: This part looks fine! Should amount also be explicitly converted to string?

0xSaksham (approved):
Altogether the changes look good. Definitely able to merge.

src/components/Orderbook.tsx
theborakompanioni: Hmm.. I think I know what you mean.. `amount` is passed as `String` to the `Balance` component. I think I'd rather want to have the type info here in the object rather than storing it as plain string here. Okay for you?

barrytra (commented):
It is somehow not working for me. 
I created a fidelity bond and then clicked start earning(which is start maker i believe). But for me it show bond value 0.
Screenshot from 2024-05-28 21-08-33
Screenshot from 2024-05-28 21-08-18

I am not sure if I am doing it the right way. please help!!
Though code part looks completely fine to me.

theborakompanioni:
> I am not sure if I am doing it the right way. please help!! Though code part looks completely fine to me.

It might take some time before the bond value appears - it worked for me after refreshing a couple of seconds later. Can you try it again? :pray:

theborakompanioni:
> > It might take some time before the bond value appears - it worked for me after refreshing a couple of seconds later. Can you try it again? 🙏
> 
> It got displayed but its value is 0.

Have you waited a couple of seconds and refreshed once or twice?

barrytra:
are you asking me to wait after creating FB but before start earning?

theborakompanioni:
> are you asking me to wait after creating FB but before start earning?

No, wait a little bit after the offers appears and the bond value is zero. Refresh the orderbook from time to time. The bond value should appear a bit delayed (your orderbook watcher actually contacts every maker–in this case "yourself"–in order to get the bond info).

theborakompanioni:
> ok i'll try that.. 

Thanks :pray: Please report back.

> What i understand is that value initially will be 0 and then it will increase.. right?

Zero means there is no bond (or not yet loaded). This is actually a good point.. do you think it would be better to show something else? e.g. _<empty>_ or a "no bond" label?

barrytra:
I refreshed it many times but still its value is still 0.
Don't know if i am doing something wrong?

>Zero means there is no bond (or not yet loaded). This is actually a good point.. do you think it would be better to show something else? e.g. or a "no bond" label?

we can do that but i think at the same time it can be confusing for the user as well. Maybe we can add some more info for user when value is 0(no bond) when he hovers over it.

barrytra:
Okay this works fine now.
There is still one issue left:
- This doesn't look much pleasing in light mode, that needs to be fixed.
Screenshot from 2024-05-28 23-09-32
We can probably fix it here or in some other PR

theborakompanioni:
> Okay this works fine now. 

Glad it worked out. Should have mentioned that the FB needs some confirmations :facepalm: Sorry :grimacing: 

> There is still one issue left:
> 
>     * This doesn't look much pleasing in light mode, that needs to be fixed.
>       We can probably fix it here or in some other PR

You are right. Balances in "light mode" on dark backgrounds (e.g. tooltips) are a problem. Needs to be addressed - I'd say it is okay to do in a follow-up PR. Okay for you?

theborakompanioni:
> yes we can do that in follow-up PR.

Agreed. Tracked in #770. Thanks @barrytra! :pray:

Issue #650, opened 2023-08-24
ui: redesign fidelity bond creation flow


theborakompanioni:
_Originally posted by @editwentyone in https://github.com/joinmarket-webui/jam/issues/372#issuecomment-1383009458_

I did a [redesign of the FB creation flow](https://www.figma.com/file/kfejZJFlwBywvLEnPEmJo1/JoinMarket-UI?node-id=5267%3A67391&t=jL3pT3ELgD9n83IM-11), that should allow nextnextnext madness, please have a look.

<img width="1457" alt="image" src="https://user-images.githubusercontent.com/96262473/212503764-929ad55b-c81b-4160-ad06-64091074f143.png">

also this flow includes a **warning**, if a user tries to use all available UTXOs (most of the time 1 of 1) for FB creation and expects that this locked up funds are participating in a collaborative tx (send/earn). [klick](https://www.figma.com/file/kfejZJFlwBywvLEnPEmJo1/JoinMarket-UI?node-id=5604%3A67014&t=jL3pT3ELgD9n83IM-11)
<img width="697" alt="image" src="https://user-images.githubusercontent.com/96262473/212503739-8e119680-5cac-4041-adf1-2dc6453db897.png">

PR #776, opened 2024-06-03
fix(fb): display error alert in modal

### LINKED ISSUE #775: (FB) : error Alert is displayed in the background of modal.
**Expected behavior**
In case of any error of problem creating new FB, it is better to display the alert message in the modal itself, so that it is easily readable for the user.

**Actual behavior**
Alert message is displayed behind  the modal which causes it to hide, and is not able to read the alert.
image

  1. Go to earn section.
  2. Click "Configure Fidelity bond"
  3. To produce error temporarily adapt the request code and throw an error [here](https://github.com/joinmarket-webui/jam/blob/407098dfb1a4b9a32647873c3a189b0cbbb898ca/src/components/fb/SpendFidelityBondModal.tsx#L294).
  4. Then proceed through steps to finally encounter this error.

barrytra:
fixes #775 
The alert message has been shifted inside modal.
However I was not able to produce the error locally, so its requested to produce the error locally while reviewing it. and please let know if it works fine.
PS: If anyone gets to know how to produce error, I would love to learn it as well.

### DIFF
--- a/src/components/fb/CreateFidelityBond.tsx
+++ b/src/components/fb/CreateFidelityBond.tsx
@@ -618,7 +618,6 @@ const CreateFidelityBond = ({ otherFidelityBondExists, wallet, walletInfo, onDon
 
   return (
     <div className={styles.container}>
-      {alert && <Alert {...alert} className="mt-0" onClose={() => setAlert(undefined)} />}
       {lockDate && timelockedAddress && selectedJar !== undefined && (
         <PaymentConfirmModal
           isShown={showConfirmInputsModal}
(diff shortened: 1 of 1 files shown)


theborakompanioni (approved):

theborakompanioni:
> PS: If anyone gets to know how to produce error, I would love to learn it as well.

e.g. this gets triggered when the wallet is not "Fidelity Bond"-enabled, when the wallet is created via the CLI and the users presses `n` on the question _Would you like this wallet to support fidelity bonds?_ (https://jamdocs.org/FAQ/#how-to-import-an-existing-wallet-via-the-command-line).

We could load the timelocked address immediately after selecting the date, so the error would get triggered earlier in the process. Currently the user selects a UTXO and only then the address is loaded.. maybe this can be considered in #650?

theborakompanioni (approved):

PR #800, opened 2024-07-19
fixed amount input field to properly interpret BTC

### LINKED ISSUE #799: Bug:  Fix the Behavior of Amount Input Field
**Expected behavior**
When 1.0 is entered in the `amount input field`, it should be considered 1 BTC.

**Actual behavior**

Instead of considering 1.0 as 1 BTC, it is considered 0.00000001 BTC

  1. Go to the Send Section (/send).
  2. Enter the amount as 1.0 in the amount input field.

amitx13:
This PR fixes #799 
Fixed the amount input field to correctly interpret `1.0` as `1 BTC`. Previously, it was incorrectly treated as a much smaller value; for example, entering `0.1` was considered `0.00000001 BTC`.

### DIFF
--- a/src/components/BitcoinAmountInput.tsx
+++ b/src/components/BitcoinAmountInput.tsx
@@ -61,6 +61,78 @@ const BitcoinAmountInput = forwardRef(
           : undefined
     }, [field, inputType])
 
+    const handleBlur = (e: React.FocusEvent<HTMLInputElement>) => {
+      setInputType({
+        type: 'text',
+        inputMode: 'decimal',
+      })
+
+      let displayValue = String(field.value?.value || '')
+      if (isValidNumber(field.value?.value)) {
+        displayValue = formatBtcDisplayValue(field.value!.value!)
+      }
+
+      form.setFieldValue(
+        field.name,
+        {
+          ...field.value,
+          displayValue,
+        },
+        false,
+      )
+      field.onBlur(e)
+    }
+
+    const handleChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+      const rawUserInputOrEmpty = e.target.value ?? ''
+      const validNumberRegex = /^-?\d*\.?\d*$/
+      if (!validNumberRegex.test(rawUserInputOrEmpty)) {
+        return
+      }
+      const floatValueOrNan = parseFloat(rawUserInputOrEmpty)
+      if (!isValidNumber(floatValueOrNan)) {
+        form.setFieldValue(
+          field.name,
+          {
+            ...field.value,
+            value: null,
+            userRawInputValue: e.target.value,
+            displayValue: e.target.value,
+          },
+          true,
+        )
+        return
+      } else {
+        const value: number = floatValueOrNan
+        let numberValues: string | undefined
+        const unit =
+          rawUserInputOrEmpty.includes('.') && parseFloat(rawUserInputOrEmpty)
+            ? unitFromValue(String(rawUserInputOrEmpty))
+            : unitFromValue(String(value))
+        if (unit === 'BTC') {
+          const splitted = String(value).split('.')
+          const [integerPart, fractionalPart = ''] = splitted
+          const paddedFractionalPart = fractionalPart.padEnd(8, '0').substring(0, 8)
+          numberValues = `${integerPart}${paddedFractionalPart}`
+        } else {
+          numberValues = value.toLocaleString('en-US', {
+            maximumFractionDigits: 0,
+            useGrouping: false,
+          })
+        }
+
+        form.setFieldValue(
+          field.name,
+          {
+            value: parseInt(numberValues, 10),
+            userRawInputValue: e.target.value,
+            displayValue: e.target.value,
+          },
+          true,
+        )
+      }
+    }
+
     return (
       <>
         <rb.InputGroup hasValidation={true}>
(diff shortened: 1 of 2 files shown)


0xSaksham (commented):
This is the most important thing mentioned by edi. It works great now!

editwentyone (dismissed):
fixed, thank you

src/components/BitcoinAmountInput.tsx
```diff
@@ -114,9 +114,9 @@ const BitcoinAmountInput = forwardRef(
               field.onBlur(e)
             }}
             onChange={(e) => {
-              const valueOrNan = parseFloat(e.target.value ?? '')
-
-              if (!isValidNumber(valueOrNan)) {
+              const valueOrNan = e.target.value ?? ''
```
theborakompanioni, line 117: Before, the var named `valueOrNan` was a "value" (number) or "NaN".. now it is the raw user input or empty, so the name should change as well, otherwise it is quite surprising.

src/components/BitcoinAmountInput.tsx
```diff
@@ -129,10 +129,12 @@ const BitcoinAmountInput = forwardRef(
                 )
                 return
               } else {
-                const value: number = valueOrNan
-
+                const value: number = floatValueOrNan
                 let numberValues: string | undefined
-                const unit = unitFromValue(String(value))
+                const unit =
+                  valueOrNan.includes('.') && parseFloat(valueOrNan)
```
theborakompanioni, line 135: Does this also work for languages where `.` is not the separator? e.g. afaik in italian it is `,` ..

theborakompanioni: If not, personally, I think it is not _that_ important. Internationalization support of number input field is quite cumbersome.

amitx13: > Does this also work for languages where `.` is not the separator? e.g. afaik in italian it is `,` ..

I may be wrong but BitcoinAmountInput only takes `numbers`, `e` & `.` as input, so i implemented on the basis of that
and also function `unitFromValue` identifies if the user input is in `sata` or `BTC` on the basis of `include('.')`

src/components/BitcoinAmountInput.tsx
amitx13: Renamed valueOrNan to rawUserInputOrEmpty for clarity

src/components/BitcoinAmountInput.tsx
theborakompanioni: > > Does this also work for languages where `.` is not the separator? e.g. afaik in italian it is `,` ..
> 
> I may be wrong but BitcoinAmountInput only takes `numbers`, `e` & `.` as inputs, so i implemented on the basis of that and also function `unitFromValue` identifies if the user raw input is in `sata` or `BTC` on the basis of `include('.')`

Yeah, it can be tricky to get it right across different browser implementations. If language settings are changed, Firefox lets me input `1,01` while Chromium does not.

amitx13: Oh, now I get it. I know it's not that important, but from a user's perspective, it might be seen as a bug in the input field.

amitx13: image
Hey, I tried what you suggested and guess what? It is indeed actually working. You were right, the `,` is converted into `.`. So Ig that solves our problem.

theborakompanioni: Would it be possible to write test cases to verify this behaviour? Or is the existing code too clunky for that?

amitx13:
Hey @theborakompanioni, if you have a moment, could you please take a look? I added a test case for `BitcoinAmountInput` and used a slightly different approach compared to what was previously used in the `tests`. The reason for this change is that using ```await user.type(inputElement, '1.0')``` was interpreting `1.0` as `1`, which ultimately resulted in `sats` being displayed. Instead, I used `fireEvent` to handle this scenario.

amitx13:
I've updated the `BitcoinAmountInput` component to only accept `.` as the decimal separator. I understand that restricting users to one separator may not be ideal, but this approach effectively solves the issues with `BitcoinAmountInput` and its related test cases.

src/components/BitcoinAmountInput.tsx
```diff
@@ -77,7 +149,6 @@ const BitcoinAmountInput = forwardRef(
             data-display-value={field.value?.displayValue}
             name={field.name}
             autoComplete="off"
-            type={inputType.type}
```
theborakompanioni, line 80: Can you let me know the reason for removing the `type` attribute?
e.g. this disabled up/down arrows for number inputs.

src/components/BitcoinAmountInput.tsx
```diff
@@ -61,6 +61,78 @@ const BitcoinAmountInput = forwardRef(
...
+      )
+      field.onBlur(e)
+    }
+
+    const handleChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+      const rawUserInputOrEmpty = e.target.value ?? ''
+      const validNumberRegex = /^-?\d*\.?\d*$/
+      if (!validNumberRegex.test(rawUserInputOrEmpty)) {
+        return
+      }
```
theborakompanioni, line 91: I am not sure about this regex test. Shouldn't the next line catch invalid number inputs (`!isValidNumber`)? What would be the impact of removing it?

src/components/BitcoinAmountInput.tsx
amitx13: I removed the `type` attribute due to inconsistencies in how different browsers handle it. For example, in Brave and Chrome, only numbers are allowed, while Firefox permits strings as well. Additionally, changing the HTML lang attribute to "IT" allows the use of ',' as a separator for Italian number formats.

Removing the `type` attribute helps ensure a more consistent input experience across different browsers.

src/components/BitcoinAmountInput.tsx
amitx13: The regex test was introduced because removing the `type` attribute allows users to enter any string. This check quickly identifies invalid numbers before further processing. The `isValidNumber` function is used to check if the entered value is a valid number, but it only accepts `number | undefined | null` as input. Since the value from e.target.value is a `string`, we use parseFloat to convert it.

However, parseFloat has issues, such as interpreting 1.0 as 1, which affects how we handle BTC vs. sats. The regex helps ensure that only valid numeric strings are passed to parseFloat, preventing such issues.

Removing the regex test would mean parseFloat could incorrectly process some inputs, and it would allow users to enter any string, leading to inconsistent behavior.

amitx13: I know it's quite cumbersome, but I think reverting it back will bring `up/down arrows`, but the problem of `1.0` or in `lang="IT"` `1,0` will persist.

editwentyone:
> I've updated the `BitcoinAmountInput` component to only accept `.` as the decimal separator. I understand that restricting users to one separator may not be ideal, but this approach effectively solves the issues with `BitcoinAmountInput` and its related test cases.

so will there be a solution in the future where we can differ between `,` and `.` as the separator to be l11n friendly?

amitx13:
> so will there be a solution in the future where we can differ between `,` and `.` as the separator to be l11n friendly?

Well in the English number format, JavaScript only accepts `.` as the decimal separator. To use `,` as the separator, we would need to change the number format to Italian. However, even if we change the number format to Italian, JavaScript will still interpret `,` as `.` for parsing purposes. The issue arises with inputs like `1.0` or `1,0` because JavaScript interprets it as `1`, which is correct generally but not in the context we need

editwentyone:
ok got it, lets keep it at `.` for the time being

theborakompanioni:
Okay, @editwentyone, as @amitx13 has argued, making it a text input only (as opposed to a number input) increases consistentcy between browsers, but lacks the ability to use browser-native up/down arrows.. I can see that this is a reasonable trade-off - what do you think?

editwentyone:
absolutely. good idea, I bet no one uses the arrows anyways!

amitx13:
> absolutely. good idea, I bet no one uses the arrows anyways!

Then, the only thing holding up the PR now are the reviews.

theborakompanioni (approved):

PR #802, opened 2024-07-22
follow-up(PR771): quick freeze/unfreeze

theborakompanioni:
Follow-up to #771.

TBD.

Still TODO:
- [x] Handle frozen/unfrozen expired/non-expired fidelity bonds correctly
- [x] Handle bond tags correctly (i.e. not using the translated display string for identification)
- [x] `ShowUtxos` should not use `ConfirmModal` (instead use an own/new distinct modal component?)
- [ ] Should `ShowUtxos` refactored to be external to `SourceJarSelector`?
- [ ] Rename `ShowUtxos` -> `QuickFreezeUtxoModal`?
- [ ] etc.

### DIFF
--- a/public/sprite.svg
+++ b/public/sprite.svg
@@ -364,7 +364,7 @@
     <symbol id="mixed" fill="none" viewBox="0 0 24 24" stroke="currentColor">
         <path stroke-linecap="round" stroke-linejoin="round" d="M4 16.8c6.708.082 6.496-10.932 15.96-9.4M16.57 4L20 7.44l-3.432 3.44M16.57 13.2l3.39 3.6-3.39 3.2M4 7.2c6.887-.124 6.381 11.044 15.56 9.6"/>
     </symbol>
-    <symbol id="Unmixed" fill="none" viewBox="0 0 24 24" stroke="currentColor">
+    <symbol id="unmixed" fill="none" viewBox="0 0 24 24" stroke="currentColor">
         <path stroke-linecap="round" stroke-linejoin="round" d="M18 17.5H5M15.602 21L19 17.5 15.602 14M18 6.5H5M15.602 10L19 6.5 15.602 3"></path>
     </symbol>
 </svg>
--- a/src/components/Balance.module.css
+++ b/src/components/Balance.module.css
@@ -18,7 +18,7 @@
   --jam-balance-deemphasize-color: #1153b5;
 }
 
-.balance {
+.balanceColor {
   color: var(--jam-balance-color);
 }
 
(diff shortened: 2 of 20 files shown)


amitx13 (commented):
Wow, the code looks way cleaner and has a precise approach! The improvements are really clear. I did notice a few UI-related aspects that might need some attention:

1. The checkbox is too close to the border.
2. The tags are positioned too close to the balance.
3. The cancel button is larger than the confirm button.
4. I also remember having a meeting with Edi where he mentioned that initially, only 5 UTXOs should be visible, and the rest should be scrollable.
I am nowhere near @editwentyone  when it comes to UI, so I think it might be helpful if @editwentyone could take a look at the modal.

Lastly, just a thought: Previously, users were not allowed to unfreeze the last UTXO and proceed with the transaction, but now they can. Considering this change, I came across two conclusions:

1. Who are we to restrict users from doing what they want? In the end, they are the ones responsible for their actions.
2. Most users learn to use software by just tinkering around rather than reading documentation. So, in my opinion, we should restrict users from doing that.
What do you think?

theborakompanioni:
Hey @amitx13! Appreciate the valuable feedback!

> Wow, the code looks way cleaner and has a precise approach! The improvements are really clear. I did notice a few UI-related aspects that might need some attention:
> 
>     1. The checkbox is too close to the border.

Changed when switching from `ConfirmModal` to an own modal. Did not give too much attention to UI details yet, but will definitely take care of it, once the code is "ready". Thanks!

> 
>     2. The tags are positioned too close to the balance.

:+1: 

>     3. The cancel button is larger than the confirm button.

Ha, how could I not notice that! Thanks :smiling_face_with_tear: 

>     4. I also remember having a meeting with Edi where he mentioned that initially, only 5 UTXOs should be visible, and the rest should be scrollable.

Yes. Good idea. I am still unsure how to make it more obvious that the list is scrollable and there are more UTXOs below..


> I am nowhere near @editwentyone when it comes to UI, so I think it might be helpful if @editwentyone could take a look at the modal.

Yes, feedback from @editwentyone is highly appreciated :pray: 
> 
> Lastly, just a thought: Previously, users were not allowed to unfreeze the last UTXO and proceed with the transaction, but now they can. Considering this change, I came across two conclusions:
> 
>     1. Maybe we shouldn’t restrict users from doing what they want, as they’re ultimately responsible for their actions.
> 
>     2. Most users tend to learn by tinkering around rather than reading documentation. So, in my opinion, we should restrict users from doing that.
>        What do you think?

Like you clear thinking on this one. But yes, this is intentional, however, open for discussion. Personally, I'd go for option 1.

theborakompanioni:
Ready for review.. it's not yet in the shape I would like it to be, but it's better than before and should be good enough to build on. Especially the tag handling should be improved and refactored in a follow-up PR.

amitx13 (approved):

0xSaksham:
image

Isn't there supposed to be a tick symbol here?

theborakompanioni:
> Isn't there supposed to be a tick symbol here?

Yes, you are right - this is done in a follow-up PR. Can you review and add feedback for #815?

> Also, if you want I can fix the UI issues here.

Any contribution is welcome! Best would be to focus on bringing your current PRs on `master` but any improvement is nice to have and if you see some defect, fixing them is always appreciated.

PR #805, opened 2024-07-30
Refactor/show utxos

barrytra:
In the show Utxos file, Instead of passing a single parameter ```isFrozen``` to ```toggle``` function, passing whole ```utxo``` data is better. As utxo's other data is need in fidelity bond section. 
I am not able to review the changes due to some backend issue :(


src/components/Send/ShowUtxos.tsx
```diff
@@ -295,22 +295,9 @@ const ShowUtxos = ({
...
-        setUnFrozenUtxos(utxos)
-      } else {
-        const utxos = frozenUtxos.map((utxo: Utxo, i: number) =>
-          i === utxoIndex ? { ...utxo, checked: !utxo.checked } : utxo,
-        )
-        setFrozenUtxos(utxos)
-      }
-    },
-    [frozenUtxos, unFrozenUtxos, setUnFrozenUtxos, setFrozenUtxos],
-  )
```
barrytra, line 313: I think shortening this piece of code would give the same results. Not sure though, please go through this.

src/components/Send/ShowUtxos.tsx
```diff
@@ -160,7 +160,7 @@ const UtxoRow = memo(
               type="checkbox"
               checked={checked}
               onChange={() => {
-                onToggle && onToggle(utxoIndex, isFrozen)
```
0xSaksham, line 163: what will happen to `isFrozen` that we're not getting from above, how would you handle frozen utxo and non-frozen utxo?

barrytra: ```utxo``` will contain all the properties of the particularly selected utxo. So to access ```isFrozen``` property we can use ```utxo.isFrozen```. Basically we are trying to access whole utxo instead of just one of its property ```isFrozen```.

theborakompanioni:
Would also be okay if  `utxo` is the only argument in the callback (removing `utxoIndex`). But can be done in #802.. what do you think @barrytra ?

theborakompanioni:
This PR is superseded by #815 and can be closed, right @barrytra ?

PR #806, opened 2024-08-04
feat: unit and visibility toggle on main wallet view

theborakompanioni:
Most wallets have the ability to switch the unit of account (BTC or sats) by clicking at the balance.

Before this commit, the balance on the main wallet view has been a visibility toggle only. If a user clicked the balance it either showed the balance in the preferred unit or hid the balance.
e.g. BTC -> hidden -> BTC -> hidden, etc.

After this commit, a click will change the unit of the balance and only with a subsequent click toggles the visibility.
e.g. BTC -> sats -> hidden -> BTC -> sats, etc.


editwentyone (approved):
I like, thanks

PR #807, opened 2024-08-08
UI : Show considered UTXOs before performing transaction

### LINKED ISSUE #773: (UI) Show "Selected UTXOs before performing transaction"
Subtask of: #689 
**Aim** : To give user an overview of the selected `UTXOs` before performing transaction 
image
**Objective**: Probably adding an extra field, i.e Selected UTXOs and showing the selected utxos to the user
@theborakompanioni and @editwentyone, please suggest if u have something else to make it more intuitive

amitx13:
This PR fixes #773 

Changes:
1. User are now able to review their `considered UTXOs` before performing a sweep
2. Successful implementation of 
image

### DIFF
--- a/src/components/PaymentConfirmModal.tsx
+++ b/src/components/PaymentConfirmModal.tsx
@@ -82,7 +82,6 @@ const ReviewConsideredUtxos = ({ utxos }: ReviewConsideredUtxosProps) => {
             onToggle={() => {
               // No-op since these UTXOs are only for review and are not selectable
             }}
-            showBackgroundColor={false}
           />
         </rb.Col>
       </rb.Collapse>
--- a/src/components/PaymentConfirmModal.tsx
+++ b/src/components/PaymentConfirmModal.tsx
@@ -1,4 +1,4 @@
-import { PropsWithChildren, useMemo } from 'react'
+import { PropsWithChildren, useMemo, useState } from 'react'
 import { Trans, useTranslation } from 'react-i18next'
 import * as rb from 'react-bootstrap'
 import Sprite from './Sprite'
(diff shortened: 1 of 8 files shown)


amitx13:
I requested the JM devs to review the `OpenAPI` Diff (https://github.com/JoinMarket-Org/joinmarket-clientserver/pull/1721) almost two weeks ago, but unfortunately, there has been no response. As a result, I went ahead and implemented the review of the `considered UTXOs` for the `sweep` transaction. Please take a look.

editwentyone:
strange, I can't review because I can't see this modal with the considered UTXOs anymore

amitx13:
> strange, I can't review because I can't see this modal with the considered UTXOs anymore

Hey, this PR needs to be rebased with TBK's follow-up. Also, just a heads-up—in the current scenario, it only works for sweep.

amitx13:
Q. Why introduced `disableCheckboxCell` and `customTheme`?

Well, because it wasn't aligning with what @editwentyone wanted, I added a `customTheme` parameter to give future developers more flexibility and ease of use.

As for `disableCheckboxCell`, previously, only the input checkbox was hidden, while the cell was still present all the time, which again wasn't aligning with the desired UI. It looked like this:

Screenshot from 2024-08-17 22-36-48
Instead of this:

image

editwentyone:
two questions: 

- why do we allow this feature only when sweeping? IMO it should be always on?
- it should be better to see the beginning and the end of the address, not two times "…" : like » "bc18da1 … 29x1a"

0xSaksham (approved):

amitx13:
> * why do we allow this feature only when sweeping? IMO it should be always on?

Hey we discussed this earlier in the meet where we came up with this common conclusion that showing the actual used UTXOs is only possible once https://github.com/JoinMarket-Org/joinmarket-clientserver/pull/1721 is merged and released. The UTXOs selected in the quick freeze/unfreeze are not actually selected for the transaction; only the JAR(mixdepth) is selected. After the successful merge of #1721, the transaction will be based on the selected UTXOs. which is why this feature is only enabled during a sweep. <br>
However, you previously suggested rephrasing `selected UTXOs` to `considered UTXOs` so we could potentially display those UTXOs in both direct send and coinjoin as well.

Thanks @editwentyone,

amitx13:
Done with the changes. @editwentyone  and @theborakompanioni , if you have a moment, please have a look.

theborakompanioni:
> Done with the changes. @editwentyone and @theborakompanioni , if you have a moment, please have a look.

Nice work.. overall I am unsure of the phrasing ("Considered UTXOs"), but that can be done in a follow-up PR. I have pushed small changes, care to take a look?

theborakompanioni (approved):

amitx13:
I was aiming for future compatibility with the updated directsend RPC-API (if it gets merged), but I agree we can go with this approach too. Thanks for the changes! If there's nothing else to add, I think we can merge this PR.

PR #820, opened 2024-08-21
fix: default for collaborators set to 1 and added dev tag

### LINKED ISSUE #780: dev: default for collaborators based on active maker count
As a developer new to the project, it is not obvious that one needs to adapt the default collaborator count (randomized between `8` - `10`) to `1` or `2` (as this is the active makers that are started when running `npm run regtest:init`). Otherwise the collaborative transaction will always fail.

It should be more obvious to know what to do, e.g. either by setting the default to be `1` or `2`, or by providing another visual hint in the UI when the app is started in dev mode (`npm run dev:start`).

nischal-shetty2:
Fixes #780

previously the default number of collaborators was randomised between 8-10, which would cause the collaboration transaction to fail. This PR fixes this issue by setting the default value of collaborations to 1 and adds a dev tag as a visual hint.

now if we directly send with 1 collaborator selected by default the transaction succeeds as expected.

https://github.com/user-attachments/assets/2bb9b413-eeb2-4047-9d93-1549a02bd52f


theborakompanioni:
> previously the default number of collaborators was randomised between 8-10, which would cause the collaboration transaction to fail. This PR fixes this issue by setting the default value of collaborations to 1 and adds a dev tag as a visual hint.

Hey @nischal-shetty2! Nice! Thanks for the contribution!

One additional change request is to only apply this when the app is in "dev mode" (as stated in the linked issue).
You can use `isDevMode` from `src/constants/debugFeatures.ts`! Would that be possible?

nischal-shetty2:
> One additional change request is to only apply this when the app is in "dev mode" (as stated in the linked issue). You can use `isDevMode` from `src/constants/debugFeatures.ts`! Would that be possible?

Sure i can do that right away!

nischal-shetty2:
> Hey @nischal-shetty2! Nice! Thanks for the contribution!

This is my first contribution to this repo, and even though it's a small one, I'm glad to start somewhere.
@theborakompanioni, are there any other changes needed?

src/components/Send/CollaboratorsSelector.tsx
```diff
@@ -78,7 +81,7 @@ const CollaboratorsSelector = ({
           max={maxNumCollaborators}
           isInvalid={usesCustomNumCollaborators && !isValidNumCollaborators(field.value, minNumCollaborators)}
           placeholder={t('send.input_num_collaborators_placeholder')}
-          value={customNumCollaboratorsInput || ''}
+          value={customNumCollaboratorsInput}
```
theborakompanioni, line 84:
```suggestion
          value={customNumCollaboratorsInput || ''}
```

theborakompanioni:
> > Hey @nischal-shetty2! Nice! Thanks for the contribution!
> 
> This is my first contribution to this repo, and even though it's a small one, I'm glad to start somewhere. @theborakompanioni, are there any other changes needed?

@nischal-shetty2 I do not think the "dev" tag is needed, because this feature is also available in production. Should it be removed? What do you think?

Also, can we use `undefined` (not `''` empty string) for `customNumCollaboratorsInput` when a user did not touch the input field yet?

src/components/Send/CollaboratorsSelector.tsx
nischal-shetty2: I removed it because customNumCollaboratorsInput will always be initialised:
```
const initialNumCollaboratorsInput = isDevMode() ? '1' : ''

const [customNumCollaboratorsInput, setCustomNumCollaboratorsInput] = useState<string>(initialNumCollaboratorsInput)
```
Hence, the fallback will be obsolete, right? Or should I add it back for some other reason / edge case?

nischal-shetty2: > Also, can we use `undefined` (not `''` empty string) for `customNumCollaboratorsInput` when a user did not touch the input field yet?

if we make `customNumCollaboratorsInput` to  `undefined` then we'll need the fall back you're right so ill add it again!

nischal-shetty2:
> I do not think the "dev" tag is needed, because this feature is also available in production. Should it be removed? What do you think?

@theborakompanioni yes, if this is the case in production as well then its definitely not required! I think it would be best not to show the dev tag
 
> Also, can we use `undefined` (not `''` empty string) for `customNumCollaboratorsInput` when a user did not touch the input field yet?

on it!

nischal-shetty2:
@theborakompanioni made the changes, could you review it now!

theborakompanioni:
> @theborakompanioni made the changes, could you review it now!

Made some small changes move the dependency on `isDevMode` from `CollaboratorsSelector` to the parent component. Other than that, looks good to me! :rocket: 

Can you review the changes and verify they work properly on your side as well? Start once with `npm run dev:start` (changes **should** apply) and once with `npm run start` (changes should **not** apply).

theborakompanioni (approved):

nischal-shetty2:
> Can you review the changes and verify they work properly on your side as well? Start once with `npm run dev:start` (changes **should** apply) and once with `npm run start` (changes should **not** apply).

tested it locally on both dev as well as prod and works perfectly! This can be merged now if everything is good!

@theborakompanioni although i noticed one more thing unrelated to this issue, on the ``Collaborator fees`` title
Screenshot 2024-08-28 at 8 17 20 PM
is it supposed to be '...' or does it have to show the value like how ``Number of collaborators`` title shows us current selected number

theborakompanioni:
> > Can you review the changes and verify they work properly on your side as well? Start once with `npm run dev:start` (changes **should** apply) and once with `npm run start` (changes should **not** apply).
> 
> tested it locally on both dev as well as prod and works perfectly! This can be merged now if everything is good!

:muscle: 

> 
> @theborakompanioni although i noticed one more thing unrelated to this issue, on the `Collaborator fees` title Screenshot 2024-08-28 at 8 17 20 PM is it supposed to be '...' or does it have to show the value like how `Number of collaborators` title shows us current selected number

This should change to something meaningful as soon as you input an amount!

nischal-shetty2:
> This should change to something meaningful as soon as you input an amount!

yes yes got it, my bad i didnt input the amount field

PR #835, opened 2024-09-05
ui:Added dropdown for mnemonic phrase input

### LINKED ISSUE #831: UI:Enhance Mnemonic Input Field with Recommended Words Suggestions
**Description:**

Enhance the user experience by providing recommended words suggestions for the mnemonic phrase input field. This will help users by offering suggestions based on the mnemonic standard, making the input process more intuitive and less error-prone.

**Suggested solution**

- [ ] dropdown feature within the mnemonic input field.
- [ ] As the user types, display recommended words from the official mnemonic word list (e.g., BIP-39).
- [ ] Allow users to select suggested words to complete their mnemonic phrase easily.

**Screenshot**
_Disclaimer: The screenshot is just for an overview. Any UI suggestions would be appreciated._

Screenshot from 2024-09-03 15-02-58

amitx13:
Fixes: #831 

Added recommended word suggestions to the mnemonic phrase input field. The feature aims to improve the usability and accuracy of the mnemonic input process

**To Test:**
Navigate to `Create new wallet ` or `Import existing wallet`.

### DIFF
--- a/src/components/MnemonicPhraseInput.tsx
+++ b/src/components/MnemonicPhraseInput.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useRef, useState } from 'react'
+import { useCallback, useEffect, useRef, useState } from 'react'
 import { Bip39MnemonicWordInput } from './MnemonicWordInput'
 import { MNEMONIC_WORDS } from '../constants/bip39words'
 import * as rb from 'react-bootstrap'
@@ -23,8 +23,8 @@ const MnemonicDropdown = forwardRef<HTMLDivElement, MnemonicDropdownProps>(({ sh
   <rb.Dropdown show={show}>
     <rb.Dropdown.Menu className={`table-responsive ${style.dropdownMenu}`} ref={ref}>
       {words.map((word) => (
-        <div className="m-1">
-          <rb.Dropdown.Item key={word} onClick={() => onSelect(word)} className="p-2">
+        <div className="m-1" key={word}>
+          <rb.Dropdown.Item onClick={() => onSelect(word)} className="p-2">
             {word}
           </rb.Dropdown.Item>
         </div>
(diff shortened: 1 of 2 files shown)


MarnixCroes (commented):
nice improvement

- after entering word 12, the word suggestion is stuck

image

- one nice addition might be to have enter keystroke select the suggestion when there is only one suggestion
- at recovery selecting the suggestion will not automatically navigate the entry point to the next entry, while it does this at wallet creation?

amitx13:
> * after entering word 12, the word suggestion is stuck

Hey @MarnixCroes, thanks for pointing that out—I appreciate your feedback. I will take a look into it.

> * one nice addition might be to have enter keystroke select the suggestion when there is only one suggestion

Sure, great suggestion! It would definitely be a nice addition.

> * at recovery selecting the suggestion will not automatically navigate the entry point to the next entry, while it does this at wallet creation?

Yes, I'm not sure, but I think it's because, at the time of wallet creation, JAM knows the mnemonic, but at the time of import, it doesn't. However, I will take a look and try to add this feature.

MarnixCroes (approved):
tACK 994f6ed3c8375e517cbf3f8ecaaaca60fde5f6ae
nice work

theborakompanioni:
> tACK [994f6ed](https://github.com/joinmarket-webui/jam/commit/994f6ed3c8375e517cbf3f8ecaaaca60fde5f6ae) nice work

Second that! 
tACK https://github.com/joinmarket-webui/jam/commit/994f6ed3c8375e517cbf3f8ecaaaca60fde5f6ae. 
So nice. :rocket:

src/components/MnemonicPhraseInput.tsx
```diff
@@ -18,17 +42,65 @@ export default function MnemonicPhraseInput({
...
+          const firstItem = dropdownRef.current.querySelector('.dropdown-item')
+          if (firstItem) {
+            ;(firstItem as HTMLElement).focus()
+          }
+        }
+      } else if (e.key === 'Enter') {
+        const matched = MNEMONIC_WORDS.filter((word) => word.startsWith(value))
+        if (matched.length === 1) {
+          handleInputChange(matched[0], wordIndex, true)
+          e.preventDefault()
```
theborakompanioni, line 98:
```suggestion
          e.preventDefault()
          handleInputChange(matched[0], wordIndex, true)
```

theborakompanioni:
> > * one nice addition might be to have enter keystroke select the suggestion when there is only one suggestion
> 
> Sure, great suggestion! It would definitely be a nice addition.

One additional thing that can be done is a follow-up PR is, that the dropdown is not focused when "tab" is hit, when the full word has been written by the user. Currently, you have to press "tab" twice to go to the next input element, ideally, only one button press is need in such situations!
(Dropdown should still be displayed so the user can see that it is part of the wordlist! It should just not focus and immediately jump to the next input element).
What do you think?

editwentyone (dismissed):
nice contribution, I tested it on safari macOS

src/components/MnemonicPhraseInput.tsx
amitx13: fixed in https://github.com/joinmarket-webui/jam/pull/835/commits/dde55148f5c9ab7de15b33be5e2977f9f9adce06

amitx13:
> One additional thing that can be done is a follow-up PR is, that the dropdown is not focused when "tab" is hit, when the full word has been written by the user. Currently, you have to press "tab" twice to go to the next input element, ideally, only one button press is need in such situations! (Dropdown should still be displayed so the user can see that it is part of the wordlist! It should just not focus and immediately jump to the next input element). What do you think?
> 
The dropdown remains visible but no longer gets focused when 'tab' is pressed after typing a full word. Now, pressing 'tab' moves directly to the next input element in these cases.

MarnixCroes (commented):
eb040fc9eaf990fff48cc1504acb7ba6076731bd

at recovery when you write the full entry, the suggestion keeps being displayed. Even after the cursor entry is onto the next field.
It goes away when you start typing.
Not a big issue but a bit odd

[Screencast from 10-09-2024 13:31:45.webm](https://github.com/user-attachments/assets/e1b936cb-4362-4913-8cc1-8d0de7335350)

edit: this is the same as https://github.com/joinmarket-webui/jam/pull/835#issuecomment-2335191056 afaict

amitx13:
> [eb040fc](https://github.com/joinmarket-webui/jam/commit/eb040fc9eaf990fff48cc1504acb7ba6076731bd)
> 
> at recovery when you write the full entry, the suggestion keeps being displayed. Even after the cursor entry is onto the next field. It goes away when you start typing. Not a big issue but a bit odd
> 

Yes, if you use `Tab` after writing a full word, it will show the dropdown so the user can see that it is part of the wordlist and navigate to the next input field.
Imho this is the right behaviour becz the purpose of `Tab` key is to navigate between focusable elements on a page, not for selecting options within a dropdown

> edit: this is the same as [#835 (comment)](https://github.com/joinmarket-webui/jam/pull/835#issuecomment-2335191056) afaict

Am I understanding this correctly, @theborakompanioni ?  Or is there something I’m missing?

theborakompanioni:
> > [eb040fc](https://github.com/joinmarket-webui/jam/commit/eb040fc9eaf990fff48cc1504acb7ba6076731bd)
> > at recovery when you write the full entry, the suggestion keeps being displayed. Even after the cursor entry is onto the next field. It goes away when you start typing. Not a big issue but a bit odd
> 
> Yes, if you use `Tab` after writing a full word, it will show the dropdown so the user can see that it is part of the wordlist and navigate to the next input field. Imho this is the right behaviour becz the purpose of `Tab` key is to navigate between focusable elements on a page, not for selecting option within a dropdown
> 
> > edit: this is the same as [#835 (comment)](https://github.com/joinmarket-webui/jam/pull/835#issuecomment-2335191056) afaict
> 
> Am I understanding this correctly, @theborakompanioni ? Or is there something I’m missing?

No, I think @MarnixCroes has a point here. It is good that `tab` jumps to the next input, but the dropdown should disappear. A user already aware of that it is in the worldlist if it is the only one displayed before `tab` is pressed. Makes sense?
Another thing: If there are multiple words to select, I need to press `tab` twice to select the first word. Can that be fixed?

MarnixCroes (approved):
27410d9aff85d1cf8c27be0e7042a18198c9224e
previous point is fixed https://github.com/joinmarket-webui/jam/pull/835#pullrequestreview-2292167630, this is smoother✅

>  Another thing: If there are multiple words to select, I need to press tab twice to select the first word. Can that be fixed?

agree that this could be an improvement as well. note that this is only when there is a scrollbar, when there are like just two suggestions then tab will go to the first suggestion at first press

amitx13:
> No, I think @MarnixCroes has a point here. It is good that `tab` jumps to the next input, but the dropdown should disappear. A user is already aware of that it is part of the worldlist if it is the only one displayed before `tab` is pressed. Makes sense? 

yes

>Another thing: If there are multiple words to select, I need to press `tab` twice to select the first word. Can that be fixed?

current behaviour: input (ja) -> dropdown appears -> pressed `Tab` once -> pressed Enter 
I'm really sorry, I'm not getting this in one go. Could you please confirm, @theborakompanioni if this is not the behavior you want?

[Screencast from 11-09-24 04:29:41 PM IST.webm](https://github.com/user-attachments/assets/1368369f-1ec4-439b-bbb6-940ba7c26004)

> Edit: Also, tests would be nice for this. e.g. what happens if a user inputs "a" (multiple words), "aban" (only one word) or "abandon" (completely written out - only one word left) and then test the `tab` and `tab` + `enter` behaviour.

Got it, but this can be done in a follow-up PR. As we discussed earlier, JAM is going to release a new version soon, so this would be a good addition. But if you need it now, I'm on it.

theborakompanioni:
> > No, I think @MarnixCroes has a point here. It is good that `tab` jumps to the next input, but the dropdown should disappear. A user is already aware of that it is part of the worldlist if it is the only one displayed before `tab` is pressed. Makes sense?
> 
> yes
> 
> > Another thing: If there are multiple words to select, I need to press `tab` twice to select the first word. Can that be fixed?
> 
> current behaviour: input (ja) -> dropdown appears -> pressed `Tab` once -> pressed Enter I'm really sorry, I'm not getting this in one go. Could you please confirm, @theborakompanioni and @MarnixCroes, if this is not the behavior you both want?

Hmm.. it works in Chromium but works differently in Firefox. Here the dropdown itself is in focus after the first tab. 



But it is not a deal-breaker and can be looked at in a follow-up PR.

> 
> > Edit: Also, tests would be nice for this. e.g. what happens if a user inputs "a" (multiple words), "aban" (only one word) or "abandon" (completely written out - only one word left) and then test the `tab` and `tab` + `enter` behaviour.
> 
> Got it, but this can be done in a follow-up PR. As we discussed earlier, JAM is going to release a new version soon, so this would be a good addition. But if you need it now, I'm on it.

Yeeah, that is what most of the time never happens when it comes to tests in a follow-up PR :wink: 
Ideally, you'll write them before or as you develop the feature. Afterwards is quite hard as you have not written the code with tests in mind. But I am always open for surprises : -)

MarnixCroes:
Merge as is and open an issue for the follow ups?

theborakompanioni:
> Merge as is and open an issue for the follow ups?

Okay with that. :raised_hands:

theborakompanioni:
@MarnixCroes Needs some small fixes. I am preparing a patch.

theborakompanioni:
Closing in favor of https://github.com/joinmarket-webui/jam/pull/1178

PR #838, opened 2024-09-07
Added rescan page translations and removed dev mode tag

### LINKED ISSUE #833: feat(rescan): enable rescanning chain
Rescanning the chain has been an experimental feature and is hidden by default.

Making it available would help non-technical users to not going through the steps outlined in https://github.com/joinmarket-webui/jamdocs/issues/47 and does away the need to use the CLI.

Steps to do:
- i18n support: translate strings in the page component
- remove the dev badge
- Keep "experimental" in the title?

nischal-shetty2:
Fixes #833

Added translations to the settings page rescan option and for the entire rescan page and also removed the dev tag and made the rescan feature available in production as well

### DIFF
--- a/src/i18n/locales/en/translation.json
+++ b/src/i18n/locales/en/translation.json
@@ -236,6 +236,7 @@
     "confirm_locking_modal_title": "Lock Wallet",
     "confirm_locking_modal_body_earn": "Earn is active. Locking the wallet will stop the service. You can close the browser window to let it run in the background.",
     "confirm_locking_modal_body_jam": "A collaborative transaction is being executed. Locking the wallet will stop the service. You can close the browser window to let it run in the background.",
+    "rescan_chain_option": "Rescan chain",
     "fees": {
       "title": "Fee Limits",
       "description": "Adjust mining fees and collaborator fees according to your needs. These settings will be reset to default values when the JoinMarket service restarts, e.g. on a system reboot. For more information, <1>see the documentation on fees</1>.",
@@ -269,6 +270,16 @@
       "error_loading_fee_config_failed": "Error while loading fee config values.",
       "error_saving_fee_config_failed": "Error while saving fee config values. Reason: {{ reason }}"
     }
+  },
+  "rescan_chain": {
+    "title": "Rescan timechain (Experimental)",
+    "subtitle": "Rescan the local timechain for wallet related transactions.",
+    "blockheight_label": "Blockheight",
+    "blockheight_info": "The height of the chain at which the rescan process is started.",
+    "error_rescanning_failed": "Error while starting the rescan process. Reason: {{ reason }}",
+    "valid_blockheight_error": "Please provide a valid blockheight value greater than or equal to {{ min }}.",
+    "submit_button": "Rescan timechain",
+    "submitting_button": "Rescaning timechain"
   },
   "logs": {
     "title": "Logs",
--- a/src/i18n/locales/en/translation.json
+++ b/src/i18n/locales/en/translation.json
@@ -236,7 +236,7 @@
     "confirm_locking_modal_title": "Lock Wallet",
     "confirm_locking_modal_body_earn": "Earn is active. Locking the wallet will stop the service. You can close the browser window to let it run in the background.",
     "confirm_locking_modal_body_jam": "A collaborative transaction is being executed. Locking the wallet will stop the service. You can close the browser window to let it run in the background.",
-    "rescan_chain_option": "Rescan chain",
+    "rescan_chain": "Rescan chain",
     "fees": {
       "title": "Fee Limits",
       "description": "Adjust mining fees and collaborator fees according to your needs. These settings will be reset to default values when the JoinMarket service restarts, e.g. on a system reboot. For more information, <1>see the documentation on fees</1>.",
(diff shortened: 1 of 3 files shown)


nischal-shetty2:
@theborakompanioni could you review this

theborakompanioni:
> @theborakompanioni could you review this

Nice!

Regarding the i18n keys: It is tried - although not 100% consistent - to have them prefixed with their intent, e.g. `"text_button_submit"`, `"text_button_submitting"` or "`label_<text>`" instead of `"submit_button"` or `"blockheight_label"`.
Okay for you to adapt that?

src/i18n/locales/en/translation.json
```diff
@@ -270,6 +271,16 @@
       "error_saving_fee_config_failed": "Error while saving fee config values. Reason: {{ reason }}"
     }
   },
+  "rescan_chain": {
+    "title": "Rescan timechain (Experimental)",
+    "subtitle": "Rescan the local timechain for wallet related transactions.",
+    "blockheight_label": "Blockheight",
```
theborakompanioni, line 277:
```suggestion
    "label_blockheight": "Rescan height",
```

src/i18n/locales/en/translation.json
```diff
@@ -270,6 +271,16 @@
...
+    "blockheight_info": "The height of the chain at which the rescan process is started.",
```
theborakompanioni, line 278:
```suggestion
    "description_blockheight": "The height of the chain at which the rescan process is started.",
```

src/i18n/locales/en/translation.json
```diff
@@ -270,6 +271,16 @@
...
+    "error_rescanning_failed": "Error while starting the rescan process. Reason: {{ reason }}",
+    "valid_blockheight_error": "Please provide a valid blockheight value greater than or equal to {{ min }}.",
```
theborakompanioni, line 280:
```suggestion
    "feedback_invalid_blockheight": "Please provide a valid value between {{ min }} and the current blockheight.",
```

src/i18n/locales/en/translation.json
```diff
@@ -270,6 +271,16 @@
...
+    "submit_button": "Rescan timechain",
```
theborakompanioni, line 281:
```suggestion
    "text_button_submit": "Rescan timechain",
```

src/i18n/locales/en/translation.json
```diff
@@ -270,6 +271,16 @@
...
+    "submitting_button": "Rescaning timechain"
```
theborakompanioni, line 282:
```suggestion
    "text_button_submitting": "Rescanning timechain..."
```

src/i18n/locales/en/translation.json
```diff
@@ -236,6 +236,7 @@
     "confirm_locking_modal_title": "Lock Wallet",
     "confirm_locking_modal_body_earn": "Earn is active. Locking the wallet will stop the service. You can close the browser window to let it run in the background.",
     "confirm_locking_modal_body_jam": "A collaborative transaction is being executed. Locking the wallet will stop the service. You can close the browser window to let it run in the background.",
+    "rescan_chain_option": "Rescan chain",
```
theborakompanioni, line 239:
```suggestion
    "rescan_chain": "Rescan chain",
```

nischal-shetty2:
@theborakompanioni these are the final objects
```
  "rescan_chain": {
    "title": "Rescan timechain (Experimental)",
    "subtitle": "Rescan the local timechain for wallet related transactions.",
    "label_blockheight": "Rescan height",
    "description_blockheight": "The height of the chain at which the rescan process is started.",
    "error_rescanning_failed": "Error while starting the rescan process. Reason: {{ reason }}",
    "feedback_invalid_blockheight": "Please provide a valid value between {{ min }} and the current blockheight.",
    "text_button_submit": "Rescan timechain",
    "text_button_submitting": "Rescanning timechain..."
  },
```
and 
```
settings: {
     "rescan_chain": "Rescan chain",
}
```

nischal-shetty2:
> Edit: Do you think we need a better/longer description text (subtitle)?

The current set of information is clear but this would definitely help the user. could you give any suggestions as to what should be added @theborakompanioni

theborakompanioni:
> > Edit: Do you think we need a better/longer description text (subtitle)?
> 
> The current set of information is clear but this would definitely help the user. could you give any suggestions as to what should be added @theborakompanioni

You are right, however, it is probably okay to do that in a follow-up PR, yes?
I think the changes here are already quite cool - so let's merge it.
After that we should think about disabling the form when a transaction is in progress, e.g. maker or taker running.
This has not been done properly as of now, as it was only available in dev mode. Also worth to do in a follow-up PR. Do you think so as well?

theborakompanioni (approved):

nischal-shetty2:
> You are right, however, it is probably okay to do that in a follow-up PR, yes? I think the changes here are already quite cool - so let's merge it. After that we should think about disabling the form when a transaction is in progress, e.g. maker or taker running. This has not been done properly as of now, as it was only available in dev mode. Also worth to do in a follow-up PR. Do you think so as well?

Yes i agree, we can consider adding better description later on after the rest of the things are sorted!

PR #840, opened 2024-09-11
dev: use `npm run dev` to run app in dev mode

theborakompanioni:
Small improvement for local development:
Changes the command for running in dev mode from `npm run dev:start` to `npm run dev` (e.g. as next.js does).

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,8 +47,10 @@
     "react-router-dom": "^6.26.1"
   },
   "scripts": {
-    "dev:start": "REACT_APP_JAM_DEV_MODE=true npm start",
-    "dev:start:secondary": "PORT=3001 JAM_BACKEND=jam-standalone JAM_API_PORT=29080 npm run dev:start",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -47,8 +47,10 @@
     "react-router-dom": "^6.26.1"
   },
   "scripts": {
-    "dev:start": "REACT_APP_JAM_DEV_MODE=true npm start",
-    "dev:start:secondary": "PORT=3001 JAM_BACKEND=jam-standalone JAM_API_PORT=29080 npm run dev:start",
+    "dev:start": "echo 'Deprecated command will be removed soon. Please use `npm run dev` instead.' && npm run dev",
```
MarnixCroes, line 50:
```suggestion
    "dev:start": "echo 'Deprecated command will be removed soon. Please use `npm run dev` instead.'",
```
this could be removed maybe?
if you don't do install first and just run npm run dev then the && in front will not work ofc..
feel free to ignore

MarnixCroes (approved):

PR #846, opened 2024-09-12
perf: optimize api request

theborakompanioni:
This PR attempts to prevent fetching UTXO labels when it is not strictly necessary.
The request to `/display` is quite expensive for heavily used wallets and users report it can take up to several minutes!

On initially unlocking a wallet the request is made at least once. After that, it will only be done when there is no label present for a UTXO returned in the `/utxo` endpoint response.

## How to test?
Do some transactions and verify that the new UTXOs all have the correct label attached. e.g. `deposit`, `cj-out`, etc.
`/display` requests previously always executed in the Jar Details Modal and the Send page.

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -67,14 +67,14 @@ export default function App() {
   }, [clearCurrentWallet])
 
   const reloadWalletInfo = useCallback(
-    (delay: Milliseconds) => {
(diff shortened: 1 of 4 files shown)


theborakompanioni:
ping @editwentyone

editwentyone (approved):
tested, works so far!

PR #848, opened 2024-09-12
fix(main): properly fade out jar tooltips

### LINKED ISSUE #839: wallet tab: look inside tooltip can be stuck
**Actual behavior**

At the wallet tab, when looking inside a jar, then exiting and hovering over the tooltip -> the tooltip gets stuck

  1. look inside a jar
  2. close the tab
  3. hover over the jar and specifically over the _look inside_ tooltip
  4 see that the tooltip is stuck at the jar

**Specifications**
  - Version: latest master
  - Platform: Ubuntu 
  - Browser: firefox
  
image

[Sc](https://github.com/user-attachments/assets/1967dc10-c40a-4650-99c4-e341ccc128d4)

theborakompanioni:
Fixes #839.

A side effect of this is, that it will not show the tooltip anymore if you focus the jars (e.g. by using the keyboard), which is not ideal but feels okay.

How to test? Follow the steps in  #839 and verify the tooltip now correctly disappears.

### DIFF
--- a/src/components/jars/Jar.tsx
+++ b/src/components/jars/Jar.tsx
@@ -209,12 +209,14 @@ const SelectableJar = ({
  */
 const OpenableJar = ({ tooltipText, onClick, ...jarProps }: OpenableJarProps) => {
   const [jarIsOpen, setJarIsOpen] = useState(false)
-  const onMouseOver = () => setJarIsOpen(true)
-  const onMouseOut = () => setJarIsOpen(false)
+  const openJar = () => setJarIsOpen(true)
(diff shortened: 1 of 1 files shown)


MarnixCroes (approved):
tACK 013aa34d83e2c2fe5b00eaff142222ed18890933

theborakompanioni:
There has been an update which seems to have fixed the tooltip behaviour. Switching back to the previous handling in 37edcd9c0570b84cc70f34517384546fa0d2b4db.

Issue #852, opened 2024-09-25
The Month dropdown select is listing the months in the wrong order when configuring a Fidelity Bond


pedromtcosta:
**Expected behavior**

The months should be listed in the correct order, from January to December,
Also, it is very likely that the `value` attribute of the `option` tag inside the `select` should be adjusted to reflect their order on the calendar, as currently it doesn't.

**Actual behavior**

December is listed as the first month, January second and so on until November. The values on the select options are also wrong, where December is `1`, January is `2` until November that is `12`. See the screenshots below:

![Screenshot 2024-09-24 at 22 40 57](https://github.com/user-attachments/assets/c8ae52e8-68cb-48b3-9632-d43577b029c4)
![Screenshot 2024-09-24 at 22 51 11](https://github.com/user-attachments/assets/b5da36a5-258f-481e-be58-67039473c005)

**Steps to reproduce the problem**
  1. Enter the `Earn` tab;
  2. Click on `Configure Fidelity Bond`;
  3. Click on the Month dropdown select to select the month;

**Specifications**
  - Version: 0.2.0
  - Platform: umbrelOS 1.2.2
  - Browser: Brave 1.69.168 Chromium: 128.0.6613.138 - macOS 14.3

**Additional context**

I haven't tried creating the fidelity bond to see if it will actually be created with the correct expiration date, but judging from the values of the `option` tags I would imagine not.

PR #853, opened 2024-09-25
Fix months order across timezones

pedromtcosta:
Fixes: #852 

The default `timeZone` for `Intl.DateTimeFormat` is based on the user's timezone, so because I'm behind UTC it get's shifted some hours back which leads to incorrect formatting of the month. Because we send the date in UTC to `Intl.DateTimeFormat` we can set UTC explicitly as the `timeZone` to solve the issue.

The test case `_selectableMonths / should display month name` was not passing on my end, and now it does.

Also fixed a supposedly typo on the `describe('_selectableMonth')` to reflect the correct name of the function that is being tested. I can adjust it back if you don't want unnecessary diff on the squashed commit.

### DIFF
--- a/src/components/fb/LockdateForm.test.tsx
+++ b/src/components/fb/LockdateForm.test.tsx
@@ -117,7 +117,7 @@ describe('<LockdateForm />', () => {
     })
   })
 
-  describe('_selectableMonth', () => {
+  describe('_selectableMonths', () => {
     const yearsRange = fb.toYearsRange(0, 2)
 
     const january2009 = new Date(Date.UTC(2009, 0))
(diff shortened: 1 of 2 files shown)


theborakompanioni:
Nice one! Thank you so much @pedromtcosta! 
High appreciated :pray:

theborakompanioni (approved):

pedromtcosta:
@theborakompanioni my pleasure! Not trying to rush anything, but do you have any idea when I can expect a new build including this fix on Umbrel OS?

theborakompanioni:
> @theborakompanioni my pleasure! Not trying to rush anything, but do you have any idea when I can expect a new build including this fix on Umbrel OS?

ETA in October. :crossed_fingers:

Issue #670, opened 2023-10-04
refactor: Use typed responses for api calls


theborakompanioni:
Use typed responses instead of generic `Response` type from `fetch`.

Currently, every component parses the body by itself (including error handling).
With this change, calling api endpoints should be much more streamlined and simplified.

e.g. from https://eckertalex.dev/blog/typescript-fetch-wrapper
```ts
async function request<T>(path: string, config: RequestInit): Promise<T> {
  const response = await fetch(new Request(path, config))
  
  if (!response.ok) {
               // TODO: use Api.Helper here instead generic Error
  throw new Error({ name: response.status, message: response.statusText })
  }
  
  // may error if there is no body, return empty
  return response.json().catch(() => ({}))
}
```

Akshanshkaushal:
@theborakompanioni I would be interested in working on this. Could you please assign it to me?

theborakompanioni:
> [@theborakompanioni](https://github.com/theborakompanioni) I would be interested in working on this. Could you please assign it to me?

I think this will be solved by generating the client from the openapi spec. So this is not something to be worked on as of now. I added the "blocked" label. 🙌

theborakompanioni:
Done with [joinmarket-api-ts](https://github.com/joinmarket-webui/joinmarket-api-ts).

PR #858, opened 2024-10-10
670 use typed responses for api calls

zmjohnso:
The goal of this PR is to begin addressing issue #670. As an example, I added a typed API response to `getWalletAll`. As the issue states, this should improve the developer experience by moving the body parsing into a single location and give some type completion when making this call in the future. If this still seems like a good direction to go in, I will start working on adding typed responses for the rest of the calls in `JmWalletApi.ts`.


theborakompanioni:
Hey @zmjohnso! Like the direction this is going.
Hope I have time to think about it a bit and provide feedback.

e.g. what do you think about generating the schema with `npx openapi-typescript https://joinmarket-org.github.io/joinmarket-clientserver/api/wallet-rpc.yaml -o ./src/libs/JmWalletApi.d.ts`?

zmjohnso:
> Hey @zmjohnso! Like the direction this is going. Hope I have time to think about it a bit and provide feedback.
> 
> e.g. what do you think about generating the schema with `npx openapi-typescript https://joinmarket-org.github.io/joinmarket-clientserver/api/wallet-rpc.yaml -o ./src/libs/JmWalletApi.d.ts`?

Oh, yes; thanks for sharing that! I think generating the types is usually better and should also save some time. I added the types in and I'm working on updating some more endpoints. I'll push some more code in the next few for you to take a peak.

theborakompanioni:
@zmjohnso How do you want to proceed? Should we include the generated files here before publishing it as a standalone lib for others to use as well?

zmjohnso:
@theborakompanioni Yes, I think just including the generated file here would be the easiest way to get the most value for Jam. But publishing a standalone lib is a neat idea; something that can be worked on in the future.

theborakompanioni:
@zmjohnso Generated api client is used in `v2` - will close the PR. Thanks for your efforts :pray:

Issue #872, opened 2024-11-25
Security feature: Show seed phrase only after entering password


oPFGKk9gtuw8nuHkzrQn:
### **Is your feature request related to a problem? Please describe.**

When logged into JAM, any user of that computer can easily access the seed phrase by simply selecting "Show seed phrase" as shown in the attached screenshot. This presents a significant security risk, as there is no additional authentication step required to view this sensitive information. This lack of a security barrier could lead to unauthorized access to the wallet if someone else gains temporary access to the browser window.
![grafik](https://github.com/user-attachments/assets/5c1e092b-aa0d-458d-8836-b76b8e140316)
![grafik](https://github.com/user-attachments/assets/940377fa-9c9f-4383-b972-283cbf3b6063)

### **Describe the solution you'd like**

Implement an additional layer of security for displaying the seed phrase. Specifically, when a user attempts to view the seed phrase, they should be prompted to enter their wallet password again. This step would ensure that only the legitimate wallet owner can view the seed phrase, adding an essential security checkpoint.

### **Describe alternatives you've considered**

 **Timed Access:** Allow the seed phrase to be visible for only a short period (e.g., 30 seconds) after successful authentication, after which the user would need to re-authenticate to view it again. Or lock the browser window after a certain amount of time.

### **Additional context**

- **Screenshot**: Attached screenshot shows the current "Show seed phrase" option without any security prompt.
- **User Impact**: Enhancing this feature would significantly improve the security posture of JAM, protecting users from potential theft or unauthorized access to their cryptocurrency wallets.

MarnixCroes:
well, if someone has access to your device they could also send all the money out.

khadar1020:
I would like to work on this @theborakompanioni can you assign me this issue

khadar1020:
https://github.com/user-attachments/assets/dd5b5db8-dfb2-47ed-932f-5a9c736ad667

This is the video @theborakompanioni I have solved this issue if any changes you want I will be doing other changes please tell me other wise I am ready to submit a PR 

Thank you

parrth20:
I would be interested in working on this. Could you please assign it to me?

theborakompanioni:
Hey @parrth20! This is also already worked on and has two PRs open: https://github.com/joinmarket-webui/jam/pull/874 and https://github.com/joinmarket-webui/jam/pull/881.

It seems https://github.com/joinmarket-webui/jam/pull/881is of higher quality and #874 should be closed.
[...]

PR #874, opened 2024-12-23
security_for_seed_phrase_request

khadar1020:
issue: #872 
I have already kept the video how it is going to work. 
@theborakompanioni can you please review my PR 

Thank you


khadar1020:
Hi @theborakompanioni can you please review my PR

khadar1020:
HI @theborakompanioni can you please review my PR

theborakompanioni:
> Hi, @theborakompanioni can you please review my PR

Sorry, I have not had the time yet. Still traveling and only partially able to code - will review as soon as I am back on my workstation.

theborakompanioni:
Hey @khadar1020!

Can we only ever have one modal shown, i.e. not overlapping? 
Also, I am unsure about the approach calling `postWalletUnlock`, when the wallet is already unlocked, since it returns `ApiAuthContext` and it probably need handling.

khadar1020:
@theborakompanioni is this ok or do you want to suggest any other approach 

0084BDBC-DE61-4747-9542-E9480A646C49

editwentyone:
lgtm. maybe also something like this above the input: 

Enter Wallet password to unlock
[Password]
Unlock (button)

nice

theborakompanioni:
> Ok when we click on the show seed phrase it will directly ask for wallet password and when the user enters the correct password then we should show them the seed phrase I think this is the best approach for ever showing the one model @theborakompanioni

Definitely better from a UI/UX perspective.
Can you comment on the statement and what your testing result were
> Also, I am unsure about the approach calling `postWalletUnlock`, when the wallet is already unlocked, since it returns `ApiAuthContext` and it probably need handling.
?

theborakompanioni:
Hey @khadar1020. Sorry I did not make myself clear enough. My question is regarding the return value of `postWalletUnlock`. It returns
```json
{
  "walletname": "wallet.jmdat",
  "token": "string",
  "refresh_token": "string"
}
```
and I want to know whether we run into problems if we keep using the token we got from the original unlock as it seems these values are not used afterwards. I have not checked myself and I want to know if you have verified. See https://joinmarket-org.github.io/joinmarket-clientserver/api/#operation/unlockwallet .

Your answer seems to be generated by an LLM.. it would be nice if you try to really address the question. Thank you :pray:

khadar1020:
I have verified that @theborakompanioni 
I am writing my final exams so because of time constraints just for formatting I have used LLM. 

Thank you

khadar1020:
I have verified that @theborakompanioni

theborakompanioni:
nack from my side for two reasons:
- Answers generated with an llm and not addressing the questions
- `token` and `refresh_token` token not used

Please don't take it personal. Reviewing takes time and effort. If something looks not functioning as intended, it would be nice if it is addressed with an explanation other than "I verified it works".

theborakompanioni:
@khadar1020 This feature is integrated in `v2` - will close the PR. Thanks for your efforts :pray:

PR #881, opened 2025-03-12
feat(settings): add secure seed phrase authentication

### LINKED ISSUE #872: Security feature: Show seed phrase only after entering password

rushil23451:
## What does this PR do?  
This PR enhances security by requiring users to enter their wallet password before displaying the seed phrase. This prevents unauthorized access in case someone gains temporary control over the user's browser.  

Additionally, the seed phrase is only visible for **30 seconds** before requiring the user to re-enter their password, adding an extra layer of security.  

Fixes #872 

Here is a video explaining the changes->

https://github.com/user-attachments/assets/e28e7359-c449-4783-92a4-f01c52335e78


theborakompanioni:
Nice @rushil23451!

Look promising! :rocket: 

Have you checked the flow mentioned in https://github.com/joinmarket-webui/jam/pull/874#issuecomment-2654721158? I see that the response data is also not used (https://github.com/joinmarket-webui/jam/pull/881/files#diff-a34cd1d40d43f9f5b929e00311e079de4b6b33c564cd01f0af5911e0a2fdd8bbR86).
It is:
```json
{
  "walletname": "wallet.jmdat",
  "token": "string",
  "refresh_token": "string"
}
```

I did not verify, just from a first glance at the code.

rushil23451:
Hey @theborakompanioni,
Just reviewed the issues you pointed out - let me explain:

1) About the response data - you mentioned it returns 'walletname', 'token' and 'refresh_token' that I'm not handling properly. I checked other parts of the codebase where the same API is called and found in Wallets.tsx they're doing:
'''
const res = await Api.postWalletUnlock({ walletFileName }, { password })
const body = await (res.ok ? res.json() : Api.Helper.throwError(res))
''' 

They're storing the response in a 'body' variable for later use. I could add this to my fix too, storing the returned response which could be used later if needed, but for my current code I don't actually need the returned data.

Does that address the issue? If you're satisfied, I'll submit another PR with this change.

2) For the build issues - I've fixed the unused imports and included the missing dependency. It now builds and compiles successfully when I run the build command locally.

rushil23451:
@theborakompanioni any feedback?

rushil23451:
@theborakompanioni 
I have now added these three lines while checking the password for seedphrases
'''
        const body = await res.json()
        const auth = Api.Helper.parseAuthProps(body)

        // Store both tokens in session
        setSession({
          walletFileName: wallet.walletFileName,
          auth: auth, // This includes both token and refresh_token
        })
'''

These code lines now parses the auth response as well as stores the returned tokens and refresh tokens in the session which now prevents the user from getting locked out after every 15minutes, Hope this now solves the issue.

theborakompanioni:
> @theborakompanioni I have now added these three lines while checking the password for seedphrases [...]

Nice! Thank you! Before I start testing, may I ask if you verified that it was a problem before and that it now works as expected?

theborakompanioni:
@rushil23451 This feature is integrated in `v2` - will close the PR. Thanks for your efforts :pray:

PR #882, opened 2025-03-16
chore: update regtest deps

theborakompanioni:
- Update ergochat to v2.15.0
- Update Bitcoin Core to v28


theborakompanioni:
@editwentyone Can you verify regtest setup is still working for you?

PR #894, opened 2025-04-27
added-rescan-progress-issue-744

### LINKED ISSUE #744: feat: show progress during rescan
See https://github.com/JoinMarket-Org/joinmarket-clientserver/pull/1696.
See also: https://github.com/joinmarket-webui/jam/issues/743

As a user, I want to have a progress on the rescan operation in order to estimate the time till a rescan completes.
Rescan can take quite some time depending on the hardware and it would be nice to have more information.

Instead of ["Rescanning in progress"](https://github.com/joinmarket-webui/jam/pull/621), show "Rescanning in progress (21%)"..

18c83fd3-25ea-4ed9-8205-2abeff9b3883:
Resolves #744.

cleaner fix


theborakompanioni:
Nice @18c83fd3-25ea-4ed9-8205-2abeff9b3883! Really looks cleaner than the first PR. Thanks!

Two things come to mind: 
- There needs to be a way to toggle this behaviour to only be executed for backend version >= v0.9.12 (not released yet). Or at least execute it once and then stop if it is unsupported e.g. if given a 404 error. 
- From my pov it seems 2s interval is quite frequent. The call will invoke `getblockchaininfo` on bitcoin core and might not need to be done this often. What do you think?

src/components/MainWalletView.tsx
```diff
@@ -61,7 +69,13 @@ const WalletHeaderRescanning = ({ walletName, isLoading }: { walletName: string;
           <rb.Placeholder className={styles.subtitlePlaceholder} />
         </rb.Placeholder>
       ) : (
-        <h2>{t('current_wallet.text_rescan_in_progress')}</h2>
+        <h2>
+          {serviceInfo?.rescanProgress !== undefined
+            ? t('current_wallet.text_rescan_in_progress_with_progress', {
+                progress: Math.round(serviceInfo.rescanProgress * 100),
```
theborakompanioni, line 75:
```suggestion
                progress: Math.floor(serviceInfo.rescanProgress * 100),
```

Should we `floor` here?

18c83fd3-25ea-4ed9-8205-2abeff9b3883: I didn't consider backwards compatibility. Will make the changes and resubmit

18c83fd3-25ea-4ed9-8205-2abeff9b3883:
Added a flag isRescanProgressApiSupported to track if the API is supported
Added logic to stop trying to fetch rescan progress.
Changed interval from 2s to 10s

theborakompanioni:
> Added a flag isRescanProgressApiSupported to track if the API is supported Added logic to stop trying to fetch rescan progress. Changed interval from 2s to 10s

:pray:  :rocket: 

What are those additional `fix_` and `update_` files?

18c83fd3-25ea-4ed9-8205-2abeff9b3883:
Utility scripts I made changes with and forgot to clean up after myself. DELETED Should be good to go now.

src/context/ServiceInfoContext.tsx
```diff
@@ -200,6 +208,60 @@ const ServiceInfoProvider = ({ children }: PropsWithChildren<{}>) => {
...
+        })
+
+        if (res.ok) {
+          const data = await res.json()
+          if (!abortCtrl.signal.aborted && data.progress !== undefined) {
+            dispatchServiceInfo({
+              rescanProgress: data.progress,
+            })
+          }
+        } else if (res.status === 404) {
```
theborakompanioni, line 237: actually got an "405 Method Not Allowed" on older version. Should we just skip the status check and do it for every `!ok`?
```suggestion
        } else {
```

False positives (e.g. temporary failures) would simply lead to the progress not being displayed, which is okay.

src/context/ServiceInfoContext.tsx
```diff
@@ -200,6 +208,60 @@ const ServiceInfoProvider = ({ children }: PropsWithChildren<{}>) => {
...
+          // If we get a 404 error, mark the API as not supported
+          if (err instanceof Api.JmApiError && err.response.status === 404) {
+            isRescanProgressApiSupported = false
+            console.log('Rescan progress API not supported by this backend version')
+          }
+        }
+      }
+    }
+
+    fetchRescanProgress()
```
theborakompanioni, line 254: We should also initialize with
```
    dispatchServiceInfo({
      rescanProgress: undefined,
    })
```
here?

Othwise it would briefly show the old progress, right? (not tested atm)

src/context/ServiceInfoContext.tsx
```diff
@@ -220,6 +220,11 @@ const ServiceInfoProvider = ({ children }: PropsWithChildren<{}>) => {
       // Skip if API is not supported (returned 404 previously)
       if (!isRescanProgressApiSupported) return
 
+      // Initialize rescanProgress to undefined to avoid briefly showing old progress
+      dispatchServiceInfo({
+        rescanProgress: undefined,
+      })
```
theborakompanioni, line 226: Not here actually.. but before you schedule with `setInterval`, no?

18c83fd3-25ea-4ed9-8205-2abeff9b3883:
improve error handling in the ServiceInfoContext.tsx file:

- Added initialization of rescanProgress to undefined to the beginning of the useEffect hook, before setting up the interval,

- Changed the error handling to catch all non-OK responses, not just 404 status. 

- Simplified the error handling in the catch block to treat all errors the same way, marking the API as not supported or temporarily unavailable.

src/context/ServiceInfoContext.tsx
```diff
@@ -59,6 +61,7 @@ interface JmSessionData {
   offer_list: Offer[] | null
   nickname: string | null
   rescanning: boolean
+  rescan_progress?: number
```
theborakompanioni, line 64: Last things before we can merge.. `/session` response does not contain a `rescan_progress` property. Please remove it.

(This causes the scanning text to flicker as the `/session` call updates it to `undefined` and the `getrescaninfo` update to the actual progress number)

Other than that, I think it is ready to be merged.

theborakompanioni:
> done. thanks for the feedback

How are you so freaking fast? Amazing.

theborakompanioni (approved):
Thanks @18c83fd3-25ea-4ed9-8205-2abeff9b3883.
Wonderful job! :pray: :orange_heart:

PR #898, opened 2025-05-02
fix(ui): fidelity bond utxo row color in dark mode

theborakompanioni:
Fixes the color of unfrozen and expired fidelity bond utxos in dark mode, e.g. in utxo selection and confirm modal.

(This should not have affected anybody if Jam has been exclusively used to control a JM node, as unfreezing Fidelity Bonds is not supported via the UI).


MarnixCroes (approved):
utACK

PR #904, opened 2025-06-05
feat: fast theme toggle in navbar

theborakompanioni:
The fast theme toggle button has been present in dev mode for a while now.
@editwentyone What do you think of making it available by default?

image
image


editwentyone:
sure, lgtm

PR #920, opened 2025-06-17
chore(v2): setup proxy and use query on login page

theborakompanioni:
Example of how to use tanstack query with the generated api client.
Please note that this is not yet fully functional and needs rework - e.g. auth token is not used for further requests.


nischal-shetty2 (approved):
lgtm🙏

PR #921, opened 2025-06-19
chore: simple token refresh

theborakompanioni:
Very basic and simple token refresh.
Currently, the interval will be triggered regardless of a wallet being unlocked.

### DIFF
--- a/src/hooks/useApiClient.ts
+++ b/src/hooks/useApiClient.ts
@@ -1,7 +1,5 @@
 import { createApiClient } from '@/lib/config'
-import { useMemo } from 'react'
 
-export const useApiClient = () => {
-  const client = useMemo(() => createApiClient(), [])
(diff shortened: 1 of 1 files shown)


src/hooks/useApiClient.ts
```diff
@@ -0,0 +1,7 @@
+import { createApiClient } from '@/lib/config'
+import { useMemo } from 'react'
+
+export const useApiClient = () => {
```
nischal-shetty2: We need to create a single instance of the client that is used throughout the application
```ts
let apiClient: ReturnType<typeof createApiClient> | null = null

export const useApiClient = () => {
  if (!apiClient) {
    apiClient = createApiClient()
  }
  return apiClient
}
```

nischal-shetty2 (approved):

PR #923, opened 2025-06-19
chore(build): add step to install Playwright browsers

nischal-shetty2:


theborakompanioni:
Can we cache the playwright deps?
e.g. https://playwrightsolutions.com/playwright-github-action-to-cache-the-browser-binaries/

Also.. in a follow-up PR we probably do not need to run them locally before every commit.. what do you think?

PR #925, opened 2025-06-19
build(deps): install @playwright/test

theborakompanioni:


nischal-shetty2:
@theborakompanioni How about completely removing playwright from the dependency? as we are not doing any visual tests as of now on the storybooks

theborakompanioni:
> @theborakompanioni How about completely removing playwright from the dependency? as we are not doing any visual tests as of now on the storybooks

I know what you mean. I think keeping it is fine for now.
However, we need the `@playwright/text` to make the cache work correctly in GH actions.

nischal-shetty2 (approved):

PR #929, opened 2025-06-20
chore(prettier): add prettierrc file and import sorting

nischal-shetty2:


theborakompanioni:
Nice work! :muscle: 

Should `generated` dir be excluded from formatting and linting? What do you think?

nischal-shetty2:
> Should `generated` dir be excluded from formatting and linting? What do you think?

Good point, even dist and other files can be explicitly ignored using `.prettierignore` file and its standard practice when having a `.prettierrc` file

Issue #933, opened 2025-06-23
v2: Settings page


theborakompanioni:
Users should be able to set all settings from the v1 app.
(e.g. lock wallet, switch theme, switch display mode, switch language, etc.)

Also:
- display seedphrase (Bonus: re-authenticate with the password, e.g. store it in the session storage hashed (10x) to be able to compare it)
- rescan chain
- link to community channels
- etc.

If something is done in a follow-up PR, please display the menu item as disabled with a tooltip that says "not implemented yet".

avinxshKD:
like to work on this, pls assign it to me

avinxshKD:
like to work on this, pls assign it to me if it is still available

theborakompanioni:
> like to work on this, pls assign it to me if it is still available

Hey @avinxshKD, `v2` branch is currently quite busy and some tasks are already being implemented. Need to tidy up a bit 😅 
You can work on a sub-task https://github.com/joinmarket-webui/jam/issues/944 if you want. Do dou want to be assigned?

theborakompanioni:
Completed in `v2`.

PR #937, opened 2025-06-28
add V2 settings page

nischal-shetty2:
Resolves: #933

### DIFF
--- a/src/components/settings/Settings.tsx
+++ b/src/components/settings/Settings.tsx
@@ -73,18 +73,6 @@ export const Settings = ({ walletFileName }: SettingProps) => {
 
   const toggleTheme = () => setTheme(resolvedTheme === 'dark' ? 'light' : 'dark')
 
-  const handleSwitchWallet = () => {
-    navigate('/switch-wallet')
-  }
-
-  const handleRescanChain = () => {
-    navigate('/settings/rescan')
-  }
-
-  const handleAdjustFeeLimits = () => {
-    setShowFeeLimitDialog(true)
-  }
-
   return (
     <div className="mx-auto max-w-2xl space-y-3 p-4">
       <div>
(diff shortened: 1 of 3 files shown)


src/components/settings/LanguageSelector.tsx
```diff
@@ -0,0 +1,43 @@
...
+  const handleLanguageChange = (languageKey: string) => {
+    i18n.changeLanguage(languageKey)
+  }
+
+  const getCurrentLanguageDescription = () => {
+    const currentLanguage = languages.find((lang) => lang.key === i18n.language)
+    return currentLanguage?.description || 'English'
+  }
+
+  return (
```
theborakompanioni, line 18: If a user never selected a wallet, it will be empty.
Should default to English.

image

src/components/settings/Settings.tsx
```diff
@@ -0,0 +1,231 @@
...
+
+      {/* Community Settings */}
+      <Card className="border-0 shadow-sm">
+        <CardHeader>
+          <CardTitle className="text-base font-medium">{t('settings.section_title_community')}</CardTitle>
+        </CardHeader>
+        <CardContent className="space-y-0">
+          <div
+            className="hover:bg-muted/50 -mx-2 flex cursor-pointer items-center justify-between rounded-md px-2 py-2"
+            onClick={() => window.open('https://matrix.to/#/%23jam:bitcoin.kyoto', '_blank')}
```
theborakompanioni, line 176: https://developer.mozilla.org/en-US/docs/Web/API/Window/open#noreferrer

"noreferrer noopener"!

src/components/settings/LanguageSelector.tsx
```diff
@@ -15,6 +15,11 @@ export const LanguageSelector = () => {
     return currentLanguage?.description || 'English'
   }
 
+  const getCurrentLanguageKey = () => {
+    const languageExists = languages.some((lang) => lang.key === i18n.language)
+    return languageExists ? i18n.language : 'en'
+  }
```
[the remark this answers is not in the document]
nischal-shetty2: Do you mean directly using `i18n.language` instead of the `getCurrentLanguageKey` function? the problem is that initially the browser detected language will be something like `en-IN` (in my case) which isn't a key in languages, so the language selector stays empty as seen here https://github.com/joinmarket-webui/jam/pull/937#discussion_r2175483240

src/hooks/useDisplayMode.ts
```diff
@@ -1,17 +1,41 @@
 import { useState } from 'react'
 
-export type DisplayMode = 'sats' | 'btc'
+export type DisplayMode = 'sats' | 'btc' | 'private'
```
theborakompanioni, line 3: Not sure if this is the best way to do it.
It seems to me this are two separate things.
One is the display currency format, sats or btc.
The other is private mode, on or off.
What do you think?

nischal-shetty2: Yes we can do it that way as well! Will the current implementation cause any issues or is it for separation of logic

theborakompanioni: No, as long as it is fixed later, I am fine with it.

PR #938, opened 2025-06-28
Feat: Add private display mode

nischal-shetty2:


theborakompanioni:
Can this setting be presisted to local storage?
Whenever the page is reloaded, the config is lost..

theborakompanioni (approved):

PR #940, opened 2025-06-30
Feat: setting seed display dialog

nischal-shetty2:


src/lib/hash.ts
```diff
@@ -0,0 +1,31 @@
+/**
+ * Securely hashes a password using PBKDF2 with SHA-256.
+ * @param password The password to hash.
+ * @param salt A salt value (should be unique per user).
+ * @param iterations Number of PBKDF2 iterations (default: 100,000).
+ * @returns The derived key as a hex string.
+ */
+export async function hashPassword(password: string, salt: string, iterations = 100000): Promise<string> {
+  if (typeof window === 'undefined' || !window.crypto?.subtle) {
```
theborakompanioni, line 9: Should we use https://github.com/paulmillr/noble-hashes for this. What do you think?

nischal-shetty2: I agree, much better to include extensively tested libraries than writing our own implementations

src/lib/hash.ts
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ * Securely hashes a password using PBKDF2 with SHA-256.
+ * @param password The password to hash.
+ * @param salt A salt value (should be unique per user).
+ * @param iterations Number of PBKDF2 iterations (default: 100,000).-
+ * @returns The derived key as a hex string.
+ * @throws {Error} If password hashing fails for any reason.
+ */
+export function hashPassword(password: string, salt: string, iterations = 100000): string {
```
theborakompanioni, line 12:
```suggestion
export function hashPassword(password: string, salt: string, iterations = 210_000): string {
```


https://github.com/paulmillr/noble-hashes/blob/main/src/pbkdf2.ts#L17 says
> c: iterations, should probably be higher than 100_000

And:
In 2023, [OWASP](https://en.wikipedia.org/wiki/OWASP) recommended to use 600,000 iterations for PBKDF2-HMAC-SHA256 and 210,000 for PBKDF2-HMAC-SHA512.[[6]](https://en.wikipedia.org/wiki/PBKDF2#cite_note-6)

nischal-shetty2: For the client side, it seriously affects performance.
I benchmarked this *(MacBook Air M1 chip)* The results are below (likely slower on weaker devices):

600,000 iterations:
`Password hashing completed in 3868.60ms`

100,000 iterations:
`Password hashing completed in 613.40ms`


[Bitwarden](https://community.bitwarden.com/t/increasing-the-number-of-iterations-of-pbkdf2-for-existing-accounts/49550?text=Bitwarden+currently+has+a+default+setting+of+100%2C001+iterations+client-side+with+an+additional+100%2C000+server-side%29%5D%28https%3A%2F%2Fcommunity.bitwarden.com%2Ft%2Fincreasing-the-number-of-iterations-of-pbkdf2-for-existing-accounts%2F49550%3Futm_source) uses 100,000 on the client side as well.
If the delay is acceptable, I'll switch to 600,000.

theborakompanioni (approved):

PR #942, opened 2025-07-05
feat: implement RescanChain component and integrate with settings page

nischal-shetty2:


src/hooks/useRescanStatus.ts
```diff
@@ -0,0 +1,49 @@
...
+export const useRescanStatus = () => {
+  const session = useSession()
+  const client = useApiClient()
+
+  const walletFileName = session?.walletFileName
+
+  const { data: rescanInfo, isLoading } = useQuery({
+    ...getrescaninfoOptions({
+      client,
+      path: { walletname: walletFileName || '' },
```
theborakompanioni, line 17: What happens when you use `useRescanStatus` when `walletFileName` is not present?

If the component needs this value to be present, it should be a required argument.
e.g. useRescanStatus( { walletFileName })

What do you think?

src/components/settings/RescanChain.tsx
```diff
@@ -0,0 +1,148 @@
...
+              <div className="bg-muted/50 mt-4 rounded-lg p-3">
+                <div className="flex items-center gap-2">
+                  <RefreshCw className="h-4 w-4 animate-spin" />
+                  <span className="text-sm">
+                    {rescanInfo?.progress ? `Rescanning... ${rescanInfo.progress}%` : 'Rescanning...'}
+                  </span>
+                </div>
+              </div>
+            )}
+          </div>
```
theborakompanioni, line 143: I cannot bring it to show me this message for more than a few milliseconds..
Even when no `getrescaninfo` request is made it disappears.

theborakompanioni: I dont know.
I think it is okay to activate the rescan state in the app and then periodically poll the api endpoint till it switches back to `rescanning := false`..

nischal-shetty2: Made some changes in the latest commit, is that what you meant?

nischal-shetty2: I have made few changes, Took few inspirations from the devel codebase as well and implemented using what understood from it

theborakompanioni: I think it still looks quite complicated.. but we can address it also later.

src/components/settings/RescanChain.tsx
```diff
@@ -0,0 +1,153 @@
...
+                    : t('rescan_chain.text_button_submit')}
+            </Button>
+
+            {(isRescanning || showCompletionMessage) && (
+              <div className="bg-muted/50 mt-4 rounded-lg p-3">
+                <div className="flex items-center gap-2">
+                  <RefreshCw className={`h-4 w-4 ${showCompletionMessage ? 'text-green-600' : 'animate-spin'}`} />
+                  <span className="text-sm">
+                    {showCompletionMessage
+                      ? 'Rescan completed successfully!'
```
theborakompanioni, line 140: Can we remove the completion message completely?

src/hooks/useRescanStatus.ts
```diff
@@ -0,0 +1,117 @@
...
+import { useApiClient } from './useApiClient'
+import { useSession } from './useSession'
+
+const RESCAN_PROGRESS_INTERVAL = 2000
+
+export const useRescanStatus = () => {
+  const session = useSession()
+  const client = useApiClient()
+  const completionTimeoutRef = useRef<NodeJS.Timeout | null>(null)
+  const [showCompletionMessage, setShowCompletionMessage] = useState(false)
```
theborakompanioni, line 13:
```suggestion
  const [showCompletionMessage, setShowCompletionMessage] = useState(false)
```

This is of no concern to this hook. It does not care what other components are doing. It should just provide the rescan data, no? What do you think?

nischal-shetty2: Yes if we remove the completion message then this entire logic wont be needed to begin with

PR #945, opened 2025-07-13
v2: Add cheatsheet to footer

nischal-shetty2:


src/lib/utils.ts
```diff
@@ -39,3 +39,23 @@ export const setIntervalDebounced = (
 export const satsToBtc = (value: string) => parseInt(value, 10) / 100000000
 
 export const btcToSats = (value: string) => Math.round(parseFloat(value) * 100000000)
+
+declare type SemVer = { major: number; minor: number; patch: number; raw?: string }
```
theborakompanioni, line 43: Wondering why `declare` is needed here. Can you explain?

theborakompanioni:
image

Nit: Scrollbar looks a bit odd with the rounded box - can only the enumerated content be in a scroll box?

src/lib/utils.ts
nischal-shetty2: its unnecessary, removed in the new commit

PR #947, opened 2025-07-18
chore: update jar balance

### LINKED ISSUE #911: After sending a transaction, detailed view of a jar still shows the old balance
**Expected behavior**

Jar's balance should reflect reality

**Actual behavior**

When sending a (non-collaborative) transaction out of a jar, the UTXO that was spent doesn't show up in the Jar's UTXO's, the jar balance in the overview is correct, but looking inside the jar givves incorrect balance. the spent TXO is also visible in jar > Details > Internal adresses

  1. spend one UTXO from a jar (freeze the others & sweep send non-collaborative transaction)
  2. observe jar's balance in Wallet distribution (correct) vs balance when looking inside the jar
  3.

**Specifications**
  - Version: 0.3.0
  - Platform: Start9
  - Browser: Chrome

**Additional context**

Image
Image

saurabhraghuvanshii:
fix #911 

- Fixed: the total displayed for a jar now sums the live UTXO values instead of using the possibly-stale jar.account_balance coming from /display.
- Passed String(utxosBalance) to the Balance component for the total line in the UTXO tab.

### DIFF
--- a/src/components/jar_details/JarDetailsOverlay.tsx
+++ b/src/components/jar_details/JarDetailsOverlay.tsx
@@ -141,6 +141,10 @@ const JarDetailsOverlay = (props: JarDetailsOverlayProps) => {
     return selectedUtxos.map((it) => it.value).reduce((acc, curr) => acc + curr, 0)
   }, [selectedUtxos])
 
+  const utxosBalance: Api.AmountSats = useMemo(() => {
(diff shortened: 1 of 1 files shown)


theborakompanioni (approved):
LGTM! Thank you @saurabhraghuvanshii :pray:

PR #949, opened 2025-07-19
fix: stop token refresh interval when no wallet is active

nischal-shetty2:


theborakompanioni:
Hmm.. shouldn't the token just be deleted when there is no active wallet?

nischal-shetty2:
>  shouldn't the token just be deleted when there is no active wallet?

Is the latest commits implementation correct? @theborakompanioni

theborakompanioni:
Let's just skip this as it is working at the moment and not mission critical. It should not distract you from the important things. Is that okay for you?

PR #963, opened 2025-08-12
feat: implement Orderbook component with pagination, sorting, and filter

nischal-shetty2:
Implementation of Orderbook component as a separate page instead of modal (which is the case in `v1`)

### DIFF
--- a/src/lib/utils.test.ts
+++ b/src/lib/utils.test.ts
@@ -11,6 +11,15 @@ import {
   factorToPercentage,
   toSemVer,
   UNKNOWN_VERSION,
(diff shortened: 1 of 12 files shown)


src/lib/utils.test.ts
```diff
@@ -261,3 +270,207 @@ describe('factorToPercentage', () => {
     expect(factorToPercentage(0.000027, 10)).toBe(0.0027)
   })
 })
+
+describe('humanReadableDuration', () => {
```
theborakompanioni, line 274: There is already a `humanReadableDuration` describe block below. Combine them?

PR #966, opened 2025-08-15
chore: enable i18n for storybook

theborakompanioni:
Enables i18n for stories and translates navbar menu items.


nischal-shetty2 (approved):
LGTM

PR #974, opened 2025-09-15
chore(v2): cleanup rescan, seed modal, fee config; add developer mode

theborakompanioni:
- add developer mode
- add jm config store
- cleanup rescan
- cleanup seed modal
- update joinmarket-api-ts client

Note: Fee dialog not 100% working correctly. Needs to be fixed in a follow-up PR.


nischal-shetty2 (approved):
LGTM

src/components/settings/RescanChain.tsx
```diff
@@ -114,31 +109,32 @@ export const RescanChain = ({ walletFileName }: RescanChainProps) => {
...
-            {isRescanning && (
+            {rescanInfo.rescanning && (
               <div className="bg-muted/50 mt-4 rounded-lg p-3">
                 <div className="flex items-center gap-2">
                   <RefreshCw className="h-4 w-4 animate-spin" />
                   <span className="text-sm">
-                    {rescanInfo?.progress ? `Rescanning... ${rescanInfo.progress}%` : 'Rescanning...'}
+                    {rescanInfo?.progress === undefined
+                      ? t('app.alert_rescan_in_progress')
+                      : t('alert_rescan_in_progress_with_progress', {
```
nischal-shetty2: hey, it's not translating here

Screenshot 2025-09-19 at 11 26 23 PM

src/components/settings/FeeLimitDialog.tsx
```diff
@@ -212,99 +173,104 @@ export const FeeLimitDialog = ({ open, onOpenChange, walletFileName }: FeeLimitD
...
 
           {/* Mining fees dropdown */}
           <div className="space-y-2">
-            <button
-              onClick={() => setMiningFeesExpanded(!miningFeesExpanded)}
-              className={`hover:bg-muted/50 flex w-full cursor-pointer items-center justify-between border-b px-4 py-2 text-left ${
-                miningFormRef.current && !miningFormRef.current.getFormData() ? 'border-red-300' : ''
-              }`}
+            <Accordion
+              type="single"
```
nischal-shetty2, line 235:
```suggestion
              type="single"
              collapsible
```

src/components/settings/FeeLimitDialog.tsx
```diff
@@ -212,99 +173,104 @@ export const FeeLimitDialog = ({ open, onOpenChange, walletFileName }: FeeLimitD
...
 
           <div className="space-y-2">
-            <button
-              onClick={() => setCollaboratorFeesExpanded(!collaboratorFeesExpanded)}
-              className={`hover:bg-muted/50 flex w-full cursor-pointer items-center justify-between border-b px-4 py-2 text-left ${
-                collaboratorFormRef.current && !collaboratorFormRef.current.getFormData() ? 'border-red-300' : ''
-              }`}
+            <Accordion
+              defaultValue="collaborator-fees"
+              type="single"
```
nischal-shetty2, line 196: The accordion isn't collapsing for both collaborator and mining fees form
the issue is Radix UI's Accordion component with type="single" doesn't allow collapsing by default. we need to explicitly set collapsible={true} (or just collapsible)

```suggestion
              type="single"
              collapsible
```

src/components/settings/RescanChain.tsx
theborakompanioni: Nice catch!

src/components/settings/FeeLimitDialog.tsx
theborakompanioni: Thanks for the careful review. This has been done deliberately (mentioned in the PR description as "Note: Fee dialog not 100% working correctly. Needs to be fixed in a follow-up PR.") - as the inner forms use ForwardRef and do not work correctly if placed within the Accordion (previously was a custom made component).
I want to revisit that at a later stage, I think there is a lot of work to do regarding forms, especially validation and input of amounts.

Okay for you if it stays this way for now?

src/components/settings/FeeLimitDialog.tsx
theborakompanioni: Same as above.

src/components/settings/FeeLimitDialog.tsx
nischal-shetty2: That sounds good! I thought it was a small fix so suggested it in this pr, but didn't know that it was deliberate. 

> Okay for you if it stays this way for now?

ofcourse!

Issue #954, opened 2025-07-30
separate private mode from DisplayMode


nischal-shetty2:
currently in [src/hooks/useDisplayMode.ts](https://github.com/joinmarket-webui/jam/pull/937/files/bafa7ba67816f8f3e07d5294de22eba6014bbd38..2f1b224c3603c3c81dacb74f8d80b94f6460afb3#diff-7721803c576843b119dbc4fa8bb73cfa4f67c80a5654e545c042f6812bc8d11d) the `private` mode is clubbed together with `sats` and `btc` mode. ideally this needs to be a separate type `private`/`hidden` and `public`/`visible`

theborakompanioni:
Already done.

PR #975, opened 2025-09-20
chore(v2): separate private mode from display mode

zmjohnso:
This PR addresses task #954.

The `src/hooks/useDisplayMode.ts` hook's  `DisplayMode` type was broken up into a `Currency` type and a simple `isPrivate` boolean. Doing so simplifies the logic where most of the time, we only care about toggling the currency type. This also greatly cleans up the logic in places like `src/components/settings/Settings.tsx`, where the separation of types is more apparent. While the individual toggles were separated, there is still a `toggleDisplayMode` for swapping between the three "display modes" on the landing page. Maybe there is still a cleaner way to do this? Also, the naming conventions could possibly still be improved. I used `currency: Currency`, but `bitcoinUnits: BitcoinUnits` might make more sense.


theborakompanioni:
LGTM

theborakompanioni (approved):

PR #985, opened 2025-11-13
refactor: use package joinmarket-api-ts instead of local client

theborakompanioni:
- use `@joinmarket-webui/joinmarket-api-ts`
- refactor: simplify navbar and symbols
- update dependencies
- pin node version


nischal-shetty2 (approved):
LGTM

PR #988, opened 2025-11-17
refactor: implemented switch wallet in login page and removed extra step 

### LINKED ISSUE #953: Implement switch wallet in login page and remove intermediate step
I am not sure of the current purpose of this component.

image
A click on "Lock Wallet" will redirect to the login page, right? Seems like an unnecessary in-between step.. what do you think?

_Originally posted by @theborakompanioni in https://github.com/joinmarket-webui/jam/issues/951#issuecomment-3124330104_

krishi-agrawal:
Resolves #953 

this PR:
1. implements switch wallet ui in login page
2. routes switch-wallet functionality to login page


theborakompanioni:
Hey @krishi-agrawal!

Thanks for the PR. I really like some of the changes and generally think the approach is fine.
Did you accidentally remove the "Create wallet" button? Also, I do not quite understand the purpose of the `AutoLockWrapper` component. I think we should generally put more effort into understanding what a good flow would look like that (a) a user expects and (b) fits JMs wallet model (e.g. only one wallet active at a time).

Is it okay for you if we take a little bit more time to think about what functionality the login page should provide?
e.g. one should see if a wallet is active and, if you have the auth token, have the ability to open it immediately. On the other hand, if you do not have the auth token (or an invalid one), have the ability to provide a password to login.
Nonetheless, one should be able to login to other (inactive) wallets as well. Moreover, if any service is active (maker or taker) it should warn the user that logging into a different wallet will terminate the operation.
So.. it is not so trivial after all..

krishi-agrawal:
Hi @theborakompanioni , thank you for the detailed review.  I did accidentally remove the create wallet button. I will revise my approach.

theborakompanioni:
Closing, please reopen if you think this is a mistake.

PR #998, opened 2025-12-29
Feat/display account xpubs

### LINKED ISSUE #620: Show xpubs for accounts

In my don't trust verify attitude, I wanted to understand what the flavors in JAM are and what they do and if the seed phrase and address derivation works as expected. I always do that for new wallets. It was more difficult here, because xpubs were not shown. I could pull it off in the end, but considerable effort was required.

Would be good to have a power user mode where users can display xpubs/zpubs for each flavor (account)

kunal-595:
fixes #620 

This PR implements a power user mode in the settings, allowing advanced users to display account extended public keys (xpubs/zpubs) for each wallet account.

Power User Mode Toggle:

Added "Enable power user mode" switch in Settings (Display section).
"Show account xpubs" menu item is only visible when power user mode is enabled.

How to Test
Run the app and open Settings.
Toggle "Enable power user mode" ON.
"Show account xpubs" menu item appears in Wallet Settings.
Click it, enter your wallet password, and verify xpubs are shown as zpub/vpub.
Test copy buttons and auto-timeout.


theborakompanioni:
@kunal-595 Thanks for the PR. Being able to import the keys as watch-only in other software is essential. However, I think there is a problem showing the xpub from the api response: afaik it exports the "wrong" values (see  https://github.com/JoinMarket-Org/joinmarket-clientserver/issues/1586) I did not re-check it myself but any change you might have you verified it yourself?

I think a good approach would be either: fix the values in the api response or calculate the values locally from the seed (there is already a dialog where a user needs to provide the passphrase to retrieve the seed - this could be done in a similar way).

What do you think?

kunal-595:
Thanks for the feedback! You're absolutely right about the API returning incorrect values .I verified this and saw it returns xpubs at the child derivation paths (like m/84'/1'/0'/0 for external addresses) instead of the account-level paths (m/84'/1'/0').

I have already implemented your suggested approach - calculating the values locally from the seed

theborakompanioni:
@kunal-595 This looks nice. Happy to merge this. Is it okay with you if I push a few adjustments before?

kunal-595:
Thanks! Yes, that’s totally fine please go ahead with the adjustments.
Also, if there are any other issues or tasks you think I could work on, I’d be happy to take them up.

theborakompanioni:
Thanks @kunal-595! Awesome contribution!
The network detection is currently fragile, but we can determine this once we have at least one address present (e.g. when `/display` request is made - which I will be working on soon™).
Also.. what do you think about exporting as qrcode? :pray: 

Again, thank you - really nice work! :pray:

kunal-595:
Hey @theborakompanioni,

Been watching the Slack chat lately and noticed a lot of people panicking about FB visibility when moving to Sparrow. It got me thinking maybe we can expand the QR/Display work to solve this?

I am thinking about a 'Recovery Info' modal that pulls the redeemScript and derivation path from jmwalletd. We could show it as a QR so people can just scan it into Sparrow/Electrum. It’s strictly metadata, not keys, so it should be safe and provide some real peace of mind.

Also, that blank page bug on Tor mobile is still a thing. I suspect our CSP headers or JS chunking might be the culprit there. I would like to audit that and maybe move to a more conservative build config for the frontend to make it more resilient.

Lastly, a small 'Network Settings' UI for directory nodes would be great so people donot have to keep digging into the .cfg files to improve their mix success.

Thinking of starting small with the FB export first. Does this align with the next release? Let me know what you think.

PR #1000, opened 2026-01-05
chore: add react-hook-form

theborakompanioni:
Adds react-hook-form and yup libraries for form hooks and schema validation.
Only applied on the Rescan page as an example, should be used through-out the app.


nischal-shetty2 (commented):
LGTM, sorry for the delay as I was a little busy earlier

PR #1003, opened 2026-01-09
chore(v2): earn and orderbook improvements

theborakompanioni:
- use selection and pinning functions from tanstack table
- improve sort icons in orderbook table
- use page title component consistently
- remove unneeded elements (e.g. loading states)
- externalize and move components to distinct files (e.g. OrderbookTable)
- rename page files (e.g. Settings -> SettingsPage)


theborakompanioni:
With these changes it becomes apparent that tanstack table re-renders more often than anticipated:
```
/home/void/workspace/github/joinmarket-webui/jam/src/components/orderbook/OrderbookTable.tsx
  233:17  warning  Compilation Skipped: Use of incompatible library

This API returns functions which cannot be memoized without leading to stale UI. To prevent this, by default React Compiler will skip memoizing this component/hook. However, you may see issues if values from this API are passed to other components/hooks that are memoized.

./jam/src/components/orderbook/OrderbookTable.tsx:233:17
  231 |   })
  232 |
> 233 |   const table = useReactTable({
      |                 ^^^^^^^^^^^^^ TanStack Table's `useReactTable()` API returns functions that cannot be memoized safely
  234 |     data: tableEntries,
  235 |     columns,
  236 |     state: {  react-hooks/incompatible-library

✖ 1 problem (0 errors, 1 warning)
```

This message was not logged before (i guess because the table has _always_ been re-rendered).
We can live with that for now, but needs to be revisited later on.

PR #1009, opened 2026-01-20
Add lock wallet confirmation modal and settings toggle

### LINKED ISSUE #944: Ask Confirmation before locking wallet
as discussed here: https://github.com/joinmarket-webui/jam/pull/939#discussion_r2175431257

kunal-595:
fixes #944 
Added a confirmation modal for locking the wallet to prevent accidental actions. Integrated the modal into the settings page and app navbar, with a toggle in settings to enable/disable it. .

### DIFF
--- a/src/components/layout/AppNavbar.tsx
+++ b/src/components/layout/AppNavbar.tsx
@@ -1,4 +1,4 @@
-import type { PropsWithChildren } from 'react'
+import { useState, type PropsWithChildren } from 'react'
 import type { SessionResponse } from '@joinmarket-webui/joinmarket-api-ts/jm'
 import type { TFunction } from 'i18next'
 import {
@@ -12,7 +12,9 @@ import {
 } from 'lucide-react'
 import { useTranslation } from 'react-i18next'
 import { Link, useNavigate, type NavigateFunction } from 'react-router-dom'
+import { useStore } from 'zustand'
 import { DevBadge } from '@/components/dev/DevBadge'
+import { LockWalletConfirmDialog } from '@/components/settings/LockWalletConfirmDialog'
 import { Button } from '@/components/ui/button'
 import { ThemeToggleButton } from '@/components/ui/jam/ThemeToggleButton'
 import { Skeleton } from '@/components/ui/skeleton'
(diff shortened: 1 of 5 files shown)


theborakompanioni:
Looks nice!

I think the confirmation dialog should only be shown when a service is active (earn, send, sweep) and does not need a dedicated setting. 
Also what about handling the dialog in `App` instead of `AppNavbar` and `Settings`?

I'd merge it nonetheless - these things can be done in follow-up PRs. What do you think?

kunal-595:
agreed, let's merge and iterate

you are right on both points:

Conditional dialog: The setting is unnecessary  we should just check [maker_running || coinjoin_in_process](vscode-file://vscode-app/c:/Users/Kunal%20Yadav/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) and show the dialog only when services are active simpler UX, less configuration.
centralize in App: Currently duplicated in [AppNavbar](vscode-file://vscode-app/c:/Users/Kunal%20Yadav/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) and [SettingsPage](vscode-file://vscode-app/c:/Users/Kunal%20Yadav/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) with identical logic moving to App.tsx elimnates that and makes it easier to trigger from anywhere

i will create follow-up issus for both:

remove the setting and make dialog conditional on service state
hoist dialog to App with a useLockWallet hook or context
Should be strightforward refactors without breaking existing behavior i can take these on unless you prefer to handle them

thanks for the review

theborakompanioni:
> agreed, let's merge and iterate

Done. :orange_heart: 

> remove the setting and make dialog conditional on service state hoist dialog to App with a useLockWallet hook or context Should be strightforward refactors without breaking existing behavior i can take these on unless you prefer to handle them

Yeah, I will handle it so you can focus on other stuff if you want :pray: 
Still so much to do for v2!


Edit: See https://github.com/joinmarket-webui/jam/pull/1011

PR #1016, opened 2026-01-22
Add log viewer overlay

kunal-595:
Adds a logs button in the footer that opens a full-screen overlay to view jmwalletd logs. Follows the same pattern as the orderbook overlay


theborakompanioni:
Nice. I'll move logs related components to their own dir, remove duplicate code and add a storybook story for the logs content. Thanks @kunal-595 :pray:

PR #1018, opened 2026-01-25
mobile logs access via floating action button

kunal-595:
#1016 
Adds floating action button on mobile devices for quick logs access during debugging , uses existing useIsMobile hook and overlay state management


theborakompanioni:
Hey @kunal-595! There is already a log button in the footer. I think that is enough. Logs don't need to be so prominently placed, as they are not available in all setups and only interesting if something fails, so _hopefully_ not necessary for the every-day normal user.

Okay for you to not include the changes?

PR #1020, opened 2026-01-29
Hide logs button when endpoint unavailable

kunal-595:
#1016 
hides the logs footer/mobile button when the feature endpoint fails (native mode), shows a clear "not suported" message in the overlay/page for dev testing


src/components/LogsOverlay.tsx
```diff
@@ -5,9 +5,14 @@ import PageTitle from '@/components/ui/jam/PageTitle'
...
-type LogsOverlayProps = WithRequiredProperty<Omit<ComponentProps<typeof Dialog>, 'children'>, 'open' | 'onOpenChange'>
+type LogsOverlayProps = WithRequiredProperty<
+  Omit<ComponentProps<typeof Dialog>, 'children'>,
+  'open' | 'onOpenChange'
+> & {
+  isSupported?: boolean
+}
 
-export function LogsOverlay({ open, onOpenChange }: LogsOverlayProps) {
+export function LogsOverlay({ open, onOpenChange, isSupported }: LogsOverlayProps) {
```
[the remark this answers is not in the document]
saurabhraghuvanshii: @kunal-595  can you address this?

saurabhraghuvanshii (approved):

src/components/LogsPage.tsx
```diff
@@ -179,6 +181,18 @@ export const LogsPage = () => {
     [t, authState],
   )
 
+  if (!isLogsSupported) {
```
theborakompanioni: What about just letting the error happen and show the normal error message?
I think it would be nice if the request was made, so we can debug and test the error scenario.

src/components/LogsContent.tsx
```diff
@@ -180,6 +181,17 @@ export const LogsContent = ({ enabled, className }: LogsContentProps) => {
     [t, authState],
   )
 
+  if (!isSupported) {
```
theborakompanioni: Same as below. Just let the error handling kick in

src/components/layout/Layout.tsx
```diff
@@ -89,12 +95,24 @@ export function LayoutInner({ onLogout, onLockWallet, children }: LayoutInnerPro
...
-        onClickLogs={() => setIsLogsOverlayOpen(true)}
+        onClickLogs={isLogsEnabled ? () => setIsLogsOverlayOpen(true) : undefined}
       />
 
       <Cheatsheet open={cheatsheet.open} onOpenChange={cheatsheet.onOpenChange} />
       <OrderbookOverlay open={isOrderbookOverlayOpen} onOpenChange={setIsOrderbookOverlayOpen} />
-      <LogsOverlay open={isLogsOverlayOpen} onOpenChange={setIsLogsOverlayOpen} />
+      <LogsOverlay open={isLogsOverlayOpen} onOpenChange={setIsLogsOverlayOpen} isSupported={isLogsSupported} />
+
+      {isMobile && isLogsEnabled && (
```
theborakompanioni: This reintroduces the button similar to https://github.com/joinmarket-webui/jam/pull/1018, right? Can it be removed or do you really, really want it in there? : D

I think the button in the footer is already enough. Logs should not be too prominent, imho. What do you think?

kunal-595: i will drop the extra button and keep access limited to the footer, that keeps logs available for debugging without making them too prominent

src/components/LogsPage.tsx
kunal-595: that makes sense, letting the request go through and showing the normal error would be useful for debuging and testing failure cases

theborakompanioni (approved):
lgtm! 🚀

PR #1023, opened 2026-02-01
Add fidelity bond creation flow

kunal-595:
fixes #650 
adds the missing wizard to create fidelity bonds from the Earn page, users can now select lockdate, pick UTXOs, and fund the timelock address with QR code display.

PR #1031, opened 2026-02-08
paginate and sort UTXOs by value

kunal-595:
#1023 
sorts UTXOs by value and adds pagination using the existing pagination components to keep the selection dialog manageable with larger wallets


theborakompanioni (approved):
💪

PR #1033, opened 2026-02-08
chore: wait for utxos to be spent and other small improvements

theborakompanioni:
A couple of improvements:
- manually reveal address on receive page (should prevent large gap limits)
- validate destination addresses (make sure it is not reused or from the same jar - if possible)
- freeze/unfreeze coins of reused addresses together
- copy addresses in jar details overlay
- wait for utxos to be spent on send page (still needs work to be resused by other components)
- store tx in session storage (can be used later to show a history during session)


kunal-595:
LGTM

PR #1035, opened 2026-02-09
Fix freeze/unfreeze toast messages

kunal-595:
#1023 
fixes typos in freeze/unfreeze toast messages by moving them to i18n with proper pluralization, and corrects a dead error branch caused by using filter() instead of every().

### DIFF
--- a/src/components/wallet/WalletJarsDetailsContent.tsx
+++ b/src/components/wallet/WalletJarsDetailsContent.tsx
@@ -132,50 +132,30 @@ export const UtxosContent = ({ enabled, walletFileName, addressSummary, jar }: U
   const onFreezeClick = async () => {
     try {
       const result = await freezeUtxos.mutateAsync(selectedUtxos)
-      const allFulfilled = result.filter((it) => it.status === 'fulfilled')
-      if (allFulfilled) {
-        // TODO: i18n
-        if (result.length === 1) {
-          toast.success('Selected UTXO has been frozen')
-        } else {
-          toast.success('Selected UTXOs have been frozed.')
-        }
+      const fulfilled = result.filter((it) => it.status === 'fulfilled')
+      const rejected = result.filter((it) => it.status === 'rejected')
+      if (rejected.length === 0) {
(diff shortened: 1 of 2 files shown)


src/components/wallet/WalletJarsDetailsContent.tsx
```diff
@@ -132,50 +132,28 @@ export const UtxosContent = ({ enabled, walletFileName, addressSummary, jar }: U
...
-      const allFulfilled = result.filter((it) => it.status === 'fulfilled')
+      const allFulfilled = result.every((it) => it.status === 'fulfilled')
       if (allFulfilled) {
-        // TODO: i18n
-        if (result.length === 1) {
-          toast.success('Selected UTXO has been unfrozen')
-        } else {
-          toast.success('Selected UTXOs have been unfrozed.')
-        }
+        toast.success(t('jar_details.utxo_list.toast_unfreeze_success', { count: result.length }))
```
saurabhraghuvanshii: @kunal-595 If I'm not wrong `count: result.length` return number of attempted UTXOs, not number of failed ones?

kunal-595: you are right result.length is the total attempted from Promise.allSettled().
I have updated it to use fulfilled.length for success and rejected.length for warnings so the counts reflect actual outcomes

saurabhraghuvanshii (approved):

saurabhraghuvanshii:
Thanks @kunal-595 lgtm!!

theborakompanioni:
Hey @saurabhraghuvanshii and @kunal-595! Appreciate the effort!
Just fyi: I think it is totally fine to focus on feature parity with the legacy version and then overhaul the whole i18n system and give it a new touch from the ground up. Some phrases are outdated, other can be improved. If you want, you can fix stuff on the fly, but it is also okay to just add `// TODO: i18n` and fix them later! Wording is important and worth spending more time on, but this should not slow you down too much while working on features. What do you think? :pray:

PR #1037, opened 2026-02-09
 extract hardcoded strings in CreateFidelityBondDialog to i18n

kunal-595:
#1023 
moves all hardcoded create-fidelity-bond wizard strings to i18n with proper pluraliztion support


theborakompanioni (approved):
lgtm

PR #1038, opened 2026-02-09
fix broken i18n keys in EarnPage and OfferCard

kunal-595:
#1023
fixes broken i18n keys in EarnPage and OfferCard by replacing hardcoded strings with proper translations


theborakompanioni:
It is okay to only change the english translation files!
As mentioned in another ticket, it is also okay to do it when feature parity is reached :pray:

PR #1039, opened 2026-02-09
fix(utils): truncate btcToSats beyond 8 decimals

parrth20:
### what this PR does : - 
- `btcToSats` now truncates beyond 8 decimals (no rounding up fractional sats).
- Updates unit tests (e.g. `0.000000005 BTC -> 0 sats`).

@theborakompanioni  @saurabhraghuvanshii

### DIFF
--- a/src/lib/utils.test.ts
+++ b/src/lib/utils.test.ts
@@ -158,6 +158,30 @@ describe('btcToSats', () => {
     expect(btcToSats('0.000000005')).toBe(0) // 0.5 sats truncates down to 0
     expect(btcToSats('0.00000003')).toBe(3) // Avoids floating point errors for exact-sat values
   })
+
+  it('should handle signs and whitespace', () => {
+    expect(btcToSats('+1')).toBe(100000000)
+    expect(btcToSats(' -1 ')).toBe(-100000000)
+    expect(btcToSats('-0.00000001')).toBe(-1)
+    expect(btcToSats('-0.000000005')).toBe(0) // Truncates towards 0 (no -0)
+  })
+
+  it('should handle exponent notation', () => {
+    expect(btcToSats('3e-8')).toBe(3)
+    expect(btcToSats('3E-8')).toBe(3)
+    expect(btcToSats('1e-8')).toBe(1)
+    expect(btcToSats('-3e-8')).toBe(-3)
+    expect(btcToSats('1e2')).toBe(10000000000) // 100 BTC
+  })
+
+  it('should return NaN for invalid values', () => {
+    expect(btcToSats('')).toBeNaN()
+    expect(btcToSats('   ')).toBeNaN()
+    expect(btcToSats('.')).toBeNaN()
+    expect(btcToSats('+')).toBeNaN()
+    expect(btcToSats('-')).toBeNaN()
+    expect(btcToSats('abc')).toBeNaN()
+  })
 })
 
 describe('parseSemanticVersion', () => {
--- a/src/lib/utils.test.ts
+++ b/src/lib/utils.test.ts
@@ -152,10 +152,11 @@ describe('btcToSats', () => {
     expect(btcToSats('0')).toBe(0)
   })
 
-  it('should handle rounding', () => {
+  it('should truncate beyond 8 decimals', () => {
     expect(btcToSats('0.123456789')).toBe(12345678) // Truncates beyond 8 decimals
     expect(btcToSats('0.000000004')).toBe(0) // Less than 0.5 sats truncates down
     expect(btcToSats('0.000000005')).toBe(0) // 0.5 sats truncates down to 0
+    expect(btcToSats('0.00000003')).toBe(3) // Avoids floating point errors for exact-sat values
   })
 })
 
--- a/src/lib/utils.test.ts
+++ b/src/lib/utils.test.ts
@@ -153,9 +153,9 @@ describe('btcToSats', () => {
   })
 
   it('should handle rounding', () => {
-    expect(btcToSats('0.123456789')).toBe(12345679) // Rounds to nearest integer
-    expect(btcToSats('0.000000004')).toBe(0) // Less than 0.5 sats rounds down
-    expect(btcToSats('0.000000005')).toBe(1) // TODO fix this 0.5 sats should rounds down to 0
+    expect(btcToSats('0.123456789')).toBe(12345678) // Truncates beyond 8 decimals
+    expect(btcToSats('0.000000004')).toBe(0) // Less than 0.5 sats truncates down
+    expect(btcToSats('0.000000005')).toBe(0) // 0.5 sats truncates down to 0
   })
 })
 
--- a/src/lib/utils.test.ts
+++ b/src/lib/utils.test.ts
@@ -5,7 +5,7 @@ import {
   walletDisplayName,
   setIntervalDebounced,
   satsToBtc,
-  btcToSats,
+  tryBtcToSat,
   percentageToFactor,
   isValidNumber,
   factorToPercentage,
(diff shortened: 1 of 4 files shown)


saurabhraghuvanshii (approved):

saurabhraghuvanshii:
@parrth20  LGTM!! wait for others to review

parrth20:
> @parrth20 LGTM!! wait for others to review

ping @theborakompanioni

theborakompanioni (changes requested):
nack. I am not in convinced. Regex is expensive and seems overkill for what can be just `Math.floor(Number.parseFloat(value) * 100_000_000)`, imho.

parrth20:
> nack. I am not in convinced. Regex is expensive and seems overkill for what can be just `Math.floor(Number.parseFloat(value) * 100_000_000)`, imho.



> nack. I am not in convinced. Regex is expensive and seems overkill for what can be just `Math.floor(Number.parseFloat(value) * 100_000_000)`, imho.

Good point on regex cost...I pushed an update removing it. btcToSats now uses simple string parsing + digit-scan (no regex) to build sats, which avoids floating point issues that make Math.floor(parseFloat(value) * 1e8) incorrect for some inputs (e.g. 0.00000003 can become 2.9999999999999996 → 2). Added a unit test for 0.00000003 -> 3 to guard this case. Let me know if you’d prefer dropping BigInt too; we can switch to parseInt since sats fit in Number.MAX_SAFE_INTEGER

src/lib/utils.ts
```diff
@@ -90,7 +90,51 @@ export function setIntervalDebounced(
...
+      const wholePartRaw = dotIndex === -1 ? numericPart : numericPart.slice(0, dotIndex)
+      const fractionalPartRaw = dotIndex === -1 ? '' : numericPart.slice(dotIndex + 1)
+      const wholePartDigits = wholePartRaw === '' ? '0' : wholePartRaw
+
+      if (isAsciiDigits(wholePartDigits) && isAsciiDigits(fractionalPartRaw)) {
+        // Truncate beyond 8 decimals (no rounding up of fractional sats).
+        const fractionalPartPadded = (fractionalPartRaw + '00000000').slice(0, 8)
+        const sats = BigInt(wholePartDigits) * 100000000n + BigInt(fractionalPartPadded || '0')
+        const result = sign * Number(sats)
+        return result === 0 ? 0 : result
```
theborakompanioni:
```suggestion
        return sign * Number(sats)
```

src/lib/utils.ts
```diff
@@ -90,7 +90,51 @@ export function setIntervalDebounced(
 
 export const satsToBtc = (value: string) => Number.parseInt(value, 10) / 100_000_000
 
-export const btcToSats = (value: string) => Math.round(Number.parseFloat(value) * 100_000_000)
+const isAsciiDigits = (value: string) => {
```
theborakompanioni, line 93: empty string leads to true?

src/lib/utils.ts
```diff
@@ -90,7 +90,51 @@ export function setIntervalDebounced(
...
+    sign = -1
+    numericPart = numericPart.slice(1)
+  } else if (numericPart.startsWith('+')) {
+    numericPart = numericPart.slice(1)
+  }
+
+  // Fast path for plain decimal notation: do a string-based conversion to avoid floating point issues.
+  if (!numericPart.includes('e') && !numericPart.includes('E')) {
+    const dotIndex = numericPart.indexOf('.')
+    const hasOnlyOneDot = dotIndex === -1 || !numericPart.includes('.', dotIndex + 1)
```
theborakompanioni: `hasOnlyOneDot` is true if there is no dot in the string?

src/lib/utils.test.ts
```diff
@@ -152,10 +152,11 @@ describe('btcToSats', () => {
     expect(btcToSats('0')).toBe(0)
   })
 
-  it('should handle rounding', () => {
-    expect(btcToSats('0.123456789')).toBe(12345679) // Rounds to nearest integer
-    expect(btcToSats('0.000000004')).toBe(0) // Less than 0.5 sats rounds down
-    expect(btcToSats('0.000000005')).toBe(1) // TODO fix this 0.5 sats should rounds down to 0
+  it('should truncate beyond 8 decimals', () => {
```
theborakompanioni, line 155: Can you add test cases for the other excution paths as well?
e.g. prefix with `+` or `-`, something resulting in `NaN`, `e` or `E` included, some invalid values you are catching, etc?

theborakompanioni:
> Good point on regex cost...I pushed an update removing it. btcToSats now uses simple string parsing + digit-scan (no regex) to build sats, which avoids floating point issues that make Math.floor(parseFloat(value) * 1e8) incorrect for some inputs (e.g. 0.00000003 can become 2.9999999999999996 → 2). Added a unit test for 0.00000003 -> 3 to guard this case. Let me know if you’d prefer dropping BigInt too; we can switch to parseInt since sats fit in Number.MAX_SAFE_INTEGER

Acknowledged, thanks for pointing it out. If BigInt is not necessary, consider dropping it.
I am not too happy about to `NaN` as result, but I guess that was in there before since we are just try to parse the value from a string. I think it is safe, as we should only ever user intergers in sats when doing calculations and converting from btc is only for displaying values. Maybe the function name and return type should reflect that by adding `AsText` and using a string as result instead of a number - but that should be considered in a follow-up PR, not here.

If you'll add additional tests for all execution path, it is fine with me :pray:

parrth20:
> Acknowledged, thanks for pointing it out. If BigInt is not necessary, consider dropping it. I am not too happy about to `NaN` as result, but I guess that was in there before since we are just try to parse the value from a string. I think it is safe, as we should only ever user intergers in sats when doing calculations and converting from btc is only for displaying values. Maybe the function name and return type should reflect that by adding `AsText` and using a string as result instead of a number - but that should be considered in a follow-up PR, not here.
> 
> If you'll add additional tests for all execution path, it is fine with me 🙏

thanks....addressed all review points in a follow-up commit:

isAsciiDigits('') no longer returns true (empty string is rejected; fractional part explicitly allows empty).
renamed the dot check to hasAtMostOneDot (it’s intentionally true when there’s no dot...plain integer strings are valid inputs).
Added unit tests covering all execution paths: +/- prefixes + whitespace, invalid strings → NaN, and e/E notation (plus a regression for 0.00000003 -> 3 to avoid float precision issues).
I kept a tiny sats === 0n ? 0 : … guard to avoid returning -0 (Vitest uses [Object.is](app://-/index.html#), so -0 would fail toBe(0)); happy to revisit dropping BigInt return-type changes in a follow-up.

parrth20:
>  but that should be considered in a follow-up PR, not here.

thanks, makes sense to keep #1039 scoped. for the follow-up:- would you prefer a narrow PR that only introduces a new explicit conversion API (e.g. \btcToSatsAsText` / stricter type semantics) while keeping `btcToSats` for compatibility, or a broader cleanup of all BTC<->sats conversion call sites in one go?

theborakompanioni:
> > but that should be considered in a follow-up PR, not here.
> 
> thanks, makes sense to keep #1039 scoped. for the follow-up:- would you prefer a narrow PR that only introduces a new explicit conversion API (e.g. \btcToSatsAsText`/ stricter type semantics) while keeping`btcToSats` for compatibility, or a broader cleanup of all BTC<->sats conversion call sites in one go?

I'd prefer to do it in one go, since we started v2 on a clean state and `btcToSats` is from legacy and backward compatibility is not an issue. But please do as you see fit!

theborakompanioni:
Thinking about it again, I have one more plea: Code changed a bit and is quite "complicated" now and I don't think we should ever need to parse scientific notation. Should we just rename to `tryBtcToSat` and return undefined for unexpected input, e.g. something that does not resemble `[+-][\d]+.?[\d]*` (best without using a regex)? Would that be okay for you? This is my last request regarding this issue :sweat_smile:

parrth20:
> Thinking about it again, I have one more plea: Code changed a bit and is quite "complicated" now and I don't think we should ever need to parse scientific notation. Should we just rename to `tryBtcToSat` and return undefined for unexpected input, e.g. something that does not resemble `[+-][\d]+.?[\d]*` (best without using a regex)? Would that be okay for you? This is my last request regarding this issue 😅

perfectly said it is a fair request...I’ve updated the PR accordingly.

I simplified the conversion to a strict tryBtcToSat(value) path, removed scientific notation handling entirely, and now return undefined for unexpected input instead of trying to coerce it. I also kept the parsing logic simple and explicit without regex.

to keep behavior safe, I updated the call sites to handle undefined gracefully and added tests covering the valid/invalid execution paths (including signs/whitespace, exponent rejection, and malformed inputs).
 please take another look...this should now match exactly the scope/style you asked for
and do let me know if you further need any mods.

theborakompanioni (approved):

PR #1046, opened 2026-02-11
add coverage for isFidelityBond() and getLocktime()

kunal-595:
#1023 
adds unit tests covering both the happy path and edge cases, including malformed derivation paths, missing locktime values and invalid month/locktime inputs


theborakompanioni (approved):
Thank you 🙏

PR #1050, opened 2026-02-12
fix(auth): throw on lock/stop refetch errors

### LINKED ISSUE #1049: [v2] Lock/Stop flows show success even when API call fails (refetch error swallowed)
Issue 
in v2, lock/stop actions use `await query.refetch()` inside `try/catch`.
TanStack Query `refetch()` suppresses errors unless `throwOnError` is enabled, so failure can be treated as "success".
Impact
User can see “wallet locked successfully” even if backend lock failed.
Auth/session can be cleared locally while backend action did not complete.
Stop-maker flow may miss proper failure handling.

Repro
1. Trigger a lock or stop action while backend/API call fails.
2. Observe success flow still continues (toast/state transition) instead of error handling.

Expected
Failure should go to `catch` and not run success side effects.

parrth20:
closes #1049 
lock/stop flows now call refetch({ throwOnError: true }) so API failures are handled by existing catch blocks instead of continuing success flow.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -116,7 +116,7 @@ function App() {
   const doOnLockWalletConfirm = async (navigate: NavigateFunction, t: TFunction<'translation', undefined>) => {
     if (!walletFileName) return
     try {
-      await lockWalletQuery.refetch()
+      await lockWalletQuery.refetch({ throwOnError: true })
       toast.success(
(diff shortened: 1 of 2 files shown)


saurabhraghuvanshii (approved):

saurabhraghuvanshii:
Thanks @parrth20

PR #1060, opened 2026-02-13
fix QR skeleton sizing in ReceivePage

kunal-595:
ReceivePage.tsx uses a dynamic Tailwind class for QR skeleton sizing, which is not detected at build time and results in missing CSS, this replaces it with a static sizing approach to ensure consistent skeleton dimensions and prevent layout shift


saurabhraghuvanshii:
@kunal-595 can you attach video recording of changing

theborakompanioni (approved):

saurabhraghuvanshii (approved):

kunal-595:
https://github.com/user-attachments/assets/f3b8135e-aced-4a3f-ac2b-a4f67129aa47

PR #1062, opened 2026-02-14
refactor(earn): move maker transition side-effects to effects

### LINKED ISSUE #1061: [v2] EarnPage has render-time side effects (setState/toasts) that should be moved to useEffect
main issue 

`EarnPage` currently triggers side effects during render:
state updates (`setIsWaitingMakerStart/Stop`)
toast side effects (`toast.dismiss`, `toast.success`)

This should not happen in render and can cause unstable behavior in StrictMode.

parrth20:
Moves maker start/stop completion side effects out of render and into `useEffect` hooks.
Changes

Added two guarded `useEffect` hooks in `src/components/earn/EarnPage.tsx`:
1.Start completion flow (`isWaitingMakerStart` -> running)
2.Stop completion flow (`isWaitingMakerStop` -> stopped)

Removed render-time side-effect block that called:
1.`setIsWaitingMakerStart/Stop`
2.`toast.dismiss(...)`
3.`toast.success(...)`
4.`scrollToTop()`

No behavior change intended, same transition/toast flow, but now executed in effects (React-safe pattern).
It's generally a good practice to follow this fixes #1061 

### DIFF
--- a/src/components/earn/EarnPage.tsx
+++ b/src/components/earn/EarnPage.tsx
@@ -66,11 +66,9 @@ export const EarnPage = ({ walletFileName }: EarnPageProps) => {
   const client = useApiClient()
   const jmSession = useStore(jmSessionStore, (state) => state.state)
   const isDeveloperMode = useStore(jamSettingsStore, (state) => state.state.developerMode)
+  const makerRunning = jmSession?.maker_running === true
 
   const walletInfo = useJamWalletInfoContext()
-
-  const [isWaitingMakerStart, setIsWaitingMakerStart] = useState(false)
-  const [isWaitingMakerStop, setIsWaitingMakerStop] = useState(false)
 
   const [moveToJarUtxo, setMoveToJarUtxo] = useState<FidelityBondUtxo | undefined>()
   const [renewBondUtxo, setRenewBondUtxo] = useState<FidelityBondUtxo | undefined>()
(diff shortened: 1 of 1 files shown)


theborakompanioni (approved):

src/components/earn/EarnPage.tsx
```diff
@@ -148,6 +146,25 @@ export const EarnPage = ({ walletFileName }: EarnPageProps) => {
     })
   }
 
+  useEffect(() => {
```
saurabhraghuvanshii, line 150: Hi @parrth20, LGTM! I was thinking—could we define a hook like `useMakerTransitionEffects`, or split it into
`useMakerStartTransitionEffect` and `useMakerStopTransitionEffect`? What do you think?

parrth20: thanks for the suggestion @saurabhraghuvanshii , makes sense.

i have kept it inline in this PR because the goal was a minimal behavioral change (just move render-time side effects into `useEffect` to fix the lint issue), and this logic currently has a single call site in `EarnPage`.

I’m happy to extract it if you prefer. my preference would be to do that in a small follow-up refactor PR (`useMakerStartTransitionEffect` + `useMakerStopTransitionEffect`) so this fix stays focused and easy to review.

PR #1065, opened 2026-02-14
refactor(ui): reuse shared page loading state

parrth20:
this PR removes duplicated page-loading UI by extracting a shared `PageLoading` component and reusing it across multiple pages.

Changes
1.Added `/src/components/ui/jam/PageLoading.tsx`.
2.Replaced duplicated loading markup in:
  LogsPage
  EarnPage
  SendPage
  SweepPage

the same loading layout/spinner/text block was repeated in multiple files. Centralizing it reduces redundancy and keeps loading-state UI consistent.

### DIFF
--- a/src/components/earn/EarnPage.tsx
+++ b/src/components/earn/EarnPage.tsx
@@ -13,6 +13,7 @@ import { Alert, AlertTitle } from '@/components/ui/alert'
 import { Button } from '@/components/ui/button'
 import { Card, CardAction, CardContent, CardDescription, CardFooter, CardHeader, CardTitle } from '@/components/ui/card'
 import { FeeConfigErrorAlert } from '@/components/ui/jam/FeeConfigErrorAlert'
+import { PageLoading } from '@/components/ui/jam/PageLoading'
 import PageTitle from '@/components/ui/jam/PageTitle'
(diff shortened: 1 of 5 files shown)


theborakompanioni (approved):

src/components/earn/EarnPage.tsx
saurabhraghuvanshii, line 30: @parrth20  remove this, if not in use

src/components/send/SendPage.tsx
saurabhraghuvanshii, line 26: same here

saurabhraghuvanshii (changes requested):

src/components/send/SendPage.tsx
parrth20: good catch , I re-checked this.
`Spinner` is still used later in both files (not only in the page-loading block):
`EarnPage`: spinner in maker/offer loading states and stop button pending state.
SendPage`: spinner in the “waiting for utxos to be marked as spent” state.

so the import is still required here. I kept it intentionally.
@saurabhraghuvanshii

saurabhraghuvanshii: Thanks @parrth20 for confirming

PR #1067, opened 2026-02-14
chore: add github PR template

saurabhraghuvanshii:
Added PR template  to follow best practice


parrth20:
> Added PR template to follow best practice

Thanks for adding this @saurabhraghuvanshii .

small suggestion: visual demo (screenshots/video) should be optional by default and required only when there are major changes or any UI/UX changes.  
for backend/internal/refactor-only PRs, it shouldn’t be mandatory.

saurabhraghuvanshii:
> > Added PR template to follow best practice
> 
> Thanks for adding this @saurabhraghuvanshii .
> 
> small suggestion: visual demo (screenshots/video) should be optional by default and required only when there are major changes or any UI/UX changes. for backend/internal/refactor-only PRs, it shouldn’t be mandatory.

Contributors can remove that requirement if the changes are purely backend-related. However, I think it is still useful to show how the site behaves after backend changes, as they can sometimes affect the UI indirectly.

parrth20:
> Contributors can remove that requirement if the changes are purely backend-related. However, I think it is still useful to show how the site behaves after backend changes, as they can sometimes affect the UI indirectly.

makes sense, agreed.

keeping it in the template gives maintainers extra confidence that the issue is fully validated end-to-end.  
and for backend-only changes, contributors can keep it brief or explicitly mention why a visual demo is not needed.

theborakompanioni:
I think mentioning creating storybook stories for all/most of UI related stuff and unit tests for everything else, would be nice.

theborakompanioni (approved):

PR #1071, opened 2026-02-15
chore: extract reason from backend error

### LINKED ISSUE #1070: [v2] login/create wallet toasts do not show backend error_description clearly
problem
In auth flows, backend errors are not always shown clearly in toasts.
`LoginPage` and `CreateWalletPage` mostly read `error.message`, but API failures can return object payloads like `error_description`, so users may see unclear messages.

repro
1.try login with wrong password (or trigger backend auth error).
2.try create wallet with an existing wallet name.
3.observe toast message is generic or less informative than backend response.

expected
toast should show the best available reason from backend:
`error_description`,`message`&`detail`
fallback to `global.errors.reason_unknown`

Proposed scope (small)
Add a tiny shared helper (e.g. `src/lib/errorReason.ts`) to extract readable reason from `unknown`.
Use it in:
   - `src/components/login/LoginPage.tsx`
   - `src/components/create/CreateWalletPage.tsx`
Keep behavior same except improved toast text.

parrth20:
fixes #1070 

changes made:
added shared helper: [errorReason.ts](app://-/index.html#)
updated login error toast to use shared extractor:   [LoginPage.tsx](app://-/index.html#)
updated create-wallet error toast to use shared extractor:  [CreateWalletPage.tsx](app://-/index.html#)

Screenshot 2026-02-15 at 10 14 18 AM

wallet already created with same name : 
image
image


parrth20:
after @saurabhraghuvanshii @theborakompanioni  you merge this don't merge above 2 PR's i need to do few cleanups but after this being merged!!

theborakompanioni (approved):

parrth20:
ping if we merge this error in PR #1073 will be resolved @theborakompanioni

saurabhraghuvanshii:
@parrth20  zero file?

parrth20:
> @parrth20 zero file?

@saurabhraghuvanshii maybe while doing commit i have messed up...i look into it!

parrth20:
> > @parrth20 zero file?
> 
> @saurabhraghuvanshii maybe while doing commit i have messed up...i look into it!

@saurabhraghuvanshii you’re right, I messed up branch state during update.
I’m force-pushing a cleaned branch with only the auth error-reason changes (or closing if already in base).

saurabhraghuvanshii (approved):

theborakompanioni:
Are all these different error message sources really necessary?
I mean.. we had `message` and `error_description` (from `Error` or `ErrorMessage`)... where do all the other come from?
If they come from the API, would it make sense to generalize this into an `Error` with `message` and then use `getErrorReason`?

parrth20:
> Are all these different error message sources really necessary? I mean.. we had `message` and `error_description` (from `Error` or `ErrorMessage`)... where do all the other come from? If they come from the API, would it make sense to generalize this into an `Error` with `message` and then use `getErrorReason`?

good question, and you’re right to call it out.

message + error_description are the only fields we really expect from the JM API contract.
the extra keys were added defensively because in practice we also get wrapped error shapes from client layers (for example nested [response.data](app://-/index.html#) / error, and sometimes detail), so the backend reason doesn’t get lost behind generic transport messages.

parrth20:
> Are all these different error message sources really necessary? I mean.. we had `message` and `error_description` (from `Error` or `ErrorMessage`)... where do all the other come from? If they come from the API, would it make sense to generalize this into an `Error` with `message` and then use `getErrorReason`?

but that said i agree with your direction, we should normalize errors once at the API boundary into a single app-level Error/ErrorMessage shape, then keep UI extraction simple.
I can open a follow-up cleanup PR to do exactly that and reduce this helper back to a stricter contract should i do that ?

theborakompanioni:
> > Are all these different error message sources really necessary? I mean.. we had `message` and `error_description` (from `Error` or `ErrorMessage`)... where do all the other come from? If they come from the API, would it make sense to generalize this into an `Error` with `message` and then use `getErrorReason`?
> 
> but that said i agree with your direction, we should normalize errors once at the API boundary into a single app-level Error/ErrorMessage shape, then keep UI extraction simple. I can open a follow-up cleanup PR to do exactly that and reduce this helper back to a stricter contract should i do that ?

No, its not too important right now, but a candidate for clean-ups indeed.
Remove everything that is not strictly needed.

PR #1082, opened 2026-02-16
Fix useEffect deps causing toast spam on EarnPage

kunal-595:
The start/stop mutation objects are new refrences on each render, causing both effects to run repetedly, narrowed dependencies to isSuccess and reset, which are the only values used, to prevent repeated toasts and resets

### DIFF
--- a/src/components/earn/EarnPage.tsx
+++ b/src/components/earn/EarnPage.tsx
@@ -155,7 +155,8 @@ export const EarnPage = ({ walletFileName }: EarnPageProps) => {
     toast.success(t('earn.alert_running'), { id: 'earn.alert_running' })
     scrollToTop()
     startMaker.reset()
-  }, [jmSession, makerRunning, startMaker.isSuccess, startMaker, t])
+    // eslint-disable-next-line react-hooks/exhaustive-deps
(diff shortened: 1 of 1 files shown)


theborakompanioni:
I guess this was introduced in a recent PR when the `isWaitingMakerStart` was moved from the `useMutation` hook, right? Can we revert to the old approach? What do you think?

theborakompanioni:
I am not experiencing this behaviour - it is currently not optimal, but I do not see what this fixes. Can you elaborate?

kunal-595:
thanks for checking after re-testing on latest v2, I do not see any user visible impact. The change was mainly to tighten the effect dependencies ,if reverting to the previous isWaitingMakerStart approach is preferred for consistency, i am happy to switch.

theborakompanioni:
> thanks for checking after re-testing on latest v2, I do not see any user visible impact. The change was mainly to tighten the effect dependencies ,if reverting to the previous isWaitingMakerStart approach is preferred for consistency, i am happy to switch.

I have a small refactor regarding this and will push it soon, okay for you?

theborakompanioni:
Closing this in favour of https://github.com/joinmarket-webui/jam/pull/1144 - please reopen if you think this is a mistake or if #1144 does not solve your issue. :pray:

PR #1069, opened 2026-02-14
V2 import wallet

parrth20:
Summary : - 
adds a complete Import Wallet flow in v2 and wires the existing login CTA to a working route/page.

Changes made 
added route constant `importWallet` in `/Users/parthbandwal/Desktop/JamSoB/jam/src/constants/routes.ts`.
added unauthenticated route mapping in `/Users/parthbandwal/Desktop/JamSoB/jam/src/App.tsx`.
enabled "Import wallet" action from login in `/Users/parthbandwal/Desktop/JamSoB/jam/src/components/login/LoginPage.tsx`.
added `/Users/parthbandwal/Desktop/JamSoB/jam/src/components/import/ImportWalletPage.tsx` with:
   - wallet name + uniqueness validation
   - seed phrase input + normalization/validation
   - password + confirm password validation
   - `recoverwallet` mutation integration
   - auth store update on success
   - success/error toast handling and navigation to home
   - better backend error reason extraction (`message` / `error_description` / `detail`)
updated query-delay option compatibility with current v2 API (`delayAfter`).

Why
Previously the Import Wallet CTA existed but was disabled/unimplemented. This closes the gap and enables wallet recovery from mnemonic in v2.

@theborakompanioni @saurabhraghuvanshii 
fixes #1068 

video low quality as only 10mb is allowed.
https://github.com/user-attachments/assets/844a22c2-b490-4980-a201-f6b3c27678cd

(4x speed)
https://github.com/user-attachments/assets/d4c47e55-b75d-4843-a146-0f584b4117a9

PR #1085, opened 2026-02-16
fix(auth): sync session auth across browser tabs

### LINKED ISSUE #1083: [v2] keep login state consistent across browser tabs (sessionStorage)
Problem
when a user is already logged in and opens JAM in a new tab, the new tab can land on the login page instead of the main app page.

why this matters
this feels like a broken login/session UX and was reported during review discussion on #1069.

Current behavior
tab A is logged in.
user opens Tab B to JAM.
tab B may show login even though the same browser already has an active session in Tab A.

proposed fix
keep `sessionStorage` (no persistence model change), and add cross-tab auth synchronization:
use `BroadcastChannel` for auth state request/response on new tab open.
broadcast auth-clear events on logout/lock.

parrth20:
summary
this follow-up PR fixes cross-tab login UX while keeping `sessionStorage` as the auth storage model.

context
raised during review discussion of #1069: opening JAM in a new tab could show login even when another tab was already authenticated.

what changed
added `useAuthCrossTabSync` hook:
     new tab requests auth state from existing authenticated tab via `BroadcastChannel`.
     existing authenticated tab responds with current auth state.
     logout/lock broadcasts an auth-cleared event to all tabs.
wired hook into app root and updated auth-clear flow to broadcast only on local logout/lock.

why this approach
preserves current `sessionStorage` security/persistence decision.
solves new-tab UX without broad auth architecture changes.
keeps fix focused and low-risk.

closes #1083 

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import { lazy, Suspense, useEffect, useMemo, useState } from 'react'
+import { lazy, Suspense, useCallback, useEffect, useMemo, useState } from 'react'
 import type { PropsWithChildren } from 'react'
 import { lockwalletOptions } from '@joinmarket-webui/joinmarket-api-ts/@tanstack/react-query'
 import { token } from '@joinmarket-webui/joinmarket-api-ts/jm'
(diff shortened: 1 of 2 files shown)


parrth20:
https://github.com/user-attachments/assets/c7191b17-3892-490e-84bb-832edf64090b

ping @saurabhraghuvanshii @theborakompanioni

theborakompanioni:
I am not sure on this one. Can we postpone discussing it to after v2 is initially released?

saurabhraghuvanshii:
Yes, we can discuss it after the release.

parrth20:
> I am not sure on this one. Can we postpone discussing it to after v2 is initially released?

thanks for the feedback tbk, that makes sense. I agree we should avoid adding scope right before v2 release. I’ll postpone this and revisit it right after the initial v2 launch. If you prefer, I can close this PR for now and reopen with a fresh one post-release, I’ll stay focused on release-critical fixes only

theborakompanioni:
Closing for now. Not planned for v2.

PR #1091, opened 2026-02-16
chore: test for Balance component and align currency symbols

theborakompanioni:
Adds a test for balance component and align currency icons. BTC icon on the left, sats icon to the right.


parrth20:
nice updates tbk @theborakompanioni , especially the added Balance tests they improve confidence a lot. I had two small clarifications:

1.CurrencySymbol styling now seems more caller-driven, was that intentional for all current usages?
2.The login active-dot tweak also looks good—just checking if you prefer keeping it in this PR or splitting for scope clarity.
no blockers from my side as such which i can see.

parrth20 (approved):

theborakompanioni:
> nice updates tbk @theborakompanioni , especially the added Balance tests they improve confidence a lot. I had two small clarifications:
> 
> 1.CurrencySymbol styling now seems more caller-driven, was that intentional for all current usages? 

Yes, BTC symbol was a unicode char, all other symbols are icons. This as been generalized. 

> 2.The login active-dot tweak also looks good—just checking if you prefer keeping it in this PR or splitting for scope clarity. no blockers from my side as such which i can see.

You are right. I try to keep my updates more scoped. Thanks for the hint. Hope it is okay that small, but potentially unrelated stuff, is included occasionally - till v2 is ready.

PR #1081, opened 2026-02-15
[v2] improve user onboarding UX on login

parrth20:
summary
add a new `OnboardingDialog` for v2 login that reuses existing legacy onboarding wording (`onboarding.*` i18n keys)
auto-show onboarding for first-time visitors and persist dismissal in local storage
add a login-page entry point to reopen onboarding anytime
include clearer step flow with progress indicators and improved visual hierarchy for mobile/desktop

Closes #1080.
@theborakompanioni @saurabhraghuvanshii

PR #1097, opened 2026-02-17
fix(onboarding): address follow-up nits from #1081 (splash text + storybook default page)

parrth20:
follow-up PR for onboarding review nits. #1081 
fix onboarding splash warning links rendering
set a neutral default start page for `storybook:up` updates script to start on `/story/button--default` instead of landing on Login stories

fixes 
image

image
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
     "prettier": "prettier --write \"**/*.{js,jsx,ts,tsx,json,css,md}\"",
     "format": "npm run lint -- --fix && npm run prettier",
     "preview": "vite preview",
-    "storybook:up": "storybook dev -p 6006 --initial-path=/story/core-button--default",
+    "storybook:up": "storybook dev -p 6006",
     "storybook:build": "storybook build",
     "test": "vitest run",
     "test:unit": "vitest run --project unit",
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
     "prettier": "prettier --write \"**/*.{js,jsx,ts,tsx,json,css,md}\"",
     "format": "npm run lint -- --fix && npm run prettier",
     "preview": "vite preview",
-    "storybook:up": "storybook dev -p 6006 --initial-path /story/button--default",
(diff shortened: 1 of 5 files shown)


saurabhraghuvanshii (approved):

package.json
```diff
@@ -18,7 +18,7 @@
     "prettier": "prettier --write \"**/*.{js,jsx,ts,tsx,json,css,md}\"",
     "format": "npm run lint -- --fix && npm run prettier",
     "preview": "vite preview",
-    "storybook:up": "storybook dev -p 6006",
+    "storybook:up": "storybook dev -p 6006 --initial-path=/story/core-button--default",
```
theborakompanioni:
```suggestion
    "storybook:up": "storybook dev -p 6006",
```
What is the reason behind this change?

I was thinking about http://localhost:6006/?path=/docs/page-login-logincard--docs -> the onboarding dialog should be separated from the LoginCard dialog.

theborakompanioni (changes requested):

package.json
parrth20: > What is the reason behind this change?
> 
> I was thinking about http://localhost:6006/?path=/docs/page-login-logincard--docs -> the onboarding dialog should be separated from the LoginCard dialog.

thanks for calling this out. the reason i added --initial-path was to avoid storybook landing on login-related stories first while i was testing onboarding follow-ups. but your point makes sense: storybook:up should stay default, and onboarding/login separation should be handled at the story level instead. i'll revert this script change in this pr and keep the splash-text fix only. i can open a follow-up for separating onboarding dialog from logincard docs.

parrth20:
ping @theborakompanioni

parrth20:

theborakompanioni:
@parrth20 I think CardLogin should not be aware of OnboardingDialog at all. Just a detail and can be done later. Approved.

theborakompanioni (approved):

parrth20:
> @parrth20 I think CardLogin should not be aware of OnboardingDialog at all. Just a detail and can be done later. Approved.

@theborakompanioni  ok I'll see into this!

PR #1105, opened 2026-02-17
feat: warn about reused addresses below UTXO table

kunal-595:
Destructive alert below the table when any UTXOs sit on reused addresses. 

https://github.com/user-attachments/assets/c6004997-2ede-402b-9726-ac15a710d985


saurabhraghuvanshii (approved):

theborakompanioni (approved):
I think it can even be displayed more prominently.
Maybe we can even indicate it in the jar view (e.g. by hinting at it via the jar icon).
Also, this lets me think about automatically solving such issues, e.g. spending all utxos on a reused address as collaborative transaction (freeze other utxos, sweep jar, unfreeze previously frozen utxos).
Let's keep that in mind when v2 reached feature parity.

kunal-595:
Agreed, making reuse visible at the Jar icon level makes sense it's too buried in the table right now.The automated freeze-sweep-unfreeze flow is a great idea for the post parity roadmap. It’s a lot of state mangement to get right but it's much better than expecting users to manual sweep,i will stick to the structured key value refactor for now to keep us moving toward v2 parity, let's get this merged and i will start on follow ups.

PR #1109, opened 2026-02-17
Optional block explorer links after sending

kunal-595:
Toggle in Display settings (off by default), when on the send success alert gets a mempool.space link.
Returns null for regtest so no broken link shows up
 

https://github.com/user-attachments/assets/2d174c62-6894-4b5b-87ab-e4b23301eb63


saurabhraghuvanshii (approved):

theborakompanioni:
I am not sure about this one @kunal-595.
Generally like the idea, but hardcoding mempool.space is not something I'd merge. Maybe it can be more generic, e.g. letting a user provide the url template - this way a local self-hosted block explorer can be used.
However, even then, maybe we can postpone the feature? Would that be okay for you?

theborakompanioni:
@kunal-595 Would you make the url configurable and potentially add a component that displays txids with a link if an explorer is configured?
If not, I'd close this issue for now and we concentrate on the feature parity to legacy. But imho it is a "nice to have" feature!

kunal-595:
I will make the URL configurable via a template and add the conditional component for txid links , will update the pr shortly.

MarnixCroes:
I agree the idea is good. But having leading people to third party block explorers is bad for privacy.

theborakompanioni (changes requested):
See above.

theborakompanioni:
Closing as wont do for now!

PR #1112, opened 2026-02-17
fix(login): use button for wallet refresh action

parrth20:
in login card, wallet refresh was attached to a clickable icon instead of a proper button. this worked for mouse users, but semantics/accessibility were weak and behavior could be inconsistent across browsers/tools.

this pr replaces that clickable wallet icon with a real icon button while keeping the same refresh behavior and visual look.

now if we hit tab+enter/enter it works fine.


saurabhraghuvanshii:
@parrth20  attach video Please

parrth20:
> @parrth20 attach video Please

there isn’t any visual/flow change in this pr, so i didn’t attach a video. this update is semantic only: i replaced a clickable icon with a real button (<button>), which is good practice for accessibility and consistent behavior. if you still want, i can share a quick clip.
here my cursor was not btn 



https://github.com/user-attachments/assets/e60eccf5-7804-42fd-af9e-9a2144598307

saurabhraghuvanshii (approved):

PR #1114, opened 2026-02-17
Implement offline queue

### LINKED ISSUE #1113: [v2] implement offline queue + retry for wallet actions
when network/api connection drops, user actions (send/import/rescan/settings updates) fail immediately and are lost from flow context.
add an offline action queue in v2 that stores retryable operations, shows their status, and retries automatically when connection is restored.

parrth20:
this pr adds an offline action queue for v2 wallet actions so user operations are not lost when network/api connectivity drops.

what i have changed
added centralized connectivity detection (browser offline + api unreachable)
added persisted offline queue with action payload + metadata + status
added background retry processor with exponential backoff
integrated queue flow for:
  send
  import wallet
  rescan chain
  fee settings updates
added global queue ui showing queued/retrying/failed states
added manual controls: retry now and cancel
fixed token refresh behavior so offline refresh failures do not force logout

@theborakompanioni please give me your feedback : 
does this queue scope look right for v2 (send/import/rescan/fee updates)?
is local persistence policy acceptable for queued actions in this release?
is the offline token-refresh behavior (no forced logout on connectivity failures) acceptable from product/security perspective?
I have sent you one video regarding this on element please go through it also.

fixes #1113

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -26,6 +26,7 @@ import ErrorPage from '@/components/error/ErrorPage'
 import ImportWalletPage from '@/components/import/ImportWalletPage'
 import { Layout } from '@/components/layout/Layout'
 import LoginPage from '@/components/login/LoginPage'
+import { OfflineActionQueueWidget } from '@/components/offline/OfflineActionQueueWidget'
 import { OrderbookPage } from '@/components/orderbook/OrderbookPage'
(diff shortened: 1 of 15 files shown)


theborakompanioni:
Closing, as this is not yet to be considered for v2. I think it is better to fail hard if there is an error.

PR #1115, opened 2026-02-17
chore: expand utxo rows with details

theborakompanioni:
Show details about a utxo in an expanded row in the Jar Details Overlay.
Just dumps the json details via `JSON.stringify`, good enough for now.

image


kunal-595:
LGTM! 
The fallback ' ' in the expand column is currently unreachable since getRowCanExpand is hardcoded to true, but it’s fine for now. The raw JSON dump is solid  but would you like me to take point on a follow-up PR to render these details as a structured, user-friendly key-value list instead?

parrth20:
Hey tbk, reviewed this PR locally and it looks good to me. I checked the UTXO row expansion + toast cleanup flow and also ran lint, unit tests, and build (all passed). optional minor polish: we could add a max-height/scroll on the JSON details block for very large UTXO payloads, structured key-value rendering can be a follow-up improvement.. Otherwise LGTM from my side.

parrth20 (approved):

parrth20:
> Hey tbk, reviewed this PR locally and it looks good to me. I checked the UTXO row expansion + toast cleanup flow and also ran lint, unit tests, and build (all passed). optional minor polish: we could add a max-height/scroll on the JSON details block for very large UTXO payloads, structured key-value rendering can be a follow-up improvement.. Otherwise LGTM from my side.

If you want i can look upon follow up i suggested and raise a , do you want me to do that ?

PR #1146, opened 2026-02-23
fix(logs): handle native/standalone logs correctly and improve logs ui

parrth20:
summary
fixes logs behavior across native and standalone modes and cleans up logs ux.

changes
add robust log endpoint handling with fallback (`/api/v0/log` -> `/jam/api/v0/log`)
detect html fallback responses and treat them as unsupported logs endpoint
show a clean unsupported warning in native mode (no duplicated message in alert + log body)
split logs viewer layout for page vs overlay so log panel sizing works correctly
improve unsupported message formatting for better readability

https://github.com/user-attachments/assets/46bf770f-0552-4026-92d6-2fb5af33d444

### DIFF
--- a/src/components/LogsContent.tsx
+++ b/src/components/LogsContent.tsx
@@ -1,7 +1,6 @@
 import { AlertTriangleIcon, Loader2Icon } from 'lucide-react'
 import { useTranslation } from 'react-i18next'
 import { LogViewer } from '@/components/logging/LogViewer'
-import type { LogViewerVariant } from '@/components/logging/LogViewer'
 import { useJmwalletdStdoutLog } from '@/components/logging/useJmwalletdStdoutLog'
 import { Alert, AlertDescription } from '@/components/ui/alert'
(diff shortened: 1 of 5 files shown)


parrth20:
I’m a bit confused here, but from my testing it looks like logs are available only in standalone mode (`npm run dev:secondary`). Am I correct?
Also, as a follow-up, should we hide/disable the Logs viewer in native mode and show only a guidance message?
@theborakompanioni

theborakompanioni (changes requested):
I don't understand anymore what this PR is supposed to do..

parrth20:
> I don't understand anymore what this PR is supposed to do..

ping @theborakompanioni 
address all almost all things we talked about, if this works or if need some changes i can do that in follow ups as #1150 was merged so i need this also to be merged to work on follow ups...!! details of changes made here are in commit message

parrth20:
@theborakompanioni said he'll look in cleanups

Issue #1149, opened 2026-02-24
[v2] improve logs page (search, highlight, stable modal/page behavior)


parrth20:
follow up issue #1145 

logs page needs better usability and consistency.
current issues:
search does not highlight matches, so it is hard to quickly spot results
clear action is not obvious in footer logs flow
logs view refreshes in a jumpy way when auth/session refresh happens
/logs page and footer modal behavior is not fully aligned

ping @theborakompanioni @saurabhraghuvanshii

parrth20:
I have raised a PR against this issue check #1150 @theborakompanioni @saurabhraghuvanshii

PR #1150, opened 2026-02-24
feat(logs): improve logs viewer ux with search highlight and smoother refresh

parrth20:
follow up #1146 
improves logs ux for both footer modal and /logs page.

closess #1149 
what changed:
aligned logs page and modal to use consistent viewer layout behavior
added search input with line filtering and dark yellow match highlight
added clear search actions (inline x + clear button)
added match count/no-match feedback
improved auto-scroll behavior to avoid jumpy movement while reading
reduced flicker during refresh by keeping previous content and decoupling query key from token refresh
simplified non-standalone fallback text

demo - https://www.youtube.com/watch?v=Bce2IeIx7qw

### DIFF
--- a/src/components/logging/useJmwalletdStdoutLog.ts
+++ b/src/components/logging/useJmwalletdStdoutLog.ts
@@ -24,10 +24,17 @@ export function useJmwalletdStdoutLog({ enabled = true }: UseJmwalletdStdoutLogP
   const { t } = useTranslation()
   const token = authState?.auth?.token
 
-  const logQuery = useQuery({
-    queryKey: ['logs', JMWALLETD_LOG_FILE_NAME, token],
+  const logQuery = useQuery<string>({
+    queryKey: ['logs', JMWALLETD_LOG_FILE_NAME],
     enabled: enabled && token !== undefined,
     retry: false,
+    refetchOnWindowFocus: false,
+    refetchInterval: (query) => {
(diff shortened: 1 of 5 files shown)


src/components/logging/useJmwalletdStdoutLog.ts
```diff
@@ -73,7 +80,7 @@ export function useJmwalletdStdoutLog({ enabled = true }: UseJmwalletdStdoutLogP
   const logFileContent = useMemo(() => {
     if (logQuery.data) return logQuery.data
     if (!isDevMode() || alert?.variant !== 'warning') return undefined
-    return `${alert.message}\n`.repeat(1_000)
+    return `${alert.message}\nRun Jam in standalone mode: npm run dev:secondary.`
```
saurabhraghuvanshii, line 83: @parrth20  why this? I don't have an Idea.

parrth20: @saurabhraghuvanshii so this change is only for the dev fallback path (when real log fetch is unavailable, e.g. non-standalone setup). previously we repeated the same warning 1000 times, which looked like real logs and made the ui noisy. i changed it to a single clear fallback line so users understand why logs are not available. in standalone mode, real logs are unchanged.

saurabhraghuvanshii:
Thanks @parrth20 lgtm!!

PR #1151, opened 2026-02-24
fix(send): validate destination address matches wallet network

kunal-595:
the send form did not enforce network consistency, allowing cross network addreses to pass initial validation this change checks the destination address network against the wallet’s detected network using useDetectNetwork() and getAddressInfo()


theborakompanioni (approved):
utACK.

💪

PR #1165, opened 2026-03-18
fix(logs): improve log file page with robust api fallback and search

parrth20:
summary
fixes log file contents not showing and improves the log file page with search functionality. both the standalone page (`/logs`) and the modal overlay work correctly.

@theborakompanioni

### DIFF
--- a/src/components/LogsOverlay.tsx
+++ b/src/components/LogsOverlay.tsx
@@ -7,10 +7,10 @@ import { LogsContent } from './LogsContent'
 
 type LogsOverlayProps = WithRequiredProperty<Omit<ComponentProps<typeof Dialog>, 'children'>, 'open' | 'onOpenChange'>
 
-export function LogsOverlay({ open, onOpenChange, ...dialogProps }: LogsOverlayProps) {
+export function LogsOverlay({ open, onOpenChange }: LogsOverlayProps) {
   const { t } = useTranslation()
   return (
-    <Dialog open={open} onOpenChange={() => onOpenChange(false)} {...dialogProps}>
+    <Dialog open={open} onOpenChange={() => onOpenChange(false)}>
       <DialogContent className="data-[state=closed]:slide-out-to-bottom data-[state=open]:slide-in-from-bottom flex h-screen max-w-screen! flex-col rounded-none border-none">
         <DialogHeader>
           <DialogTitle className="flex items-center gap-2">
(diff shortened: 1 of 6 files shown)


parrth20:
image

theborakompanioni:
Whenever new content is loaded it scrolls from top to bottom again.
I like the simplification, I don't like adding the extra code in [src/lib/api/logs.ts](https://github.com/joinmarket-webui/jam/pull/1165/changes#diff-66b2f581f4cf7787302bbc2ada1c27338a3d24fa5424b4937f669d3dad5e3ae9): Why the complexity? If there is an error, show the error, if there is no error, show the logs.., no?

parrth20:
> Whenever new content is loaded it scrolls from top to bottom again. I like the simplification, I don't like adding the extra code in [src/lib/api/logs.ts](https://github.com/joinmarket-webui/jam/pull/1165/changes#diff-66b2f581f4cf7787302bbc2ada1c27338a3d24fa5424b4937f669d3dad5e3ae9): Why the complexity? If there is an error, show the error, if there is no error, show the logs.., no?

simplified [src/lib/api/logs.ts](https://github.com/joinmarket-webui/jam/pull/1165/changes#diff-66b2f581f4cf7787302bbc2ada1c27338a3d24fa5424b4937f669d3dad5e3ae9) by removing the extra fallback/error complexity.
updated logs loading flow so behavior is straightforward: if request errors, show error; if successful, show logs.
adjusted log viewer refresh behavior to reduce visible flicker/scroll jump when content updates.

parrth20:
ping @theborakompanioni

src/components/LogsOverlay.tsx
```diff
@@ -7,10 +7,10 @@ import { LogsContent } from './LogsContent'
 
 type LogsOverlayProps = WithRequiredProperty<Omit<ComponentProps<typeof Dialog>, 'children'>, 'open' | 'onOpenChange'>
 
-export function LogsOverlay({ open, onOpenChange, ...dialogProps }: LogsOverlayProps) {
```
theborakompanioni, line 10: Why remove `dialogProps`?

src/components/logging/useJmwalletdStdoutLog.ts
```diff
@@ -23,17 +23,10 @@ export function useJmwalletdStdoutLog({ enabled = true }: UseJmwalletdStdoutLogP
...
   const token = authState?.auth?.token
 
-  const logQuery = useQuery<string>({
-    queryKey: ['logs', JMWALLETD_LOG_FILE_NAME],
+  const logQuery = useQuery({
+    queryKey: ['logs', JMWALLETD_LOG_FILE_NAME, token],
     enabled: enabled && token !== undefined,
     retry: false,
-    refetchOnWindowFocus: false,
-    refetchInterval: (query) => {
```
theborakompanioni, line 31: Why remove the `refetchInterval`?

src/components/LogsOverlay.tsx
parrth20: It’s not critical/imp for current behavior

theborakompanioni: Should it be removed it from the component type as well then?
Isn't it confusing to be able to add e.g. `modal={true}` and it not having any effect?

PR #1166, opened 2026-03-18
JoinMarket NG backend for regtest environment

m0wer:
Adds JoinMarket NG as a backend in the regtest setup to test compatibility.

Most functionality working! Tumbler (sweep) does not work here, see #1232.

TODO:

- [ ] Check switch wallet feature.
- [ ] Check rescan from height feature.
- [ ] Expose more configuration options, such as forced 0 fee for bondless toggle.

## Dev setup

```shell
npm run regtest:rebuild &&  npm run regtest:up &&  npm run regtest:init
npm run regtest:mine &
npm run jm-ng:dev
```

Wallet password is `test`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,7 +84,7 @@ There are many ways you can contribute: testing, sharing ideas, writing document
 
 Want to get your hands dirty with code? Great! [CONTRIBUTING.md](CONTRIBUTING.md) and [docs/developing.md](docs/developing.md) are where it's at.
 
-For local development against a separately running `joinmarket-ng` backend, start `jmwalletd` and the jm-ng orderbook watcher first, then run Jam with `npm run dev`. Jam's default Vite proxy targets `https://127.0.0.1:28183` for `/api` and `/jmws`, and `http://127.0.0.1:8080` for `/obwatch`. If you are using Jam's regtest jm-ng services instead, run `npm run jm-ng:dev`.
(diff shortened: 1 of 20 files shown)


theborakompanioni:
@m0wer Would it be possible to keep the joinmarket-clientserver as well and add jm-ng as an additional instance?

m0wer:
> @m0wer Would it be possible to keep the joinmarket-clientserver as well and add jm-ng as an additional instance?

Sure! That's actually what we have in the compatibility tests in the jm-ng repo (a mix of reference and ng makers and takers).

What would be your preferred setup? Leave the current 3 jm-ref instances and add a jm-ng one? What about the directory server? Anything else?

theborakompanioni:
> > @m0wer Would it be possible to keep the joinmarket-clientserver as well and add jm-ng as an additional instance?
> 
> Sure! That's actually what we have in the compatibility tests in the jm-ng repo (a mix of reference and ng makers and takers).
> 
> What would be your preferred setup? Leave the current 3 jm-ref instances and add a jm-ng one? What about the directory server? Anything else?

Yeah, maybe just leave the three joinmarket-clientserver instances as is and just add two jm-ng ones. Ideally devs can change the primary backend via env vars and running `npm run dev` and `npm run jm-ng:dev` (or similar).
Regarding the directory server: Would also be nice to switch between them via env vars, but if that adds too much complexity, I am fine with running both or replacing it completely (I suppose the old setup works with your new directory server nicely..). What do you think? Any other recommendations?

theborakompanioni:
Closing this: Already in `master` (454daa2839cf5f0a9aca66c15ba5abaa36ae9417)

PR #1168, opened 2026-03-21
chore(import): prepare rescan by splitting import flow steps 

theborakompanioni:
Prepares the import flow to add rescanning options.
Splits the flow in 3 steps:
1. wallet details
2. import details
3. confirmation


parrth20:
@theborakompanioni should i work on this?

theborakompanioni:
> @theborakompanioni should i work on this?

No, I am on it : -)

src/components/import/ImportWalletPage.tsx
```diff
@@ -194,221 +106,70 @@ const ImportWalletPage = () => {
     } catch (error: unknown) {
       const reason = getErrorReason(error, t('global.errors.reason_unknown'))
       toast.error(t('import_wallet.error_importing_failed', { reason }))
+      throw error
```
parrth20: optional nit
remove throw error after the toast to avoid unhandled promise-rejection noise.

src/components/import/ImportDetailsForm.tsx
```diff
@@ -0,0 +1,178 @@
...
+interface ImportDetailsFormValues {
+  seedPhrase: string
+}
+
+const normalizeSeedPhrase = (value: string | undefined) =>
+  (value ?? '')
+    .toLowerCase()
+    .split(/[\s,]+/)
+    .map((word) => word.trim())
+    .map((word) => word.replace(/^\d+\.$/, ''))
```
parrth20: optional nit
normalizeSeedPhrase replaces numbered tokens like 1. with an empty string but does not remove empty tokens before joining. like a pasted numbered phrase like 1. abandon 2. ... therefore keeps blank words and fails validation, which is a regression from prior behavior that filtered empties. Adding filter(Boolean) before join fixes this, what's your take on this?

parrth20 (approved):
LGTM!!

src/components/import/ImportWalletPage.tsx
theborakompanioni: Thanks! Nice catch. Done.

src/components/import/ImportDetailsForm.tsx
theborakompanioni: Good point. 
In general I think we should do as little changes to user input as possible, but this feels to make sense.

PR #1169, opened 2026-03-23
memoize router creation using useMemo

### LINKED ISSUE #1155: [v2] avoid recreating browser router on every app render
right now `createBrowserRouter()` is being called inside the `App` component body, which means a new router instance gets created on every render.

since `App` re-renders on state, context, and translation changes, this causes the router to be recreated repeatedly. react router expects the router instance to be stable, and recreating it can reset internal navigation state, interrupt pending fetchers, and potentially cause subtle routing bugs.

the router should be memoized with the correct dependencies or refactored so it is created once and remains stable across renders.

theAnuragMishra:
This PR memoizes router creation using `useMemo` . prevents router from being recreated every time app state other than the dependencies change.

while this works, i think there are other better options, maybe. like moving to react 19 which does memoization automatically (not really sure if the compiler would memoize cases like these). or just removing conditional routing and adding state guards in the components for the routes.

fixes #1155

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -87,7 +87,7 @@ export default function App() {
     [reloadCurrentWalletInfo],
   )
 
-  const router = createBrowserRouter(
+  const router = useMemo(()=>createBrowserRouter(
     createRoutesFromElements(
       <Route
         id="base"
@@ -218,7 +218,7 @@ export default function App() {
         v7_normalizeFormMethod: true,
       },
     },
-  )
+  ), [currentWallet, sessionConnectionError]);
 
   if (settings.showOnboarding === true) {
     return (


theborakompanioni:
Hey @theAnuragMishra. Thanks for the PR.
I think the ticket is specifically about `v2` -> can you check the branch?

theAnuragMishra:
ah yes sorry. i missed `[v2]` . i will close this one and try to fix in the v2 branch.

theAnuragMishra:
@theborakompanioni before i try to work on the issue in v2, i'd like to confirm what appraoch would be preferable. like i described, we could make a simple fix by memoizing, or follow what you may call the idiomatic way, that is, create a stable router once outside App and the route elements read state themselves instead of passing a lot of props into the route elements.

theborakompanioni:
> @theborakompanioni before i try to work on the issue in v2, i'd like to confirm what appraoch would be preferable. like i described, we could make a simple fix by memoizing, or follow what you may call the idiomatic way, that is, create a stable router once outside App and the route elements read state themselves instead of passing a lot of props into the route elements.

Yeah, if that is possible, let's do it.
I am not an too familiar with the best practices - I'll learn a thing or two from your upcoming PR :slightly_smiling_face:

theAnuragMishra:
#1175 . some more work needs to be done though. but let's move the discussion there and close this pr.

PR #1171, opened 2026-03-24
build: add tanstack query eslint plugin

theborakompanioni:
Apply @tanstack/eslint-plugin-query.

See: https://github.com/TanStack/query/blob/main/docs/eslint/exhaustive-deps.md

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -49,7 +49,7 @@ import { EarnReportPage } from './components/earn/report/EarnReportPage'
 import { LockWalletConfirmDialog } from './components/ui/jam/LockWalletConfirmDialog'
 import { Spinner } from './components/ui/spinner'
 import { WalletJarsDetailsPage } from './components/wallet/WalletJarsDetailsPage'
-import { type RescanInfo, useRescanStatus } from './context/JamSessionInfoContext'
+import { useRescanStatus } from './context/JamSessionInfoContext'
 import { JamSessionInfoContextProvider } from './context/JamSessionInfoContextProvider'
 import { useJamWalletInfoContext } from './context/JamWalletInfoContext'
 import { useJmWebsocket } from './hooks/useJmWebsocket'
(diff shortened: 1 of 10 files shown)


src/App.tsx
```diff
@@ -443,26 +443,27 @@ const RELOAD_WALLET_INFO_DELAY: {
 const WalletInfoAutoReload = ({ walletFileName }: { walletFileName: WalletFileName }) => {
   const { rescanInfo: currentRescanInfo } = useRescanStatus()
   const { refetch: refetchWalletBalance, utxosHashHex } = useJamWalletInfoContext()
-  const [previousRescanInfo, setPreviousRescanInfo] = useState<RescanInfo>(currentRescanInfo)
+  const [previousRescanning, setPreviousRescanning] = useState<boolean>(currentRescanInfo.rescanning)
```
parrth20, line 460: @tanstack/query/exhaustive-deps is disabled in the new lint PR better to refactor this useQuery (or make refetchWalletBalance stable via ref/callback) so we don’t bypass the rule we just introduced.

src/components/LogsContent.tsx
```diff
@@ -32,7 +32,9 @@ export const LogsContent = ({ enabled, className }: LogsContentProps) => {
         </Alert>
       )}
 
-      {logFileContent && <LogViewer fileName={fileName} value={logFileContent} refresh={refresh} />}
+      {logFileContent && (
+        <LogViewer fileName={fileName} value={logFileContent} refresh={async () => void (await refresh())} />
```
parrth20: optional nit
refresh={async () => void (await refresh())} creates a new callback every render; pass a stable callback directly to avoid unnecessary child callback churn.

src/components/logging/useJmwalletdStdoutLog.ts
```diff
@@ -71,24 +76,19 @@ export function useJmwalletdStdoutLog({ enabled = true }: UseJmwalletdStdoutLogP
...
-  }, [logQuery.data])
+    return logQueryIsFetched || logQueryData !== undefined
+  }, [enabled, logQueryData, logQueryIsFetched, token])
 
   const refresh = useCallback(async () => {
-    if (token === undefined) return
-    await logQuery.refetch()
-  }, [logQuery, token])
+    if (enabled === false || token === undefined) return
+    return await logQueryRefetch()
```
parrth20: optional nit
refresh currently returns logQueryRefetch() result shape; make it explicitly Promise<void> (await logQueryRefetch()) so LogsContent can pass refresh directly without wrapper conversion

parrth20 (commented):
added review please otherwise LGTM

src/App.tsx
theborakompanioni: I think it is okay for this query. Maybe tackle it in a follow-up PR?

src/components/LogsContent.tsx
theborakompanioni: Done.

src/components/logging/useJmwalletdStdoutLog.ts
theborakompanioni: Done.

parrth20 (approved):

PR #1178, opened 2026-03-29
feat(create): verify  menominc step

theborakompanioni:
Adds a mnemonic verification step in the create wallet flow.
Additionally, it locks the wallet after creation, in order to not run into a session timeout.

## How to test
See the [storybook for `CreateStepVerifyMnemonic` component](http://localhost:6006/?path=/docs/page-create-createstepverifymnemonic--docs), or manually test the create flow in dev mode.

## :camera_flash: 

image
image

### DIFF
--- a/src/components/create/CreateStepVerifyMnemonic.tsx
+++ b/src/components/create/CreateStepVerifyMnemonic.tsx
@@ -1,4 +1,4 @@
-import { useCallback, useMemo, useState } from 'react'
+import { useCallback, useEffect, useMemo, useState } from 'react'
 import { useMutation } from '@tanstack/react-query'
 import { CheckCircle2Icon, ChevronLeftIcon } from 'lucide-react'
 import { useTranslation } from 'react-i18next'
@@ -7,6 +7,7 @@ import { isDebugFeatureEnabled } from '@/constants/debugFeatures'
 import { cn } from '@/lib/utils'
 import type { MnemonicPhrase } from '@/types/global'
 import { DevBadge } from '../dev/DevBadge'
+import { MaskedText } from '../ui/jam/MaskedText'
 import { Spinner } from '../ui/spinner'
 
 const skipWalletBackupVerification = isDebugFeatureEnabled('skipWalletBackupVerification')
@@ -45,7 +46,6 @@ export const CreateStepVerifyMnemonic = ({ mnemonicPhrase, onVerified, onBack }:
 
       if (word !== expectedWord) {
         setWrongButtonIndex(shuffledIndex)
-        setTimeout(() => setWrongButtonIndex(undefined), 600)
         return
       }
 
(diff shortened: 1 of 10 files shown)


src/components/create/CreateWalletPage.tsx
```diff
@@ -80,22 +132,33 @@ const CreateWalletPage = () => {
...
         throwOnError: true,
       })
 
+      // Step #3: lock wallet
+      // since wallet is immediately unlocked, let's lock the wallet here.
+      // The token is only a valid certain amount of time, and the user
+      // might take longer than the validitiy time.
+      // after the mnemonic phrase is confirmed, we unlock the wallet again.
+      await delayedPromise(210) // wait some time before locking
+      await lockWalletMutation.mutateAsync({
```
parrth20: If lockWalletMutation fails after wallet creation, control goes to the generic create-failure catch (line 177 (line 177)), so user sees “Failed to create wallet” even though wallet may already exist.

so i suggest handling lock failure separately (warn + continue to confirm) to avoid “created-but-reported-failed” flow.

src/components/create/CreateStepVerifyMnemonic.tsx
```diff
@@ -0,0 +1,200 @@
...
+    (word: string, shuffledIndex: number) => {
+      if (pickedIndicesSet.has(shuffledIndex)) return
+      if (wrongButtonIndex !== undefined) return
+
+      const nextPosition = selectedWords.length
+      const expectedWord = mnemonicPhrase[nextPosition]
+
+      if (word !== expectedWord) {
+        setWrongButtonIndex(shuffledIndex)
+        setTimeout(() => setWrongButtonIndex(undefined), 600)
```
parrth20: optional nit
setTimeout(() => setWrongButtonIndex(undefined), 600) has no cleanup; if component unmounts quickly, timeout can still fire.
ig we should store timeout id in a ref and clearing it in useEffect cleanup.

src/components/create/CreateWalletPage.tsx
```diff
@@ -80,22 +132,33 @@ const CreateWalletPage = () => {
...
       }
 
+      // Step #2: create wallet
       const walletFileName = walletDisplayNameToFileName(walletName)
-      const { data: createData } = await createwallet({
-        client,
+      const createData = await createWalletMutation.mutateAsync({
         body: {
-          walletname: walletFileName,
+          walletname: encodeURIComponent(walletFileName),
```
parrth20: optional nit
walletname in request body is encodeURIComponent(walletFileName) while body fields are usually plain values; path params are where encoding is needed.
safer way is to send raw walletFileName here to avoid future surprises if allowed chars expand

src/components/create/CreateWalletPage.tsx
theborakompanioni: Good catch! Done!

src/components/create/CreateWalletPage.tsx
theborakompanioni: You are absolutely correct (not at NIT at all!). Done!

src/components/create/CreateStepVerifyMnemonic.tsx
theborakompanioni: Done!

PR #1180, opened 2026-03-29
feat(send): ability to specify tx fee

theborakompanioni:
Adds the ability to specify a tx fee in send form for both direct and collaborative transactions.

Also, groups and simplifies the fee handling by adding type `JamFeeConfigValues` (Using sats/vbyte instead of sats/kilo-vybte).
Replaces component `TxFeeInputField` with `TxFeeForm` that has a yup schema and validation and can be used as nested element (via `useFormContext`). 

## :camera_flash: 

image

image


src/lib/feeConfig.ts
```diff
@@ -1,11 +1,62 @@
...
+      : TX_FEE_UNITS.BLOCKS
+
+  const txFeeInBlocks = txFeeUnit === TX_FEE_UNITS.BLOCKS ? txFeesInBlocksOrSatsPerKiloVByte : undefined
+  const txFeeInSatsPerVbyte =
+    txFeeUnit === TX_FEE_UNITS.SATS_PER_VBYTE ? txFeesInBlocksOrSatsPerKiloVByte! / 1_000 : undefined
+
+  return {
+    maxCjAbsoluteFee: Number.isSafeInteger(maxCjAbsoluteFee) ? maxCjAbsoluteFee : undefined,
+    maxCjRelativeFee: Number.isFinite(maxCjRelativeFee) ? maxCjRelativeFee : undefined,
+    txFeeFactor: Number.isFinite(maxCjRelativeFee) ? txFeeFactor : undefined,
```
parrth20: Tiny typo but important: this checks `Number.isFinite(maxCjRelativeFee)` for txFeeFactor. Shouldn’t this check `Number.isFinite(txFeeFactor)` instead? Right now valid tx_fees_factor can be dropped if max_cj_fee_rel is missing.

src/components/settings/FeeLimitDialog.tsx
```diff
@@ -56,54 +59,46 @@ export const FeeLimitDialog = ({ open, onOpenChange, walletFileName, ...dialogPr
...
         : undefined,
+      ...toTxFeeFormDefaultValues(feeConfigValues),
     }
   }, [feeConfigValues])
 
   const miningFeesForm = useForm<MiningFeesFormValues, unknown, MiningFeesFormValues>({
     mode: 'onChange',
     disabled: isSubmitting || isLoadingConfig,
-    values: miningFeeFormInitialValues,
+    defaultValues: miningFeeFormDefaultValues,
```
parrth20: umm can we avoid defaultValues here and sync with values/reset on config updates? Since this dialog stays mounted, these initial defaults may become stale after config refetch and show old values.

src/components/settings/FeeLimitDialog.tsx
```diff
@@ -56,54 +59,46 @@ export const FeeLimitDialog = ({ open, onOpenChange, walletFileName, ...dialogPr
...
+          ? factorToPercentage(feeConfigValues.maxCjRelativeFee)
+          : undefined,
     }
   }, [feeConfigValues])
 
   const collaboratorFeesForm = useForm<CollaboratorFeesFormValues, unknown, CollaboratorFeesFormValues>({
     mode: 'onChange',
     disabled: isSubmitting || isLoadingConfig,
-    values: collaboratorFeesFormInitialValues,
+    defaultValues: collaboratorFeesFormDefaultValues,
```
parrth20: same here

src/lib/feeConfig.test.ts
```diff
@@ -1,65 +1,81 @@
...
   it('should return false when both max fee values are present', () => {
-    const values: FeeConfigValues = {
+    const values = toJamFeeConfigValues({
       max_cj_fee_abs: '1500',
       max_cj_fee_rel: '0.00025',
       tx_fees: '3',
       tx_fees_factor: '0.2',
       max_sweep_fee_change: '0.8',
-    }
+    })
```
parrth20, line 163: we can we add one regression test for `tx_fees_factor` parsing independent of `max_cj_fee_rel`? That would lock the conversion logic and prevent this kind of typo from slipping in again.

src/components/send/TxFeeForm.tsx
```diff
@@ -0,0 +1,158 @@
...
+                    id="txFeeInSatsPerVbyte"
+                    {...register('txFee.txFeeInSatsPerVbyte', {
+                      disabled,
+                      valueAsNumber: true,
+                    })}
+                    type="number"
+                    min={MIN_TX_FEE_IN_SATS_PER_VBYTE}
+                    max={MAX_TX_FEE_IN_SATS_PER_VBYTE}
+                    step={1}
+                  />
```
parrth20: Optional Nit
small UX thought: schema allows decimal sats/vB (min 1.001), but input uses step={1}. Should we use step='any' (or 0.001) so decimal fee rates are easier to enter?

parrth20 (changes requested):
@theborakompanioni  looks good overall,  just one blocker in `feeConfig.ts` (txFeeFactor guard typo). Fix that and I’ll approve 💪🏻

parrth20 (commented):
Image

check this also

src/lib/feeConfig.ts
theborakompanioni: Nice catch! :muscle:

src/components/send/TxFeeForm.tsx
theborakompanioni: Done! (Step "0.01" feels more natural for the sats/vbyte)

src/components/settings/FeeLimitDialog.tsx
theborakompanioni: Yeah, I see what you mean. Reverted.

However, regardless of what we use, the form stays edited, even when the user never submitted the form.
We can add `key={"dialog-${open}"}` to force a re-render of the dialog.. or reset the form on close.. both feels very brittle and awkward.. to you have an idea? (we'll just leave it at that for now - reset button is now always visible for all users as a workaround).

src/lib/feeConfig.test.ts
theborakompanioni: Done.

src/components/settings/FeeLimitDialog.tsx
theborakompanioni: Done.

src/components/send/TxFeeForm.tsx
```diff
@@ -0,0 +1,159 @@
...
+                <InputGroup>
+                  <InputGroupInput
+                    id="txFeeInSatsPerVbyte"
+                    {...register('txFee.txFeeInSatsPerVbyte', {
+                      disabled,
+                      valueAsNumber: true,
+                    })}
+                    type="number"
+                    min={Math.floor(MIN_TX_FEE_IN_SATS_PER_VBYTE)}
+                    max={MAX_TX_FEE_IN_SATS_PER_VBYTE}
```
parrth20, line 119: the schema minimum for sats is 1.001, but the input min becomes 1, so 1 looks allowed in the UI and then Yup rejects it, could we align this with the schema, maybe min={MIN_TX_FEE_IN_SATS_PER_VBYTE} and a matching step?

src/components/send/SendPage.tsx
```diff
@@ -68,6 +72,8 @@ export const SendPage = ({ walletFileName }: SendPageProps) => {
   const { refetch: refetchWalletInfo, waitForUtxosToBeSpent, setWaitForUtxosToBeSpent } = useJamWalletInfoContext()
   const jmSession = useStore(jmSessionStore, (state) => state.state)
   const { enabled: isDeveloperMode } = useDeveloperMode()
+
+  const feeConfigValidation = useFeeConfigValidation({ walletFileName })
```
parrth20, line 76: optional nit
calls `useFeeConfigValidation` separately from the later destructuring, I’d suggest to reuse one object it'll make code more cleaner

parrth20 (dismissed):
Otherwise LGTM !! 🚀

src/components/send/TxFeeForm.tsx
theborakompanioni: This was deliberately put in there, as the step with "any" will always append `.001` but step 1 (e.g. from `1.0001` to `2.0001`, at least on firefox. Does the same happen for you?
Having a warning that the input is invalid and not submitting the form is the same, as if we have min=`1.001` and the user manually providing 1.. no?

This felt better than having a step of `0.001` which does not feel useful when manipulating the fee.

Let's think about whether to remove the min/max altogether, and rely on the validation, I guess that would be better in general.

src/components/send/SendPage.tsx
theborakompanioni: Good catch! Slipped through. Done!

src/components/send/TxFeeForm.tsx
parrth20: yes we can remove the min/max completely that would be better, but this can be done in follow-ups we can merge this as of now

parrth20 (dismissed):

PR #1181, opened 2026-03-30
feat(send): utxo selection in send with address-grouped freeze/unfreeze

parrth20:
this PR adds coin-control style utxo selection to the send flow, so you can choose which utxos are eligible before sending. selected utxos remain spendable, unselected ones are frozen, and utxos on the same address are automatically grouped to preserve privacy. it also shows a toast when selection is auto-adjusted and includes a small reusable update to the utxo table to support default selection and selection rules.

ping @theborakompanioni

### DIFF
--- a/src/components/send/SendForm.tsx
+++ b/src/components/send/SendForm.tsx
@@ -26,6 +26,7 @@ import type { Utxo } from '@/hooks/useQueryUtxos'
 import type { FeeConfigValues } from '@/hooks/useFeeConfigValidation'
 import type { BalanceSummary } from '@/lib/balanceSummary'
 import { parseBip21Uri, type Bip21ParseResult } from '@/lib/bip21'
+import * as fb from '@/lib/fidelityBondUtils'
 import { utxoTags } from '@/lib/tags'
 import {
   cn,
@@ -385,7 +386,7 @@ export function SendForm({
 
   const defaultUtxoRowSelection = useMemo<RowSelectionState>(() => {
     return (sourceJar?.utxos || []).reduce((acc, utxo) => {
-      if (utxo.frozen === false && utxo.locktime === undefined) {
+      if (utxo.frozen === false && !fb.utxo.isFidelityBond(utxo)) {
         acc[utxo.utxo] = true
       }
       return acc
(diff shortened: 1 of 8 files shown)


src/components/send/SendForm.tsx
```diff
@@ -356,6 +381,138 @@ export function SendForm({
...
+    const utxosToUnfreeze = mutableUtxos.filter((it) => selectedAddresses.has(it.address) && it.frozen === true)
+
+    if (utxosToFreeze.length === 0 && utxosToUnfreeze.length === 0) {
+      setShowUtxoSelectorDialog(false)
+      return
+    }
+
+    try {
+      const result = await applyUtxoSelectionMutateAsync({ utxosToFreeze, utxosToUnfreeze })
+      await refetchWalletInfo()
```
theborakompanioni: Nit: I dont think the wallet info needs to be reloaded, since we are just freezing/unfreezing, right? Keep in mind that reloading might take an awful amount of time on raspis controlled over tor. What do you think?

theborakompanioni (changes requested):
Generally like it a lot!

One request would be to extract the dialog into an own component to keep it outside the `SendForm`, would that be possible? Maybe just pass a callback, e.g. `onJarClicked`. What do you think?

src/components/send/SendForm.tsx
```diff
@@ -356,6 +381,138 @@ export function SendForm({
...
+    setShowUtxoSelectorDialog(true)
+  }, [defaultUtxoRowSelection, sourceJar])
+
+  const onApplyUtxoSelection = useCallback(async () => {
+    if (!sourceJar) return
+
+    // Keep same-address UTXOs together to avoid accidental privacy leaks.
+    const selectedUtxoIds = new Set(selectedSourceJarUtxos.map((it) => it.utxo))
+    const selectedAddresses = new Set(selectedSourceJarUtxos.map((it) => it.address))
+    const mutableUtxos = sourceJar.utxos.filter((it) => it.locktime === undefined)
```
theborakompanioni: NIT: there is an `fb.isFidelityBond` utility method

src/components/send/SendForm.tsx
```diff
@@ -472,7 +697,11 @@ export function SendForm({
                       void trigger('destination.address')
                     }
                   }}
-                  disabled={disabled || jar.balanceSummary.calculatedAvailableBalanceInSats <= 0}
+                  disabled={
```
theborakompanioni: Should we also display the frozen funds here?
We can select a Jar and then freeze all the utxos in there <- it would still be selected an sending will fail (expected), but maybe indicate to the user that all funds are frozen now?

Image

e.g. here you see that it seems the selected jar has funds available... (when in fact, all utxos are frozen).

parrth20:
> > One request would be to extract the dialog into an own component to keep it outside the `SendForm`, would that be possible? Maybe just pass a callback, e.g. `onJarClicked`. What do you think?
> 
 Okay I've just refactored it !!

parrth20:
ping @theborakompanioni

src/components/ui/jam/SelectableJar.tsx
```diff
@@ -42,8 +44,15 @@ export const SelectableJar = ({
           disabled={disabled}
         />
         <span className="text-xs">{name}</span>
-        <div className="flex items-center text-sm">
-          <Balance valueString={String(totalBalance)} />
+        <div className="flex min-w-[110px] items-center justify-center text-sm">
```
theborakompanioni: In this case.. can we just `Jar` component here?

parrth20: I spotted one subtle thing, JarIcon selection highlight would be lost if we use Jar directly, so I’m adding an optional isSelected prop to Jar and forwarding it to keep current behavior unchanged.

theborakompanioni (approved):

PR #1183, opened 2026-04-03
chore(ui): use legacy jar icons

parrth20:
ping @theborakompanioni


theborakompanioni (approved):
Approved. 🚀 

Nit: Just because they are from the legacy branch, you don't have to name them LegacyJar ; -)

PR #1192, opened 2026-04-05
fix: avoid crash when direct-send output lookup misses

### LINKED ISSUE #1188: Unchecked `outputs.find()` result crashes Send success handler
In `src/components/Send/index.tsx`, `sendPayment` dereferences the return value of `outputs.find(...)` without a null check. When the destination address doesn't match any entry in `txinfo.outputs` (e.g. address normalisation differences, send-to-self, API changes), `find()` returns `undefined` and `.value_sats` / `.address` throw a `TypeError`.

The error is caught by the outer `catch`, so the user sees a **danger alert despite the payment having succeeded on-chain**. Worse, `setWaitForUtxosToBeSpent` is never reached ,leaving the UI stuck in a loading state until manual refresh.

## Relevant code

```ts
// src/components/Send/index.tsx — inside sendPayment, after res.ok
const output = outputs.find((o: any) => o.address === destination)
// output is never null-checked
setPaymentSuccessfulInfoAlert({
  variant: 'success',
  message: t('send.alert_payment_successful', {
    amount: output.value_sats,   // throws if output === undefined
    address: output.address,
    txid,
  }),
})
setWaitForUtxosToBeSpent(inputs.map((it: any) => it.outpoint))  // never reached on crash
```

## Proposed fix

Guard the `find()` result. If the output can't be located, still mark success and kick off the UTXO poller — the payment already went through:

```ts
const output = outputs.find((o: any) => o.address === destination)

if (!output) {
  setWaitForUtxosToBeSpent(inputs.map((it: any) => it.outpoint))
  success = true
  return success
}

setPaymentSuccessfulInfoAlert({
  variant: 'success',
  message: t('send.alert_payment_successful', {
    amount: output.value_sats,
    address: output.address,
    txid,
  }),
})
setWaitForUtxosToBeSpent(inputs.map((it: any) => it.outpoint))
success = true
```

saschabuehrle:
Fixes #1188

The success path assumed outputs.find(...) always returns a match and dereferenced it immediately. When no exact match is found, that raised a TypeError after the transaction succeeded.

This keeps the success flow alive by falling back to the known destination and amount for the alert message, then continues with the UTXO wait tracking as before.

Greetings, saschabuehrle

### DIFF
--- a/src/components/Send/index.tsx
+++ b/src/components/Send/index.tsx
@@ -262,11 +262,12 @@ export default function Send({ wallet }: SendProps) {
           txinfo: { outputs, inputs, txid },
         } = await res.json()
         const output = outputs.find((o: any) => o.address === destination)
+        const successfulOutput = output ?? { value_sats: amountSats, address: destination }
         setPaymentSuccessfulInfoAlert({
           variant: 'success',
           message: t('send.alert_payment_successful', {
-            amount: output.value_sats,
-            address: output.address,
+            amount: successfulOutput.value_sats,
+            address: successfulOutput.address,
             txid,
           }),
         })


theborakompanioni:
Hey @saschabuehrle Somewhat clever fix. However, that is not something I'd expect if my expected output is missing from the transaction. 

Note: All work is done in the `v2` branch (which we soon merge into `devel`).

PR #1196, opened 2026-04-07
feat: forbid sweep with direct send

### LINKED ISSUE #810: Forbid a "direct send" when a jar is "swept" or vice versa
to not hurt themselves, the user shouldn't be able to sweep a jar AND combine it with a "direct send" aka deactivated coinjoin send.

- so either its "I want to sweep" then the toggle stayes ON but "deactivated" to toggle off
- or its I want to send a direct send, so the user turns off the toggle but then the "sweep" button gets also deactivated on the amount input

kanishka0411:
Fixes #810 

Disables the sweep button when direct send is active, and disables the CoinJoin toggle when sweep is active. Tooltip hints explain why each control is disabled. A form validation check also blocks the invalid combo at submit time.

### DIFF
--- a/src/components/Send/AmountInputField.tsx
+++ b/src/components/Send/AmountInputField.tsx
@@ -35,6 +35,9 @@ export const AmountInputField = ({
   const form = useFormikContext<any>()
   const ref = useRef<HTMLInputElement>(null)
 
+  const isCoinJoin = form.values.isCoinJoin !== false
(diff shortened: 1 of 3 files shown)


parrth20:
@kanishka0411 we are not working on legacy branch !
if you want to make any contributions do it in v2 branch as of now !!

Issue #1079, opened 2026-02-15
v2: Parity to Legacy 


parrth20:
- [ ] bug: log file contents not shown
- [x] feat: improve log file page (search functionality, modal and page should work
- [x] feat: validate mnemonic phrase on import page
- [x] feat: confirm user wrote down mnemonic phrase
- [x] feat: earn report
  - all functionality from legacy
  - display useful charts or visualizations (?)
- [x] chore: improve orderbook summary (use `Balance` component)
  - display useful charts or visualizations (?) 
- [x] feat(import): ability to provide import options (blockheight, address import limit "gap limit")
- [ ] feat: send collaborative transactions (send)
  - with custom transaction fee (if supported by backend version!) 
- [ ] feat: scheduler functionality (sweep)
- [x] feat: "utxo selection" in send (aka "quick freeze/unfreeze" / "coin control")
- [x] chore: after (or before) a scheduler run: "To run the scheduler you need UTXOs with 5 or more confirmations. Fund your wallet and wait for 5 more blocks." message should disappear after 5 blocks have passed
- [x] feat: scheduler show more information, e.g. 
  - info about schedule (e.g. entries and their states)
  - display wait time between transactions
  - current state, e.g. "waiting $time before creating and broasting next transaction", "waiting for transaction to be confirmed", "tranaction $txid confirmed", etc.
- [x] bug: refresh token after wallet is created on confirm page (bug: if a user stays longer than the token validity time on the confirm page, next screens will show errors)
- [ ] refactor: all forms should use `react-hook-form` and `yup`
- [x] chore: use jar icons from legacy version
- [ ] test: Storybook stories for all core components
- [ ] test: tests for all utility functions
- [ ] refactor: adapt and rethink current i18n keys
- [ ] refactor: remove all unnecessary stuff from the codebase -> keep it simple
- [x] refactor: fee config form
- [ ] refactor: use tanstack table in Earn Report
- [x] refactor: Fullscreen Earn Report
- [x] chore: Earn Report as Page and Overlay (add link to side menu)
- [ ] fix: fidelity bond flow can be improved (use balance, jar names, react-hook-form and yup schema, etc.)
- [ ] fix(earn): try prevent "no confirmed coins" on earn page
- [x] feat: chain rescan after wallet import till balance is found
- [ ] fix: refresh "The scheduler requires your UTXOs to have 5 or more confirmations. Wait for X more blocks." automatically.

[...]

PR #1197, opened 2026-04-08
[v2] refactor(earn): to use tanstack table state in earn reports table

kishore08-07:
This PR resolves a task(refactor: use tanstack table in Earn Report) in #1079 .

The EarnReportContent table currently manages pagination, filtering, and state in a hybrid manner where manual react state coexisting with tanstack’s internal state. This causes dual-state synchronization issues and unnecessary re-renders.

**Changes made:**
- Refactored to let tanstack table fully control pagination, filtering, and derived values by eliminating manual pagination and filter state (currentPage, itemsPerPage)
- Wired onPaginationChange and onGlobalFilterChange handlers to useReactTable in tanstack
- Replaced manual calculations(totalPages,totalItems) with table.getPageCount() and table.getFilteredRowModel().rows.length.


kishore08-07:
Hi @theborakompanioni 👋

I have refactored the earn report table to fully use TanStack state.
Kindly requesting a review whenever you have time, thanks in advance for your time 🙏

src/components/earn/report/EarnReportContent.tsx
```diff
@@ -304,20 +292,19 @@ export const EarnReportContent = ({ className, enabled }: EarnReportContentProps
...
+                }
+                totalItems={table.getFilteredRowModel().rows.length}
+                onPageChange={(page) => table.setPageIndex(page - 1)}
                 onItemsPerPageChange={(newItemsPerPage) => {
-                  setItemsPerPage(newItemsPerPage)
-                  const size =
-                    newItemsPerPage === -1 ? table.getPrePaginationRowModel().rows.length || 1 : newItemsPerPage
-                  table.setPageSize(size)
-                  setCurrentPage(1)
+                  if (newItemsPerPage === -1) {
```
parrth20, line 308: this refactor loses the old “Show all” behaviour when the dataset changes. before, itemsPerPage === -1 was the source of truth, so “show all” kept tracking allEntries.length after a refresh or after adding demo rows. In the PR, pagination.pageSize stores a concrete row count instead, so after selecting “Show all”, any new rows added later spill onto extra pages instead of staying visible on the single “all” page.

src/components/earn/report/EarnReportContent.tsx
```diff
@@ -304,20 +292,19 @@ export const EarnReportContent = ({ className, enabled }: EarnReportContentProps
...
-                totalPages={totalPages}
-                itemsPerPage={itemsPerPage}
-                totalItems={allEntries.length}
-                onPageChange={(page) => {
-                  setCurrentPage(page)
-                  table.setPageIndex(Math.max(0, page - 1))
-                }}
+                currentPage={table.getState().pagination.pageIndex + 1}
+                totalPages={table.getPageCount()}
+                itemsPerPage={
```
parrth20: itemsPerPage={pagination.pageSize === allEntries.length ? -1 : pagination.pageSize} conflates a normal page size with “Show all”. If the table has exactly 25, 50, or 100 rows, the UI will report “Show all” even when the user is just on one of the standard page sizes. That loses the distinction between explicit user choice and a coincidental row count.

parrth20 (changes requested):
why did you switch from an explicit -1 ‘Show all’ state to inferring it from pageSize?

src/components/earn/report/EarnReportContent.tsx
kishore08-07: since tanstack Table only supports { pageIndex, pageSize }, there’s no place for a sentinel like -1. So my approach was to translate -1 into allEntries.length and pass that as pageSize. At the moment of selecting “Show all”, this works since it displays everything.

now i understand the issue, as i treated that value as persistent intent, but it’s really just a snapshot. When new rows are added (refresh/demo data), pageSize remains the old count, and the new rows overflow onto another page. The previous -1 worked because it forced recalculation on every render.

Approach to fix this: Introduce an isShowAll boolean alongside tanstack’s pagination. On every allEntries change, a useEffect re-syncs table.setPageSize(allEntries.length) when isShowAll is true. This way, isShowAll captures persistent intent, while pageSize remains a derived value for rendering.

src/components/earn/report/EarnReportContent.tsx
kishore08-07: Agreed, this was an incorrect inference.
After removing -1, I tried to infer “Show all” by checking pageSize === allEntries.length, but that conflates user intent with a coincidental dataset size. And yes, if the table has exactly 50 rows, selecting page size 50 would incorrectly appear as “Show all”.

Fix: Instead of inferring, I’ll use the same isShowAll flag.

kishore08-07:
@parrth20 I have addressed the feedback and fixed the “Show all” behavior using an explicit isShowAll state.
Please re-review when you have time, thanks 🤝

parrth20:
LGTM @kishore08-07 
thanks 🤝

parrth20 (approved):

PR #1198, opened 2026-04-08
fix(receive): sync jar selection with async loading

23bsm038-Mahi:
#### Problem

`selectedSourceJarIndex` was initialized before jars were loaded asynchronously.
This could result in:

* invalid `mixdepth` (undefined)
* inconsistent UI (QR, badge, selected jar)

#### Solution

* Sync jar selection when jars update
* Use `jarIndex` instead of array index
* Guard address generation until a valid jar exists
* Keep form defaults in sync with selected jar

#### Impact

* Prevents invalid API calls
* Keeps UI consistent after async wallet load

#### Notes

This change is isolated to the receive module and avoids conflicts with ongoing PRs.

### DIFF
--- a/src/components/receive/ReceiveForm.tsx
+++ b/src/components/receive/ReceiveForm.tsx
@@ -1,4 +1,4 @@
-import { useMemo } from 'react'
+import { useEffect, useMemo } from 'react'
 import { yupResolver } from '@hookform/resolvers/yup'
 import type { TFunction } from 'i18next'
 import { useForm, useWatch, type Resolver, type SubmitHandler } from 'react-hook-form'
@@ -76,6 +76,7 @@ export const ReceiveForm = ({ className, defaultValues, onSubmit, jars, disabled
     handleSubmit,
     formState: { errors, isSubmitting, isValid },
     setValue,
+    reset,
   } = useForm<ReceiveFormValues, unknown, ReceiveFormValues>({
     mode: 'onChange',
     defaultValues,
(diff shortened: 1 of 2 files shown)


23bsm038-Mahi:
Hi @theborakompanioni 
I’ve fixed the async jar selection issue in the receive flow to ensure consistent state and prevent invalid selections.
Would really appreciate your review whenever you have time.  
Thanks in advance!

parrth20:
@23bsm038-Mahi lint errors !!

parrth20:
@23bsm038-Mahi thanks for contribution !!! but the core premise of the fix does not look valid for this receive flow. In ` JamWalletInfoContextProvider.tsx` , jars 0..4 are pre-seeded before async wallet data finishes loading, so the receive page already starts with a valid jar instead of an “undefined mixdepth” state because of that, the extra resync logic in `ReceivePage.tsx` (line 35) and the form reset loop in `ReceiveForm.tsx` (line 73) look unnecessary rather than a required bug fix.

parrth20 (changes requested):
I’d prefer to keep the current behaviour unless there’s a concrete repro showing an actual failure case

23bsm038-Mahi:
Thanks for the review!
I understand your point — since the jars are already pre-seeded and there’s no clear repro of an actual issue, this change does add unnecessary complexity. I’ll go ahead and close/simplify this PR and fix the lint errors separately if needed.
Appreciate the feedback !

parrth20:
@23bsm038-Mahi just wanted to tell you if you're replying don't just copy paste from AI and do, if you have some knowledge about it then do things(atleast when we take our time and give a review).

23bsm038-Mahi:
I got your point. I will make sure to explain my reasoning more clearly in my own words in future contributions. Thanks for your time and review. 

> @23bsm038-Mahi just wanted to tell you if you're replying don't just copy paste from AI and do, if you have some knowledge about it then do things(atleast when we take our time and give a review).

PR #1199, opened 2026-04-08
fix(receive): ignore share cancellation to prevent false error toast

23bsm038-Mahi:
# Fix: ignore share cancellation in receive flow

# Problem

Cancelling the native share dialog was incorrectly treated as an error, triggering an unnecessary error toast.

# Solution

* Detect cancellation using `AbortError` / `NotAllowedError`
* Ignore these cases as expected user behavior
* Preserve error handling for genuine failures

#Impact

* Improves UX by preventing false error messages
* Aligns behavior with native share expectations

# Scope

* Minimal change limited to `ReceivePage.tsx`
* No impact on other flows

# Notes

This change avoids conflicts with ongoing PRs by targeting an isolated part of the receive module.

### DIFF
--- a/src/components/receive/ReceivePage.tsx
+++ b/src/components/receive/ReceivePage.tsx
@@ -100,7 +100,13 @@ export const ReceivePage = ({ walletFileName }: ReceivePageProps) => {
           title: 'Bitcoin Address',
           text: bitcoinAddress,
         })
-        .catch(() => {
+        .catch((error: unknown) => {
+          if (
+            error instanceof DOMException
+            && (error.name === 'AbortError' || error.name === 'NotAllowedError')
+          ) {
+            return
+          }
           toast.error(t('receive.error_share_address_failed'))
         })
     } else {


23bsm038-Mahi:
Handles expected cancellation behavior of navigator.share() to improve UX consistency.

theborakompanioni:
Hey @23bsm038-Mahi I am interested in why you think the toast should not be shown.
According to https://developer.mozilla.org/en-US/docs/Web/API/Navigator/share#exceptions
a `NotAllowedError` is also thrown "if being blocked due to security considerations" and `AbortError` when "there are no share targets available.". Wouldn't a user want to be informed about those states?

23bsm038-Mahi:
@theborakompanioni Good point!
You're right these errors can represent real failures. In practice though, they’re often triggered when the user simply dismisses the share dialog, and showing an error toast in that case can feel misleading from a UX perspective.
The intention here was to avoid false error feedback for user-initiated cancellations, while still preserving error handling for genuine failures.

theborakompanioni:
Thanks @23bsm038-Mahi, but we are keeping the current behaviour for now. Let's do it after v2 is released and people report it as confusing/error/etc. Approach okay for you?

23bsm038-Mahi:
> Thanks @23bsm038-Mahi, but we are keeping the current behaviour for now. Let's do it after v2 is released and people report it as confusing/error/etc. Approach okay for you?
okay, got it. we can look into this after v2 if users face issues.

PR #1202, opened 2026-04-10
fix(utils): make BTC and sats formatting deterministic across locales

Ashish-Kumar-Dash:
Summary
This PR aligns formatting behavior with expected runtime semantics and stabilizes tests across environments.

What changed
1. Kept production formatting locale behavior unchanged (runtime/host locale is still used).
2. Added a test-only default number locale in vitest setup so baseline tests are deterministic.
3. Added explicit locale-variance coverage (en-IN) to verify formatting follows runtime locale when locale changes.

Why
The prior issue appeared as cross-locale test instability.  
This fix avoids forcing en-US for all users in production while still making CI and local tests reliable.

Validation
1. Unit tests passing: 186/186.
2. Build passing.
3. Lint passing with only pre-existing warnings (no new lint errors).

Scope
1. utils.ts
2. vitest.setup.ts
3. utils.test.ts

### DIFF
--- a/vitest.setup.ts
+++ b/vitest.setup.ts
@@ -2,14 +2,6 @@ import * as matchers from '@testing-library/jest-dom/matchers'
 import { cleanup } from '@testing-library/react'
 import { expect, afterEach } from 'vitest'
 
-const TEST_DEFAULT_NUMBER_LOCALE = 'en-US'
-const originalNumberToLocaleString = Number.prototype.toLocaleString
-
-Number.prototype.toLocaleString = function (...args: Parameters<typeof Number.prototype.toLocaleString>) {
-  const [locales, options] = args
-  return originalNumberToLocaleString.call(this, locales ?? TEST_DEFAULT_NUMBER_LOCALE, options)
-}
-
 expect.extend(matchers)
 
 afterEach(() => {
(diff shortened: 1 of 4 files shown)


vitest.setup.ts
```diff
@@ -2,6 +2,14 @@ import * as matchers from '@testing-library/jest-dom/matchers'
...
 import { expect, afterEach } from 'vitest'
 
+const TEST_DEFAULT_NUMBER_LOCALE = 'en-US'
+const originalNumberToLocaleString = Number.prototype.toLocaleString
+
+Number.prototype.toLocaleString = function (...args: Parameters<typeof Number.prototype.toLocaleString>) {
+  const [locales, options] = args
+  return originalNumberToLocaleString.call(this, locales ?? TEST_DEFAULT_NUMBER_LOCALE, options)
+}
+
```
theborakompanioni: I guess this can be replaced by adding the following to `vitest.config.ts`:
```patch
diff --git a/vitest.config.ts b/vitest.config.ts
index a0a4fef..a29010a 100644
--- a/vitest.config.ts
+++ b/vitest.config.ts
@@ -8,6 +8,9 @@ import viteConfig from './vite.config'
 export default defineConfig((args: ConfigEnv): ViteUserConfig => {
   return mergeConfig(viteConfig(args), {
     test: {
+      env: {
+        LC_ALL: 'en_US.UTF-8', // set default browser locale in tests
+      },
       projects: [
         {
           extends: true,
```

PR #1205, opened 2026-04-13
refactor(settings): use react-hook-form for fee settings forms

parrth20:
summary
migrate the collaborator fee settings form to `react-hook-form` and `yup`
migrate the mining fee settings form to `react-hook-form` and `yup`
keep the existing `FeeLimitDialog` integration and simplify tx fee input callbacks
ping @theborakompanioni

### DIFF
--- a/src/components/settings/CollaboratorFeesForm.tsx
+++ b/src/components/settings/CollaboratorFeesForm.tsx
@@ -6,7 +6,7 @@ import { useTranslation } from 'react-i18next'
 import * as yup from 'yup'
 import { Alert, AlertDescription } from '@/components/ui/alert'
 import { CJ_FEE_ABS_MIN, CJ_FEE_ABS_MAX, CJ_FEE_REL_MIN, CJ_FEE_REL_MAX } from '@/constants/jam'
-import { isValidNumber, factorToPercentage, percentageToFactor, formatSats } from '@/lib/utils'
+import { factorToPercentage, percentageToFactor, formatSats } from '@/lib/utils'
 import { Field, FieldDescription, FieldLabel } from '../ui/field'
 import { InputGroup, InputGroupAddon, InputGroupInput } from '../ui/input-group'
 import { SatSymbol } from '../ui/jam/CurrencySymbol'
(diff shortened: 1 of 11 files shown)


src/components/settings/CollaboratorFeesForm.tsx
```diff
@@ -15,92 +18,125 @@ interface CollaboratorFeesFormProps {
...
+      })
 
-      if (!maxCjFeeRel) {
-        newErrors.maxCjFeeRel = t('settings.fees.feedback_invalid_max_cj_fee_rel', {
-          min: factorToPercentage(CJ_FEE_REL_MIN),
-          max: factorToPercentage(CJ_FEE_REL_MAX),
+      return yup
+        .object({
+          maxCjFeeAbs: yup
+            .string()
```
theborakompanioni: Cane we use number validation here?
Striving for consistency in the forms: With  validation messages for `min`, `max`, `required`,etc. and transform to null if not a safe/finite number)

Similar to e.g. `EarnForm`:
```ts
      offerAbsoluteFee: yup
        .number()
        .integer(t('earn.feedback_invalid_abs_fee'))
        .transform((value) => (Number.isSafeInteger(value) ? Number(value) : null))
        .when('offerType', {
          is: (val: OfferType) => val === OFFERTYPE_ABS,
          then: (schema) =>
            schema
              .min(JAM.OFFER_FEE_ABS_MIN, t('earn.feedback_invalid_abs_fee'))
              .required(t('earn.feedback_invalid_abs_fee')),
          otherwise: (schema) => schema.nullable().optional(),
        }),
```

What do you think?

src/components/settings/CollaboratorFeesForm.tsx
```diff
@@ -1,6 +1,9 @@
-import { useState, useEffect, forwardRef, useImperativeHandle, useCallback } from 'react'
+import { useEffect, forwardRef, useImperativeHandle, useMemo } from 'react'
```
theborakompanioni: We should get rid of `forwardRef` and `useImperativeHand` once and for all.
But this can be done in a follow-up PR.

parrth20: Sure will address this in a follow-up PR.

src/components/settings/CollaboratorFeesForm.tsx
parrth20: umm yess, It's a good catch, I've kept it string-based mainly to stay conservative in this refactor, but number validation would definitely read better and match the existing form patterns more closely.

theborakompanioni (approved):
LGTM

parrth20:
LGTM as well we can merge this!!

PR #1206, opened 2026-04-14
feat: add data-journey-state with tests, docs, and friction audit

23bsm038-Mahi:
What:
Added a data-journey-state attribute to the main wallet container.

How:
Used a small helper to derive the state based on wallet balance and fee configuration (loading, empty, needs_fee, ready). Also handled a few edge cases like invalid or negative balance and added unit + component tests.

Why:
The goal is to make the UI a bit more aware of the wallet state so it can guide users better.

Also:
Added a short wallet state friction audit and updated the developer docs with how to simulate different journey states during testing.

Note:
Some unrelated tests fail locally due to locale formatting and Playwright setup, but tests related to this change pass.

### DIFF
--- a/docs/developing.md
+++ b/docs/developing.md
@@ -126,3 +126,12 @@ npm install && npm run dev
 - yup (GitHub): https://github.com/jquense/yup
 - react-hook-form (GitHub): https://github.com/react-hook-form/react-hook-form
 - react-query (GitHub): https://github.com/TanStack/query
+
+## Simulating Wallet Journey States
+
+The main wallet page exposes its current state with `data-journey-state` on `data-testid="main-wallet"`. This is useful when checking UI behavior in tests or during development.
+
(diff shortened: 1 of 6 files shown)


23bsm038-Mahi:
Hi @theborakompanioni, I added a data-journey-state attribute to the main wallet container based on wallet balance and fee config.Used a small helper to decide the state (loading, empty, needs_fee, ready) and handled basic edge cases. Also added some unit and component tests to check it works as expected. Idea is to make the UI more aware of the wallet state so we can guide users better later.
Let me know if this should be placed somewhere else or if anything should be changed.

23bsm038-Mahi:
Hi, I’ve also added the wallet state friction audit and updated the developer docs with a section on simulating journey states.
Please let me know if anything should be improved or adjusted.

PR #1207, opened 2026-04-15
chore: remove switch wallet page

theborakompanioni:
Removes the `SwitchWalletPage` component as it is unfinished and not really needed.
If a user wants so switch a wallet, he can lock or "log out" of the current wallet and choose a different one.


parrth20:
@theborakompanioni looks mostly clean and good , but one leftover from the switchwallet page removal is still present,

the English locale removes `wallets.alert_wallet_open` and `settings.button_switch_wallet`, but the same switch-wallet-only keys still exist in the other locale files. Since the page, route, and settings link are gone, can you remove those keys from `de`, `fr`, `it`, `pt_BR`, `ru`, `zh_Hans`, and `zh_Hant` too?

if you want me i can also commit that changes ?

parrth20 (approved):

parrth20:
@theborakompanioni or we can make a follow-up for this what you think?

theborakompanioni:
> @theborakompanioni or we can make a follow-up for this what you think?

Yes, we need to cleanup i18n handling from the ground up anyway. But lets first focus on features, unused i18n keys is a concern in the cleanup phase before releasing v2.

PR #1209, opened 2026-04-15
fix(v2): remove manual walletname encoding and rely on generated client path serialization 

Ashish-Kumar-Dash:
## Summary
This PR removes manual walletname URL encoding at generated-client callsites and standardizes on passing raw walletname values.

## Why
The generated API client already encodes path params internally. Pre-encoding at callsites can cause double-encoding edge cases, especially for wallets created outside Jam.

## What changed
- Replaced manual walletname encoding with raw walletname in generated-client path usage.
- Kept behavior consistent with the existing raw walletname flow.

## Validation
- Lint: passes (only existing warnings, no new ones).
- Build: passes.
- Targeted unit tests for touched send/sweep logic: pass.

## Notes
No functional API shape changes were introduced; this is a consistency and safety cleanup.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -93,7 +93,7 @@ function App() {
     {
       ...lockwalletOptions({
         client,
-        path: { walletname: encodeURIComponent(walletFileName || '') },
+        path: { walletname: walletFileName || '' },
       }),
       enabled: false,
     },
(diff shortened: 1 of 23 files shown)


parrth20 (approved):

parrth20:
@Ashish-Kumar-Dash LGTM thanks! 
but i think we could we add a small regression test for a wallet name containing a space or `%` to lock this behavior down? since this fix relies on the generated path serializer encoding the wallet name exactly once, a test would help make sure we don’t accidentally reintroduce manual pre-encoding later.
what are your thoughts ???

Ashish-Kumar-Dash:
Hey! I agree & have added a small regression unit test, do check it out

parrth20 (changes requested):
lint errors !!

Ashish-Kumar-Dash:
yup yup I was fixing them only, didn't expect for a review to be this fast :)

theborakompanioni (approved):

PR #1210, opened 2026-04-15
chore: fix features path and rename jam api lib file

theborakompanioni:
Fixes the path to the "features" endpoint and renames the `api/logs.ts` file to `api/jam.ts`.
See https://github.com/joinmarket-webui/jam-docker/blob/master/standalone/nginx/default.conf#L135


parrth20 (approved):

parrth20:
> Fixes the path to the "features" endpoint and renames the `api/logs.ts` file to `api/jam.ts`. See https://github.com/joinmarket-webui/jam-docker/blob/master/standalone/nginx/default.conf#L135

I've checked this against the jam-docker nginx route,
LGTM thanks !

PR #1214, opened 2026-04-16
Fix infinite loading in dev mode

### LINKED ISSUE #1213: Keeps loading until i hit refresh after selecting a wallet
**Expected behavior**

It should finish loading after some time.

**Actual behavior**

It keeps loading until manually hit refresh.

1. log off of your existing wallet
2. sign in again.

theAnuragMishra:
## What does this PR do?
This pr fixes the bug which causes loading state to stay forever when run in the dev mode. 

changes: replaced `useQuery` with `useEffect`s becasue the latter is better suited for side effects, and replaced `useState` with `useRef` for trackign previous rescanning.

- Fixes #1213 

#### Video Demo (if applicable):

https://github.com/user-attachments/assets/16e6c9cd-2315-40d8-bcc9-815cbe024d05

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import { lazy, Suspense, useEffect, useMemo, useState } from 'react'
+import { lazy, Suspense, useEffect, useMemo, useRef, useState } from 'react'
 import type { PropsWithChildren } from 'react'
 import { lockwalletOptions } from '@joinmarket-webui/joinmarket-api-ts/@tanstack/react-query'
 import { token } from '@joinmarket-webui/joinmarket-api-ts/jm'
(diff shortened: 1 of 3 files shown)


theborakompanioni:
Hmm.. I do not really like the fix for no specific reason, probably because of using refs.
Maybe the component does not really make sense here and should be moved into a context entirely.
Also.. locally what seems to have fixed the problem is [consuming the `AbortSignal`](https://tanstack.com/query/v5/docs/framework/react/guides/query-cancellation)... from the docs:
> By default, queries that unmount or become unused before their promises are resolved are not cancelled.  [...] However, if you consume the AbortSignal, the Promise will be cancelled

What do you think?

theAnuragMishra:
> Also.. locally what seems to have fixed the problem is consuming the AbortSignal

so are you passing it to `reFetchWalletBalance` and cancelling things depending on the signal? 
I can onnly guess why this works...maybe because it cancels the first of the two requests made because of strictmode and that results in no racing state changes and no multiple rerendering. 
But anyways this isn't really a fix to our problem. Usage of abort signal is usually to save resources or prevent outdated changes to affect something.

I agree that we could take this out into a new context but still you would need a ref to tracking previous scanning state, right?  why do you not like it? I think a ref is very suitable here because it is mode to be used to for things like these.

> useRef is a React Hook that lets you reference a value that’s not needed for rendering.

https://react.dev/reference/react/useRef

theAnuragMishra:
@theborakompanioni we don't even need to create another context. just move these effects updating wallet info into the provider itself. clears mess from the router and is a better pattern overall. but i still can't don't think `useRef` is bad for this purpose or there's another way.

tell me if you want to move the reload effects into the provider itself?

theborakompanioni:
> @theborakompanioni we don't even need to create another context. just move these effects updating wallet info into the provider itself. clears mess from the router and is a better pattern overall. but i still can't don't think `useRef` is bad for this purpose or there's another way.

you are probably right. going to review again in the coming days :pray:

theAnuragMishra:
@theborakompanioni so not planning to move the effects into the context itsefl?

theborakompanioni:
> @theborakompanioni so not planning to move the effects into the context itsefl?

Not yet. Focusing on feature parity with legacy version. After that, you can refactor.

PR #1223, opened 2026-04-20
Fix render-phase state update in AccountXpubsDialog

Ashish-Kumar-Dash:
This PR removes a render phase state update in AccountXpubsDialog and keeps timeout handling lifecycle safe.
The change prevents unnecessary re renders and makes expiry behavior more predictable and maintainable.

### DIFF
--- a/src/components/settings/AccountXpubsDialog.tsx
+++ b/src/components/settings/AccountXpubsDialog.tsx
@@ -191,12 +191,8 @@ export const AccountXpubsDialog = ({
   const { t } = useTranslation()
 
   const [passwordVerifiedAt, setPasswordVerifiedAt] = useState<number>()
-  const isPasswordVerified = useMemo(() => passwordVerifiedAt !== undefined, [passwordVerifiedAt])
   const [timeLeft, setTimeLeft] = useState(autoCloseTimeout)
-
-  if (timeLeft <= 0 && passwordVerifiedAt !== undefined) {
-    setPasswordVerifiedAt(undefined)
-  }
+  const isPasswordVerified = passwordVerifiedAt !== undefined && timeLeft > 0
 
   const secondsLeft = Math.max(0, Math.round(timeLeft / 1_000))
 
@@ -206,10 +202,14 @@ export const AccountXpubsDialog = ({
   const queryClient = useQueryClient()
   const { jars } = useJars()
 
-  const seedQueryOptions = getseedOptions({
-    client,
-    path: { walletname: walletFileName },
-  })
+  const seedQueryOptions = useMemo(
+    () =>
+      getseedOptions({
+        client,
+        path: { walletname: walletFileName },
+      }),
+    [client, walletFileName],
+  )
 
   const {
     data: seedQueryData,
@@ -225,7 +225,7 @@ export const AccountXpubsDialog = ({
     select: (data) => data.seedphrase.split(/\s+/) as MnemonicPhrase,
   })
 
-  const accountXpubsQueryKey = [walletFileName, 'xpubs']
+  const accountXpubsQueryKey = useMemo(() => [walletFileName, 'xpubs'], [walletFileName])
 
   const accountXpubs = useQuery({
     queryKey: accountXpubsQueryKey,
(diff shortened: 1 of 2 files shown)


theAnuragMishra (approved):
Looks good to me overall!
Just one suggestion/question, does the `useMemo` really have any use here? Can we not just do `const isPasswordVerified = passwordVerifiedAt !== undefined && timeLeft > 0`?

kishore08-07 (changes requested):
Thanks @Ashish-Kumar-Dash,
Nice cleanup by removing the render‑phase state update and makes the timer lifecycle safer. 

One concern: when the timer runs out, the UI flips back to the password form, but passwordVerifiedAt and cached seed/xpub data are still sitting in memory. Could we add a small timeout cleanup (reset passwordVerifiedAt state + clear queries, or just call handleClose) when timeLeft hits 0? so expiry feels complete.

src/components/settings/AccountXpubsDialog.tsx
```diff
@@ -191,12 +191,11 @@ export const AccountXpubsDialog = ({
   const { t } = useTranslation()
 
   const [passwordVerifiedAt, setPasswordVerifiedAt] = useState<number>()
-  const isPasswordVerified = useMemo(() => passwordVerifiedAt !== undefined, [passwordVerifiedAt])
   const [timeLeft, setTimeLeft] = useState(autoCloseTimeout)
-
-  if (timeLeft <= 0 && passwordVerifiedAt !== undefined) {
-    setPasswordVerifiedAt(undefined)
-  }
+  const isPasswordVerified = useMemo(
```
kishore08-07: Also, useMemo around isPasswordVerified feels a bit heavy for a simple boolean check, a plain const would be efficient.

src/components/settings/AccountXpubsDialog.tsx
```diff
@@ -256,7 +255,11 @@ export const AccountXpubsDialog = ({
 
     const xpubsDisplayedAt = Math.max(accountXpubs.dataUpdatedAt, passwordVerifiedAt)
     const interval = setInterval(() => {
-      setTimeLeft(Math.max(0, xpubsDisplayedAt + autoCloseTimeout - Date.now()))
+      const nextTimeLeft = Math.max(0, xpubsDisplayedAt + autoCloseTimeout - Date.now())
+      setTimeLeft(nextTimeLeft)
+      if (nextTimeLeft <= 0) {
+        clearInterval(interval)
```
theAnuragMishra, line 262: add `setPasswordVerifiedAt(undefined)`.

theAnuragMishra (changes requested):

Ashish-Kumar-Dash:
Thanks guys, made the suggested changes

kishore08-07 (approved):
Thanks @Ashish-Kumar-Dash, LGTM 🤝

src/components/settings/AccountXpubsDialog.tsx
```diff
@@ -225,7 +221,7 @@ export const AccountXpubsDialog = ({
     select: (data) => data.seedphrase.split(/\s+/) as MnemonicPhrase,
   })
 
-  const accountXpubsQueryKey = [walletFileName, 'xpubs']
+  const accountXpubsQueryKey = useMemo(() => [walletFileName, 'xpubs'], [walletFileName])
```
theAnuragMishra, line 228: Do we really need to memoize this?

theAnuragMishra (changes requested):

src/components/settings/AccountXpubsDialog.tsx
Ashish-Kumar-Dash: Yes, in the current implementation it is needed for effect stability.   Without memoization, the array key is recreated every render, so the expiry cleanup effect sees a changed dependency and can rerun unnecessarily.  
If we refactor the effect to depend directly on walletFileName and construct keys inline,  maybe we can remove useMemo, but I'd consider it out of scope for this PR as it is a targeted safety fix, not changing dependency or refactor. If you want, I can do a follow up cleanup PR to remove useMemo

theAnuragMishra (approved):

src/components/settings/AccountXpubsDialog.tsx
theborakompanioni: The same is true for `seedQueryOptions`, no? It must also be memoized for the newly introduced `useEffect` to be stable as it used `seedQueryOptions.queryKey`.

Ashish-Kumar-Dash: You're right! mb, it should also be memoized to keep dependency stable across renders. I'll wrap it in usememo asw. thanks for pointing it out

src/components/settings/AccountXpubsDialog.tsx
```diff
@@ -202,10 +202,13 @@ export const AccountXpubsDialog = ({
...
 
-  const seedQueryOptions = getseedOptions({
+  const seedQueryOptions = useMemo(
+  () => getseedOptions({
     client,
     path: { walletname: walletFileName },
-  })
+  }),
+  [client, walletFileName],
+)
```
theborakompanioni: NIT: format.. run `npm run format`

(We should really do this in pre-commit hooks and validate in GH actions.. 😅 )

Ashish-Kumar-Dash: you're right, also I have squashed commits to a single commit for easy reference

theborakompanioni (approved):

PR #1226, opened 2026-04-22
feat(wallet): add data-journey-state to MainWalletPage with tests

rxiddhi:
## What does this PR do?

This PR adds a derived `data-journey-state` attribute to the main wallet container in `MainWalletPage`.

It also:
- extracts the journey-state derivation into a small pure helper
- derives the state from existing wallet and service signals only
- adds unit tests for the derivation logic
- adds React Testing Library coverage for the rendered attribute
- adds an explicit precedence test to ensure `service-offline` wins over `no-wallet`

This is an additive UI-only change intended to make follow-up state-aware guidance work easier to implement and test. It does not change backend behavior or existing wallet flows.

## Motivation / Context

Jam already has the wallet and service state needed to identify common first-run situations, but that state is not currently surfaced in one predictable place on the main wallet route.

Exposing a deterministic journey-state marker is a small, low-risk step toward better onboarding and contextual wallet guidance.

## Dependencies

- No backend or API changes
- No new dependencies

## Risk

Low. This change only adds a DOM attribute and pure derivation logic, with unit and RTL coverage.

#### Image Demo

The screenshot below shows the main wallet page with Chrome DevTools open and the main wallet container selected in the Elements panel.

The selected node renders:

- `data-journey-state="empty-wallet"`

This demonstrates that the derived journey state is present in the DOM and available for follow-up guidance work, even though this PR does not introduce a visible UI change yet.

image


theborakompanioni:
Closing this: triage.
Reopen anytime if you think this is a mistake.

PR #1231, opened 2026-04-27
chore: reload utxos after blockheight change

theborakompanioni:
After a collaborative transaction, a warning is displayed that some UTXOs need more confirmation to be included in as inputs: "To run the scheduler you need UTXOs with 5 or more confirmations. Fund your wallet and wait for 5 more blocks.".

Before this PR, the message stays until the user reload, even if the UTXOs have the need amount of confirmations.
After this PR, UTXOs are reloaded when blockheight changes and the messages disappears automatically if confirmation preconditions are satifisfied.

Additionally: 
- Change update handling of "initialRowSelection" in `UtxoTable` (e.g. utxos can change anytime, prevent update to selection while user is editing)
- Display "tags" and "confirmations" before "address" in `UtxoTable`

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -48,7 +48,7 @@ import { EarnReportPage } from './components/earn/report/EarnReportPage'
 import { LockWalletConfirmDialog } from './components/ui/jam/LockWalletConfirmDialog'
 import { Spinner } from './components/ui/spinner'
 import { WalletJarsDetailsPage } from './components/wallet/WalletJarsDetailsPage'
-import { useRescanStatus } from './context/JamSessionInfoContext'
+import { useJamSessionInfoContext, useRescanStatus } from './context/JamSessionInfoContext'
 import { JamSessionInfoContextProvider } from './context/JamSessionInfoContextProvider'
 import { useJamWalletInfoContext } from './context/JamWalletInfoContext'
 import { useJmWebsocket } from './hooks/useJmWebsocket'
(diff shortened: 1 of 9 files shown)


src/App.tsx
```diff
@@ -472,13 +478,38 @@ const WalletInfoAutoReload = () => {
...
       return () => abortCtrl.abort('useEffect(AFTER_RESCAN) ended')
     },
     [refetchWalletBalance, currentRescanInfo.rescanning],
   )
 
+  useEffect(
+    function refetchWalletInfoAfterBlockHeightIncrease() {
+      if (currentBlockHeight === undefined) return
+
+      const blockHeightIncreased =
```
parrth20: Small nit
this is named `blockHeightIncreased`, but the check is !==, so it also runs if the height ever goes backwards/stale. If that’s intentional, maybe rename it to `blockHeightChanged` for better understanding, otherwise we could compare `currentBlockHeight` > `previousBlockHeightRef.current`.

src/hooks/useUtxoSelectionDialog.ts
```diff
@@ -175,27 +168,25 @@ export const useUtxoSelectionDialog = ({ walletFileName, jars, addressSummary }:
...
   const dialogProps: UtxoSelectionDialogProps = {
     open,
-    isApplying,
+    isSubmitting,
     selectedCount: selectedUtxos.length,
     filter,
     tableEntries,
     initialRowSelection,
-    enableRowSelection: (row) => !fb.utxo.isFidelityBond(row.original.utxo),
+    enableRowSelection: isSubmitting ? false : undefined,
```
parrth20, line 176: tiny UX thought because this now disables row selection during submit: should we also guard the select-all checkbox state in the table header, or is it fine since row-level selection is already blocked?

parrth20 (commented):
LGTM !!

src/App.tsx
theborakompanioni: Good catch. `blockHeightChanged` is better.

src/hooks/useUtxoSelectionDialog.ts
theborakompanioni: Totally! Good catch. Thank you :pray:

PR #1232, opened 2026-04-28
V2 JoinMarket NG tumbler

m0wer:
This is a continuation of #1166 that adds breaking changes to the jmwalletd API.

The tumbler redesign shares and endpoint for the schedule.

By default the tumbler will fail since the default setting is to not accepted bondless makers with more than 0 fees. Need to find a way to switch that off for testing.


theborakompanioni:
Closing: Done with https://github.com/joinmarket-webui/jam/pull/1339 and https://github.com/joinmarket-webui/jam/pull/1305.

Please reopen if you think this is a mistake.

PR #1234, opened 2026-04-29
chore: valid number must be finite

theborakompanioni:
Change semantic of `isValidNumber` and introcude new utility function `isValidInteger`.

Before, `+Infinity` and `-Infinity` were treated as valid numbers, did not make any sense.
Using `isValidInteger` ensures numbers dont grow too large. However, validation of user inputs should still set min/max boundaries (e.g. checking for a **positive** integer, etc.)

### DIFF
--- a/src/components/settings/fees/CollaboratorFeesForm.schema.ts
+++ b/src/components/settings/fees/CollaboratorFeesForm.schema.ts
@@ -1,7 +1,7 @@
 import type { TFunction } from 'i18next'
 import * as yup from 'yup'
 import { CJ_FEE_ABS_MAX, CJ_FEE_ABS_MIN, CJ_FEE_REL_MAX, CJ_FEE_REL_MIN } from '@/constants/jam'
-import { factorToPercentage, formatSats } from '@/lib/utils'
+import { factorToPercentage, formatSats, isValidInteger } from '@/lib/utils'
 import type { AmountSats } from '@/types/global'
 
 export type CollaboratorFeesFormValues = {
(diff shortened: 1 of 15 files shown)


src/components/settings/fees/CollaboratorFeesForm.schema.ts
```diff
@@ -28,13 +28,13 @@ export function createCollaboratorFeesFormSchema({
...
         .integer(maxCjFeeAbsoluteMessage)
-        .transform((value) => (Number.isSafeInteger(value) ? Number(value) : null))
+        .transform((value) => (isValidInteger(value) ? value : null))
         .min(CJ_FEE_ABS_MIN, maxCjFeeAbsoluteMessage)
         .max(CJ_FEE_ABS_MAX, maxCjFeeAbsoluteMessage)
         .required(maxCjFeeAbsoluteMessage),
       maxCjFeeRelInPercent: yup
         .number()
-        .transform((value) => (Number.isFinite(value) ? Number(value) : null))
+        .transform((value) => (isValidInteger(value) ? value : null))
```
parrth20, line 37: i think this one should stay `isValidNumber`. `maxCjFeeRelInPercent` is a percentage field and the input allows decimals (`step="0.0001"`), so `isValidInteger` would reject valid values like `0.002%` / `0.0001%`... the absolute fee field above makes sense as integer, but this relative one should allow finite decimals.

parrth20 (commented):
Overall LGTM !!

src/components/settings/fees/CollaboratorFeesForm.schema.ts
theborakompanioni: Damn.. I am such a noob. Good catch.

parrth20 (approved):

PR #1224, opened 2026-04-21
refactor(forms): use react-hook-form for sweep and fidelity bond flows

parrth20:
summary
migrate the sweep destination form to react-hook-form and yup
migrate the fidelity bond creation flow to react-hook-form and yup
kept the existing validation behaviour while removing manual form state

ping @theborakompanioni

PR #1235, opened 2026-04-30
refactor(forms):replace yup path parsing with array-level validation and optimize cross-field checks

kishore08-07:
## What does this PR do?

- This is a follow-up PR of https://github.com/joinmarket-webui/jam/pull/1224 to improve and enhance validation logic.

### Changes
- Removed  yup path parsing; switched to array-level validation with explicit index-based errors
- Optimized cross-field validation to avoid redundant checks
- Separated field-level (format) and cross-field (duplicate/reuse) validation
- Reduced unnecessary re-validation using resolver context
- Improved type safety by removing unsafe casts and using typed schemas
- Updated tests to reflect the new validation behavior


src/components/sweep/SweepPage.tsx
```diff
@@ -97,14 +97,14 @@ export const SweepPage = ({ walletFileName }: SweepPageProps) => {
...
+  const schema = useMemo(() => sweepFormSchema(t), [t])
   const initialDestinations = useMemo(() => buildSweepDestinationValues(DESTINATION_ADDRESS_COUNT_PROD), [])
-  const form = useForm<SweepFormValues, unknown, SweepFormValues>({
+  const form = useForm<SweepFormValues, SweepResolverContext, SweepFormValues>({
     mode: 'onChange',
     defaultValues: {
       destinations: initialDestinations,
     },
-    resolver: yupResolver(schema as yup.AnyObjectSchema) as Resolver<SweepFormValues, unknown, SweepFormValues>,
+    resolver: yupResolver(schema, { context: { addressSummary: walletInfo.addressSummary } }),
```
parrth20: I think this context is being passed in the wrong place. `yupResolver` receives RHF's form context as the second argument of the resolver function, and that ends up overriding `schemaOptions.context`, so `this.options.context` can become `undefined` here. That would make the reused-address check always fall back to `{}` in the app, even though the schema unit test passes because it calls `schema.validate(..., { context })` directly.

Maybe we should pass it through `useForm` instead:

```ts
const form = useForm<SweepFormValues, SweepResolverContext, SweepFormValues>({
  mode: 'onChange',
  defaultValues: { destinations: initialDestinations },
  context: { addressSummary: walletInfo.addressSummary },
  resolver: yupResolver(schema),
})

parrth20 (changes requested):
@kishore08-07 Nice direction overall, the array-level validation is cleaner than parsing Yup paths. I've left one comment though because the resolver context wiring looks wrong and can break reused-address validation in the actual form.

src/components/sweep/SweepPage.tsx
kishore08-07: Thanks @parrth20 for pointing that out! I’ve addressed it by passing addressSummary through useForm’s context , so it flows correctly into the resolver.

parrth20 (approved):
@kishore08-07 LGTM now, thanks for updating this.

PR #1239, opened 2026-05-03
fix(earn): try prevent "no confirmed coins" on earn page

parrth20:
Uses confirmed available wallet balance for Earn minsize validation to avoid starting maker with only unconfirmed coins.


theborakompanioni (dismissed):
LGTM.

While it prevents to show the user the "unconfirmed coins" error, it does give them reason and that might be confusing.. what do you think?

Image

Can you think of a better UX?
e.g. 
- "warning alert" with "Your balance is not yet confirmed. You need to wait 1 more block" (that auto-disappears since we are reloading utxos when blockheight changes)

Just an example there are definitely better solutions.
(Same is also for "no balance" -> Should there be a warning with reason and link with "fund wallet" that navigates to receive page?)

parrth20:
> LGTM.
> 
> While it prevents to show the user the "unconfirmed coins" error, it does give them reason and that might be confusing.. what do you think?
> 
> Image
> Can you think of a better UX? e.g.
> 
> * "warning alert" with "Your balance is not yet confirmed. You need to wait 1 more block" (that auto-disappears since we are reloading utxos when blockheight changes)
> 
> Just an example there are definitely better solutions. (Same is also for "no balance" -> Should there be a warning with reason and link with "fund wallet" that navigates to receive page?)

I've added an explicit warning for unconfirmed/no spendable balance and disabled only the Start action, so users know why Earn is unavailable without hitting the backend error first.

parrth20:
ping @theborakompanioni

src/components/earn/EarnForm.tsx
```diff
@@ -126,6 +126,7 @@ interface EarnFormProps {
   isWaitingMakerStart: boolean
   onSubmit: SubmitHandler<EarnFormValues>
   offerMinsizeMax: AmountSats
+  submitDisabled?: boolean
```
theborakompanioni: NIT: just use disabled - much more simple.

src/components/earn/EarnPage.tsx
```diff
@@ -246,10 +255,35 @@ export const EarnPage = ({ walletFileName }: EarnPageProps) => {
         <CardContent>
           <p className="text-muted-foreground mb-4 text-sm">{t('earn.market_explainer')}</p>
 
+          {earnBalanceWarning && (
```
theborakompanioni: NIT: Can this be split in two components?
One alert is fine, if n > 1 it is easier to read if they are separated.

src/i18n/locales/en/translation.json
```diff
@@ -500,6 +500,10 @@
     "alert_loading_offer": "Loading offer...",
     "alert_running": "Offering sats to earn more sats...",
     "alert_coinjoin_in_progress": "A collaborative transaction is currently in progress.",
+    "alert_unconfirmed_balance_title": "Balance is not confirmed yet",
+    "alert_unconfirmed_balance_description": "Your wallet has available funds, but they need at least one confirmation before the earn service can start. This should update automatically after the next block.",
+    "alert_no_spendable_balance_title": "No spendable balance available",
+    "alert_no_spendable_balance_description": "Fund your wallet on the <1>Receive page</1> to start earning.",
```
theborakompanioni, line 506: NIT: Wording can be improved, but this can be done later once we have functional parity with legacy 🙏

theborakompanioni (approved):
LGTM!
Test both warnings + auto hide after block height increase 💪

PR #1240, opened 2026-05-03
chore(regtest): initial joinmarket-ng backend for regtest environment

theborakompanioni:
Adds jmwalletd, directory servers, makers, and bitcoind to the regtest docker-compose environment so jam can be developed against the jm-ng backend end-to-end.

- chore(regtest): wire docker compose env and vite dev proxy for jm-ng

Connects the local regtest containers into a single compose network, adds tor and fund-wallet helpers, and points Vite's dev proxy at the local jmwalletd instance.

- fix(tags): recognize cj-change, used-empty, and flagged UTXO statuses

joinmarket-ng's WalletService emits four statuses jam's JmPlainTagValue didn't cover: 'cj-change' (deanonymising change from our CJ), 'used-empty' (previously-used address with zero balance), and 'flagged' (address shared in a CJ that later failed). They fell through to the generic 'default' badge with no distinguishing styling.

- docs(dev): explain jm-ng backend workflow

- fix(auth): only clear auth on invalid_token 401s

Previously every 401 response cleared the session and logged the user out. jm-ng returns 401 for service-state errors (e.g. POST /tumbler/stop when nothing is running), which should never drop the session. Inspect the WWW-Authenticate header and only clear auth when the server signals ``error="invalid_token"``.
jm-ng `jmwalletd` exposes the HTTPS API and WebSocket on separate TCP ports (28183 and 28283 by default), not the same port as previously assumed. Restore:

- chore(dev): distinct server configs for various backends

Add `joinmarket-clientserver`, `joinmarket-ng` and `jam-standalone` as backend options.

### DIFF
--- a/docker/regtest/docker-compose-common.yml
+++ b/docker/regtest/docker-compose-common.yml
@@ -1,28 +1,31 @@
-
 services:
-
   joinmarket_native:
     build:
       context: ./dockerfile-deps/joinmarket/latest
       dockerfile: Dockerfile
     restart: unless-stopped
     environment:
-      ENSURE_WALLET: "true"
-      REMOVE_LOCK_FILES: "true"
+      ENSURE_WALLET: 'true'
+      REMOVE_LOCK_FILES: 'true'
       jm_blockchain_source: regtest
       jm_network: testnet
       jm_rpc_host: bitcoind
       jm_rpc_port: 43782
-      jm_directory_nodes: ${JM_DIRECTORY_NODES:?You must set the onion address generated in prepare step to your .env file}
+      jm_directory_nodes: ${JM_ALL_DIRECTORY_NODES:?You must set the directory node addresses in generated env file}
+      jm_socks5_host: tor
+      jm_socks5_port: 9050
+      jm_tor_control_host: tor
+      jm_tor_control_port: 9051
       jm_minimum_makers: 1 # necessary to do coinjoins with this regtest setup; default is 4
       jm_taker_utxo_age: 1 # faster testing of scheduler runs; default is 5
       jm_maker_timeout_sec: 30 # easier testing of maker timeouts on regtest (and "Stall Monitor" retries); default is 60
     expose:
       - 62601 # obwatch
+      - 27183 # joinmarketd daemon
       - 28183 # jmwalletd api
       - 28283 # jmwalletd websocket
     healthcheck:
-      test: ["CMD", "supervisorctl", "status"]
+      test: ['CMD', 'supervisorctl', 'status']
       interval: 10s
       timeout: 10s
       retries: 20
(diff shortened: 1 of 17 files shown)


docker/regtest/docker-compose-common.yml
```diff
@@ -1,28 +1,31 @@
...
       jm_blockchain_source: regtest
       jm_network: testnet
       jm_rpc_host: bitcoind
       jm_rpc_port: 43782
-      jm_directory_nodes: ${JM_DIRECTORY_NODES:?You must set the onion address generated in prepare step to your .env file}
+      jm_directory_nodes: ${JM_ALL_DIRECTORY_NODES:?You must set the directory node addresses in generated env file}
+      jm_socks5_host: tor
+      jm_socks5_port: 9050
+      jm_tor_control_host: tor
+      jm_tor_control_port: 9051
```
m0wer, line 18: Did not work without but didn't look into why.

m0wer (dismissed):
ACK!

docker/regtest/docker-compose-common.yml
theborakompanioni: Gonna try it and report back.

kishore08-07 (dismissed):
@m0wer , @theborakompanioni LGTM 🚀

docker/regtest/tor/conf/torrc
```diff
@@ -0,0 +1,15 @@
...
+SocksPort 0.0.0.0:9050
+
+# Control port for maker onion service handling
+ControlPort 0.0.0.0:9051
+CookieAuthentication 1
+CookieAuthFile /var/lib/tor/control_auth_cookie
+
+# Hidden service for JoinMarket NG directory node
+HiddenServiceDir /var/lib/tor/ng-directory
+HiddenServicePort 5222 172.30.0.21:5223
```
parrth20, line 19: Small thing I might be wrong or missing but...`prepare-setup.sh` now generates both ref + jm-ng onion addresses and `JM_ALL_DIRECTORY_NODES` points clients to both, but this torrc only exposes the jm-ng directory hidden service. Since `tor_init` also copies `/var/lib/tor/ref-directory`, should we add a second `HiddenServiceDir` / `HiddenServicePort` for the reference directory node too? Otherwise the ref onion in `JM_ALL_DIRECTORY_NODES` looks dead.

docs/developing.md
```diff
@@ -8,6 +8,28 @@ A place to collect useful information for developers that doesn't really fit els
...
+If you are using Jam's own regtest environment, the initialized jm-ng services are exposed on specific host ports. In that case run:
+
+```bash
+npm run jm-ng:dev
+```
+
+If your separately running jm-ng services use different ports, you can override them directly:
+
+```bash
+JAM_BACKEND=joinmarket-ng
```
parrth20: Tiny doc fix: this first env assignment is on its own line, so `JAM_BACKEND` will not be passed into `npm run dev`. Could we add a trailing `\` here or put the command on one line? Otherwise users following this example still hit the default backend.

vite.config.ts
```diff
@@ -3,31 +3,66 @@ import react from '@vitejs/plugin-react'
...
 
+const BACKEND_ENV_JOINMARKET_CLIENTSERVER_DEFAULT: BackendConfigEnvironmentVariables = {
+  JMWALLETD_API_PORT: 28183,
+  JMWALLETD_WEBSOCKET_PORT: 28283,
+  JMOBWATCH_PORT: 62601,
+}
+
+const BACKEND_ENV_JOINMARKET_NG_DEFAULT: BackendConfigEnvironmentVariables = {
+  JMWALLETD_API_PORT: 28183,
+  JMWALLETD_WEBSOCKET_PORT: 28183,
```
parrth20, line 46: Just curious: is `JMWALLETD_WEBSOCKET_PORT: 28183` intentional for the jm-ng default? The PR text/docs mention separate API/websocket ports in some places, but the default points both to 28183. If jm-ng’s default websocket is still 28283 outside the regtest host mapping, this may make `JAM_BACKEND=joinmarket-ng npm run dev` fail unless users override it.

docker/regtest/tor/conf/torrc
m0wer: Good point. Unless ref creates its own onion through Tor's control port, which I don't think is the case.

docs/developing.md
m0wer: Yes please!

Or `export ...`

vite.config.ts
m0wer: The default port is 28183 and accepts websockets connections there:
```
# Mounted at /ws, /jmws, and /api/v1/ws in app.py.
```

I changed this during development so could use some cleanup.

parrth20:
@m0wer pushed a small commit for both review points...added the ref directory hidden service and fixed the docs env example. Left the jm-ng websocket default as-is since 28183 is intentional

m0wer (approved):
LGTM

parrth20 (approved):
LGTM 🚀🚀

kishore08-07 (approved):

theborakompanioni:
Yeah, I think the tor config still does not work as expected (`jm_regtest_joinmarket_directory_node` not reachable?), but that can be done in follow-ups.

Thanks everyone :pray:

PR #1243, opened 2026-05-09
chore(ui): improve footer on small screens

theborakompanioni:
image

## How to test
`run storybook:up`
http://localhost:6006/?path=/story/layout-appfooter--default&globals=backgrounds.grid:!false;viewport.value:525-568

### DIFF
--- a/src/components/layout/AppFooter.tsx
+++ b/src/components/layout/AppFooter.tsx
@@ -84,7 +84,7 @@ export function AppFooter({
               </a>
             </div>
           </div>
-          <div className="order-last flex inline-block flex-1 text-xs sm:order-first">
+          <div className="order-last flex-1 text-xs sm:order-first">
             <Trans i18nKey="footer.warning">
               This is pre-alpha software.
               <Button
--- a/src/components/layout/AppFooter.tsx
+++ b/src/components/layout/AppFooter.tsx
@@ -1,16 +1,15 @@
 import { useState, type ComponentProps } from 'react'
 import { BlocksIcon, BookOpenIcon, FileQuestionMarkIcon, ScrollTextIcon } from 'lucide-react'
 import { useTranslation, Trans } from 'react-i18next'
-import { useStore } from 'zustand'
 import { Button } from '@/components/ui/button'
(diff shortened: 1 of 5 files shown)


src/components/layout/AppFooter.tsx
```diff
@@ -38,57 +37,65 @@ export function AppFooter({
...
+                target="_blank"
+                rel="noopener noreferrer"
+                className="text-muted-foreground underline"
+              >
+                v{jamVersion?.raw}
+              </a>
+            </div>
+          </div>
+          <div className="order-last flex inline-block flex-1 text-xs sm:order-first">
+            <Trans i18nKey="footer.warning">
```
parrth20: Small UX nit: `footer.warning` includes a `<br/>` in the translation string, and this gets rendered through this `<Trans>`. Since the wrapper is now `flex`, that line break may not behave like normal text flow. Could we make this wrapper a normal block/text container instead of `flex inline-block`?

parrth20 (commented):
@theborakompanioni LGTM !!

parrth20 (changes requested):

src/components/layout/AppFooter.tsx
```diff
@@ -38,57 +37,65 @@ export function AppFooter({
...
+                target="_blank"
+                rel="noopener noreferrer"
+                className="text-muted-foreground underline"
+              >
+                v{jamVersion?.raw}
+              </a>
+            </div>
+          </div>
+          <div className="order-last flex-1 text-xs sm:order-first">
+            <Trans i18nKey="footer.warning">
```
parrth20, line 88: Small UX nit: `footer.warning` includes a <br/> in the translation string, and this gets rendered through this <Trans>. Since the wrapper is now flex, that line break may not behave like normal text flow. Could we make this wrapper a normal block/text container instead of flex inline-block?

parrth20 (approved):
Rechecked this end-to-end. the latest wrapper change here fixes my earlier `footer.warning` concern, and the small-screen footer looks good now. LGTM from my side.

PR #1244, opened 2026-05-09
chore(websocket): add JmWebsocketContext

theborakompanioni:
Add a context for websocket stuff.
Websocket connection has been shared, but we really only need one connection (yet).
No custom message handling by children, since we also do not need that at the moment.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -51,10 +51,9 @@ import { WalletJarsDetailsPage } from './components/wallet/WalletJarsDetailsPage
 import { useJamSessionInfoContext, useRescanStatus } from './context/JamSessionInfoContext'
 import { JamSessionInfoContextProvider } from './context/JamSessionInfoContextProvider'
 import { useJamWalletInfoContext } from './context/JamWalletInfoContext'
-import { useJmWebsocket } from './hooks/useJmWebsocket'
+import { JmWebsocketContextProvider } from './context/JmWebsocketContextProvider'
 import { getErrorReason } from './lib/errorReason'
 import { jmSessionStore } from './store/jmSessionStore'
-import { jmTxStore, type JmTxInfo } from './store/jmTxStore'
 import type { Milliseconds } from './types/global'
 
 const DevSetupPage = lazy(() => import('@/components/dev/DevSetupPage'))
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -248,27 +247,24 @@ function App() {
...
-              onOpenChange={() => setLockWalletDialogContext(undefined)}
-              onConfirm={() => doOnLockWalletConfirm(lockWalletDialogContext?.navigate, lockWalletDialogContext?.t)}
-              makerRunning={makerRunning}
-              coinjoinInProgress={coinjoinInProgress}
-            />
-          )}
-          <RouterProvider router={router} />
-          <Toaster closeButton />
-        </QueryClientProvider>
+        <JmWebsocketContextProvider>
```
parrth20, line 250: only one small concern this provider now wraps the whole app, including login/create/import routes. Since `useJmWebsocket` connects immediately and keeps reconnecting, we may now open unauthenticated websocket connections before a wallet is loaded, and after logout the socket can also stay around until the backend closes it. Could we either move this provider inside the protected route, or add an `enabled/authenticated` guard to the websocket hook?

parrth20 (changes requested):

src/App.tsx
theborakompanioni: Good observation. I guess this is the expected behaviour.
Connection can be established without loaded wallet and sensitive info is only sent once the connection is authenticated. I'd keep this approach if it is okay for you.

kishore08-07 (commented):
nit: `enableAuthentiction` is misspelled; could you rename it to `enableAuthentication`?

Everything else looks fine to me.

src/App.tsx
parrth20: I get what you are saying as...that no sensitive data is sent before authentication, so security-wise this is fine. my concern is more about lifecycle/noise: the app will still open and retry the websocket on unauthenticated routes....If we’re okay with that behaviuor for now, I’m fine with keeping it, but long-term an `enabled` guard might keep the connection lifecycle cleaner.

parrth20 (approved):
Otherwise LGTM we are GTG !!!

kishore08-07 (approved):

PR #1245, opened 2026-05-09
chore(send): collaborative send

theborakompanioni:
Enhance the collaborative send flow to improve reliability and UX by saving the last payment attempt, better error reporting, and minor refactors.
Tries to mitigate missing infos on failed attempts: `joinmarket-clientserver` does not provide good info about errors or other failures.
Flow:
- store payment attempt + utxos hash in session storage
- wait till taker operation ends
- if utxos hash changed, show a success message
- if utxos has did not change, show an error message with payment attempt details

Note:
- if a user did not start the taker operation in this session, no info can be shown and collaborative send flow just ends without any information at all - this is not good UX, but we cannot really do anything about it.


src/context/JamSessionInfoContext.ts
```diff
@@ -1,15 +1,30 @@
 import { createContext, useContext, type Dispatch, type SetStateAction } from 'react'
+import type { SendFormValues } from '@/components/send/types'
 
 export interface RescanInfo {
   updatedAt: number
   rescanning: boolean
   progress?: number
 }
 
+export interface PaymentAttempt {
```
parrth20, line 11: since this attempt is persisted in session storage, can we include `walletFileName` here and only expose the attempt when it matches the current wallet?

Otherwise an attempt from wallet A could survive logout/import and show up on wallet B, which may show the wrong success/failure state or keep the send form disabled.

src/components/send/SendPage.tsx
```diff
@@ -189,128 +187,54 @@ export const SendPage = ({ walletFileName }: SendPageProps) => {
     },
   })
 
-  const coinjoinRunning = jmSession?.coinjoin_in_process === true
-  const isWaitingCoinjoinStart = startCoinjoinMutationIsPending || (startCoinjoinMutationIsSuccess && !coinjoinRunning)
-  const isWaitingCoinjoinStop = stopCoinjoinMutationIsPending || (stopCoinjoinMutationIsSuccess && coinjoinRunning)
-  const collaborativeFlowActive = coinjoinRunning || isWaitingCoinjoinStart || isWaitingCoinjoinStop
+  const isWaitingCoinjoinStart = startCoinjoinMutationIsPending || (startCoinjoinMutationIsSuccess && !takerRunning)
```
parrth20, line 192: edge case: if `/docoinjoin` succeeds but `coinjoin_in_process` never flips true, or flips too quickly before the session poll sees it, this can stay in “Starting collaborative transaction...” forever.

Maybe we should add a timeout/reset fallback here, or allow the stored attempt to move into the completion/ended state after some time.

src/context/JamSessionInfoContext.ts
theborakompanioni: Will do!

src/components/send/SendPage.tsx
theborakompanioni: Good observation. While that is certainly true, it complicates the code and might create even more problems. I will revisit and see what can be done. Maybe address in a follow-up.

src/context/JamSessionInfoContext.ts
theborakompanioni: Done.

src/components/send/SendPage.tsx
```diff
@@ -500,34 +408,122 @@ export const SendPage = ({ walletFileName }: SendPageProps) => {
...
+                              }}
+                            >
+                              {t('global.done')}
+                            </Button>
+                          </div>
+                        </AlertDescription>
+                      </Alert>
+                    ) : (
+                      <Alert variant="success">
+                        <AlertTriangleIcon />
```
parrth20: smoll UI nit: this is the success branch, but it still uses `AlertTriangleIcon`. Maybe `CheckCircle2Icon` would make the completed state feel less like a warning.

theborakompanioni: Done!

parrth20 (approved):
I’d skip the small typo/JSON-alert/style comments for now.... as they’re not worth blocking this PR, and you already said the flow is heuristic-based because the API does not give exact taker success details !

PR #1249, opened 2026-05-14
chore(ui): improve activity indicators

theborakompanioni:
Improve activity indicators on navbar items and login wallet select.
Use `animate-ping` with a slowed down animation duration.

## Test
Run `npm run storybook:up` and see http://localhost:6006/?path=/docs/jam-activityindicator--docs

### DIFF
--- a/src/components/ui/jam/ActivityIndicator.tsx
+++ b/src/components/ui/jam/ActivityIndicator.tsx
@@ -1,0 +1,31 @@
+import type { PropsWithChildren } from 'react'
+import { cn } from '@/lib/utils'
+
+type ActivityIndicatorProps = {
+  active?: boolean
+  animateEnter?: boolean
+  className?: string
+}
+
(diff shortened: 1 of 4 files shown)


src/components/ui/jam/ActivityIndicator.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+export const ActivityIndicator = ({ active = true, animateEnter = false, className }: ActivityIndicatorProps) => {
+  return (
+    <span
+      className={cn('relative flex size-2', className, {
+        hidden: !active,
+        'animate-in fade-in-0 zoom-in-0 duration-2000': animateEnter,
+      })}
+    >
+      <span className="light:bg-green-500 absolute inline-flex h-full w-full animate-ping rounded-full bg-green-200 opacity-75 [animation-duration:_2.1s]" />
```
kishore08-07, line 18: We can consider using motion-safe:animate-ping, but if always animating is the intent here, this can be skipped.

PR #1247, opened 2026-05-13
chore(storybook): add core component stories

parrth20:
Adds Storybook coverage for reusable Jam and core UI components so they can be inspected in isolation across common states and layouts.

This PR keeps the scope focused on component stories only:
- Adds stories for reusable Jam UI pieces like page title, loading/error alerts, copy/masked text, jar helpers, cheatsheet, sorting, and table pagination.
- Adds stories for core primitives like checkbox, radio group, tabs, textarea, dialog, field, input group, dropdown menu, popover, sheet, table, and pagination.
- Keeps stories small and useful for reviewing component states, variants, and responsive behavior.
- Avoids full page-flow stories for now, since those are broader and can be handled later if needed.
- Leaves GitHub Pages deployment as a separate follow-up task/PR.

This follows the direction we discussed :
- Do not split this too fine-grained per component; one Storybook coverage pass is enough.
- Think from a UI/design review perspective and cover components that are useful to inspect.
- Use Storybook as a base for future mobile/small-screen UI checks and iterative improvements.

fixes a part of #1246

PR #1252, opened 2026-05-19
chore(storybook): deploy Storybook preview to GitHub Pages

parrth20:
summary
add a Storybook static build script
add a GitHub Pages workflow that builds Storybook on PRs and deploys latest devel
document the local Storybook build command and hosted preview URL

note
It's a deployment-only follow-up to #1247 !!

### DIFF
--- a/.github/workflows/storybook-pages.yml
+++ b/.github/workflows/storybook-pages.yml
@@ -1,0 +1,77 @@
+name: Storybook Pages
+
+on:
+  push:
+    branches: [devel]
+    paths:
+      - '.github/workflows/storybook-pages.yml'
+      - '.storybook/**'
+      - 'package.json'
+      - 'package-lock.json'
+      - 'src/**'
+  pull_request:
+    branches: [devel]
+    paths:
+      - '.github/workflows/storybook-pages.yml'
+      - '.storybook/**'
+      - 'package.json'
+      - 'package-lock.json'
+      - 'src/**'
+
+permissions:
+  contents: read
+  pages: write
+  id-token: write
+
+concurrency:
+  group: storybook-pages
+  cancel-in-progress: true
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-version: ['v24.11.1']
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
+
+      - name: Setup (Node.js ${{ matrix.node-version }})
+        uses: actions/setup-node@v6
+        with:
+          node-version: ${{ matrix.node-version }}
+
+      - name: Install
+        run: npm clean-install --no-audit --no-fund
+
+      - name: Build Storybook
+        run: npm run storybook:build:pages
+
+      - name: Prepare Pages artifact
+        if: github.event_name == 'push' && github.ref == 'refs/heads/devel'
+        run: touch storybook-static/.nojekyll
+
+      - name: Configure Pages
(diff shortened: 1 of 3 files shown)


parrth20:
ping @theborakompanioni

.github/workflows/storybook-pages.yml
```diff
@@ -0,0 +1,77 @@
...
+      - 'package-lock.json'
+      - 'src/**'
+  pull_request:
+    branches: [devel]
+    paths:
+      - '.github/workflows/storybook-pages.yml'
+      - '.storybook/**'
+      - 'package.json'
+      - 'package-lock.json'
+      - 'src/**'
```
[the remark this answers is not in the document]
parrth20: @theborakompanioni I don’t think missing these would be a big practical issue often, but it is technically valid because config changes can affect the build. Should we include them for safety, or keep the path filter focused only on Storybook-related files !! ???

.github/workflows/storybook-pages.yml
```diff
@@ -0,0 +1,77 @@
...
+on:
+  push:
+    branches: [devel]
+    paths:
+      - '.github/workflows/storybook-pages.yml'
+      - '.storybook/**'
+      - 'package.json'
+      - 'package-lock.json'
+      - 'src/**'
+  pull_request:
```
theborakompanioni: If the intention is that storybook is built, but not deployed on Pull Requests, what do you think about moving this to a different action workflow?

Then the conditions `if: github.event_name == 'push' && github.ref == 'refs/heads/devel'` can be dropped and the flows become much simpler.

What do you think?

.github/workflows/storybook-pages.yml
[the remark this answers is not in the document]
theborakompanioni: Can we remove the `paths` section alltogether? Just redeploy on every push to `devel` is fine. What do you think?

.github/workflows/storybook-pages.yml
```diff
@@ -0,0 +1,77 @@
...
+        run: npm clean-install --no-audit --no-fund
+
+      - name: Build Storybook
+        run: npm run storybook:build:pages
+
+      - name: Prepare Pages artifact
+        if: github.event_name == 'push' && github.ref == 'refs/heads/devel'
+        run: touch storybook-static/.nojekyll
+
+      - name: Configure Pages
```
theborakompanioni, line 40: Is this step really needed?
The [reference flow](https://storybook.js.org/docs/sharing/publish-storybook#github-pages) does not have this step. Maybe not necessary?

package.json
```diff
@@ -24,6 +24,7 @@
     "preview": "vite preview",
     "storybook:up": "storybook dev -p 6006",
     "storybook:build": "storybook build",
+    "storybook:build:pages": "storybook build --output-dir storybook-static",
```
theborakompanioni: `storybook:build` already has `storybook-static` as output dir by default does it?
Either remove the custom command or add it to `storybook:build` to be explicit. What do you think?

theborakompanioni (dismissed):
Small nits.
Also, I thought we settled on deploying to `jam-storybook` (https://joinmarket-webui.github.io/jam-storybook)? But this is also fine with me - we can change it anytime.

ACK.

parrth20:
ping @theborakompanioni I simplified it now: removed the path filters, dropped the PR-only workflow branching, removed the extra pages build script and .nojekyll step, and reused the normal storybook:build. This should keep the Pages workflow closer to the reference flow and easier to maintain.

theborakompanioni (approved):

PR #1254, opened 2026-05-21
chore(storybook): add mobile/tablet viewports

theborakompanioni:
- update storybook-pages actions
- disable isolation for storybook tests (2x speed-up and less "Cannot connect to the iframe" errors)
- add mobile/tablet viewports

image

### DIFF
--- a/.storybook/preview.tsx
+++ b/.storybook/preview.tsx
@@ -5,6 +5,7 @@ import { ThemeProvider } from 'next-themes'
 import { I18nextProvider } from 'react-i18next'
 import { MemoryRouter } from 'react-router-dom'
 import { CoreTypes, GlobalTypes } from 'storybook/internal/csf'
+import { DEFAULT_VIEWPORT, MINIMAL_VIEWPORTS, INITIAL_VIEWPORTS } from 'storybook/viewport'
 import { JamDisplayContextProvider } from '../src/context/JamDisplayContextProvider'
 import i18n from '../src/i18n/config'
 import '../src/index.css'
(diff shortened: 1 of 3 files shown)


.storybook/preview.tsx
```diff
@@ -5,6 +5,7 @@ import { ThemeProvider } from 'next-themes'
 import { I18nextProvider } from 'react-i18next'
 import { MemoryRouter } from 'react-router-dom'
 import { CoreTypes, GlobalTypes } from 'storybook/internal/csf'
+import { DEFAULT_VIEWPORT, MINIMAL_VIEWPORTS, INITIAL_VIEWPORTS } from 'storybook/viewport'
```
parrth20, line 8: This new `storybook/viewport` import seems to make the Storybook Vitest run unstable on a fresh cache. When I applied this PR locally, the first `npm run test:storybook` run failed because Vite had to newly optimize `storybook/viewport` and reloaded mid-run...the second run passed after the dependency was cached.

Could we add `storybook/viewport` to `optimizeDeps.include` for the Storybook/Vitest config, or otherwise ensure it is pre-bundled before tests start???

theborakompanioni: Could not reproduce. Okay to proceed without?

parrth20: Yes, okay to proceed without it.

parrth20: Looks like it was probably a local cache issue on my side. Since CI passes and you could not reproduce it, I’m fine leaving it as-is.

parrth20 (approved):

PR #1255, opened 2026-05-21
refactor: move functions and cleanup

theborakompanioni:
Adds function `clamp` and move `median` to `lib/utils.ts`.


parrth20 (approved):
LGTM !!

PR #1290, opened 2026-06-19
fix: handle locale-specific decimal separators in BTC balance

### LINKED ISSUE #1289: Balance crashes when rendering BTC amounts in decimal-comma locales
## Description

Rendering a BTC-denominated balance can crash Jam when the browser locale uses a decimal comma, such as `it-IT`.

The production build reports:

```text
TypeError: u is not iterable
```

The error occurs because the formatted BTC value is locale-aware, while `BitcoinBalance` assumes that the decimal separator is always `.`
. For an Italian locale, a value such as `0.26120925` is formatted as `0,26120925`; splitting on `.` leaves the fractional part undefined
, and spreading it throws the exception.

## Steps to reproduce

1. Use a browser configured with the Italian locale (`it-IT`).
2. Open a wallet in Jam.
3. Switch the display currency so that a BTC balance is rendered.
4. Observe the error page.

## Expected behavior

BTC balances are rendered with the locale's decimal separator without crashing.

## Actual behavior

Jam displays its generic error page with `TypeError: u is not iterable`.

## Environment

- JoinMarket NG: `0.32.0`
- Jam frontend bundled with the JoinMarket NG release
- Locale: `it-IT`

dajiaohuang:
Locale-formatted BTC values (e.g. '0,26120925' for it-IT) caused a crash when splitting on '.' because the comma decimal separator left fractionalPart undefined.

Normalize the formatted number string to replace locale decimal separators with '.' before splitting.

Fixes #1289

### DIFF
--- a/src/components/ui/jam/Balance.tsx
+++ b/src/components/ui/jam/Balance.tsx
@@ -67,10 +67,16 @@ const BitcoinBalance = ({
   ...props
 }: BitcoinBalanceProps) => {
   const numberString = formatBtc(satsToBtc(String(value)))
-  // Normalise locale-specific decimal separators (e.g. "," in it-IT) to "."
+  // Normalise locale-specific decimal separators to "." and strip
+  // thousands separators (e.g. it-IT "12.345,67890123" → "12345.67890123").
   const normalizedString = numberString.replace(
-    /^([^.,]*)[.,](\d+)$/,
-    (_m, intPart, fracPart) => `${intPart}.${fracPart}`
+    /^([\d.,]*[.,])?([^.,]*)[.,](\d+)$|^([^.,]+)$/,
+    (_m, _preamble, intPart, fracPart, _whole) => {
+      if (intPart !== undefined) {
+        return `${intPart.replace(/[.,]/g, "")}.${fracPart}`
+      }
+      return _m  // no decimal separator, leave as-is
+    }
   )
   const [rawIntegerPart, fractionalPart] = normalizedString.split(DECIMAL_POINT_CHAR)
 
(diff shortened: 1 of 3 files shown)


fr00ller:
Thanks for the quick fix — it addresses the exact it-IT value I reported (0,26120925).

One concern is that the regex only matches formatted values containing a single . or , and ASCII fractional digits. For example, an it-IT grouped value such as 12.345,67890123 is not normalized; the following split('.') then treats the thousands separator as the decimal separator.

Locales using a different decimal separator can still trigger the original error. For example, ar-EG may produce ١٢٬٣٤٥٫٦٧٨٩٠١٢٣, for which fractionalPart remains undefined.

Because formatBtc() uses the browser locale via toLocaleString(undefined, ...), this is not limited to the languages supported by Jam.

Would you consider using Intl.NumberFormat.formatToParts() instead? It provides semantic integer, group, decimal, fraction, and sign parts without guessing separators, while preserving localized output.
I have a locally tested implementation and regression tests for it-IT and ar-EG, and I would be happy to share the diff or help adapt it to this PR.

theborakompanioni:
@dajiaohuang Nice, thanks! Can you format (`npm run format`) and push? Otherwise lgtm! Awesome :muscle:

theborakompanioni:
Formatting the integer part now differs from before. However, this feels okay for now, as it will most likely only concern a very limited number of users.

## Before
image

## After
image

theborakompanioni (dismissed):

theborakompanioni:
Okay for you as well @fr00ller ? :pray:

theborakompanioni (dismissed):

theborakompanioni (approved):

fr00ller:
> Okay for you as well @fr00ller ? 🙏

Yes, this is okay with me as well. The integer-part formatting change is acceptable from my side.
Thanks everyone for the quick fix and for handling this so promptly.

PR #1295, opened 2026-06-21
Fix orderbook own-offer pin/highlight with pagination

fr00ller:
## Summary

- apply own-offer row pinning against the pre-pagination row model
- apply own-offer row selection/highlighting against the pre-pagination row model
- keep the existing filtering/search behavior intact


theborakompanioni (approved):
@fr00ller lgtm! thanks! 🙏

PR #1296, opened 2026-06-23
fix(login): wallet preselection

theborakompanioni:
Fixes wallet preselection (for joinmarket-ng backend).
joinmarket-clientserver provides 'None' as `wallet_name` in session responses, whereas joinmarket-ng returns `''` (empty string).
This PR ensures that either the active wallet is preselected or a single wallet ending with `.jmdat` if present.


kishore08-07 (approved):
LGTM 🚀

PR #1298, opened 2026-06-26
dev(regtest): rectify tor setup and upgrade to core v29

theborakompanioni:
This PR updates bitcoind to v29.0 and fixes the tor setup for legacy clientserver instances in the regtest environment.

Note: First attempt was to upgrade to v30 directly. However, it seems v0.9.12 still was not able to start (`WARNING: Cannot import descriptor without private keys to a wallet with private keys enabled`, but, if private keys are disabled, there are other errors).
This should be the last regtest setup with support for clientserver, after that, let's upgrade to Core v30 and remove clientserver completely.


kishore08-07 (approved):
Lgtm 🚀

PR #1305, opened 2026-07-02
refactor: migrate API client to joinmarket-ng-api-ts 

kishore08-07:
## Summary
This PR implements the migration from the legacy `@joinmarket-webui/joinmarket-api-ts` API client package to  `@joinmarket-webui/joinmarket-ng-api-ts`.

> Note:
> Sweeping is now a non-atomic two-step flow (creating the plan, then starting it). If the start call fails after the plan is successfully created, the plan remains on the backend. However, subsequent retries will automatically overwrite it due to the force: true flag

## Key Changes

- Global Error Type refactoring (replacing `ErrorMessage` with `unknown / FastAPI-style errors`).

- `SweepPage` adaptations around Tumbler API endpoints:
   - Adapted `SweepPage` to utilize the new Tumbler API endpoints: `tumblerplan,` `tumblerstart`, `tumblerstatus`, and `tumblerstop` (replacing the old `getschedule` / `runschedule` / `stopcoinjoinoptions` endpoints).
   - Implemented a data-shape bridge in `SweepPage.tsx` (toSchedule mapper) that maps the new `TumblerPlanResponse` phases into the existing 7-tuple ScheduleEntry formats.
   - Rewrote mock queries and mutations in `SweepPage.test.tsx` to match the new Tumbler API methods.
- Test updates to mock the new endpoints and packages.


src/components/import/ImportStepImportDetails.tsx
```diff
@@ -15,7 +15,7 @@ type ImportStepImportDetailsProps = ComponentProps<typeof ImportDetailsForm> & {
 
 export const ImportStepImportDetails = ({ sessionInfo, onBack, ...importFormProps }: ImportStepImportDetailsProps) => {
   const { t } = useTranslation()
-  const isSessionActive = sessionInfo?.session === true
+  const isSessionActive = sessionInfo?.session === true && !!sessionInfo.wallet_name
```
theborakompanioni: Odd. Why depend additionally on wallet_name?

src/components/import/ImportStepWalletDetails.tsx
```diff
@@ -13,7 +13,7 @@ type ImportStepWalletDetailsProps = ComponentProps<typeof CreateWalletForm> & {
 }
 
 export const ImportStepWalletDetails = ({ sessionInfo, ...createFormProps }: ImportStepWalletDetailsProps) => {
-  const isSessionActive = sessionInfo?.session === true
+  const isSessionActive = sessionInfo?.session === true && !!sessionInfo.wallet_name
```
theborakompanioni: Same here. Why?

src/components/layout/AppNavbar.tsx
```diff
@@ -60,7 +60,7 @@ const WalletPreview = ({
           <div className="flex min-h-5 min-w-[150px] items-center">
             {rescanInfo?.rescanning === true ? (
               <div className="cursor-wait motion-safe:animate-pulse">
-                {rescanInfo.progress !== undefined
+                {rescanInfo.progress != null
```
theborakompanioni: `isValidNumber`?

package.json
```diff
@@ -71,7 +71,7 @@
   },
   "dependencies": {
     "@hookform/resolvers": "5.4.0",
-    "@joinmarket-webui/joinmarket-api-ts": "0.3.0",
+    "@joinmarket-webui/joinmarket-ng-api-ts": "^1.0.0",
```
theborakompanioni: Please pin it to the exact version.

src/components/layout/AppNavbar.tsx
kishore08-07: Good catch! I've replaced it with `isValidNumber(rescanInfo.progress)`.

package.json
kishore08-07: Done.

src/components/import/ImportStepImportDetails.tsx
kishore08-07: In `joinmarket-ng-api-ts` types, `wallet_name` is generated as optional (`wallet_name?: string`). so added `&& !!sessionInfo.wallet_name` check which ensures we only render `<OtherWalletActiveAlert />` when `wallet_name` is actually defined. Without it, the `isSessionActive`  could still render and pass undefined as `WalletFileName`. The `as WalletFileName` cast only satisfies TypeScript and it doesn't prevent undefined at runtime.

theborakompanioni: Yes, I understand. This is an edge case, but `session` can still be true, so import flow should be prevented. Better to show "Unknown wallet active"?

kishore08-07: Agreed, reverted the additional check.

src/components/earn/report/hooks/useQueryYieldgenReport.ts
```diff
@@ -11,23 +10,15 @@ export type { EarnReportEntry } from '@/components/earn/report/hooks/earnReportP
...
-        // Wrong type generated by API! Try to parse actual format `{ yigen_data: [] }`!
-        // see: https://github.com/JoinMarket-Org/joinmarket-clientserver/pull/1836 (last checked: 2026-03-09)
-        const data = response.data as unknown
-        if (data && typeof data === 'object' && 'yigen_data' in data && Array.isArray(data.yigen_data)) {
-          lines = (data.yigen_data ?? []) as string[]
-        } else {
-          lines = response.data ?? []
-        }
+        const data = response.data
+        const lines = Array.isArray(data) ? data : (data?.yigen_data ?? [])
```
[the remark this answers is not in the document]
kishore08-07: Resolved in [c9710ef](https://github.com/joinmarket-webui/jam/pull/1305/commits/c9710ef8823c2d278f9066372c2fed60926fc03b).

src/context/JamSessionInfoContextProvider.tsx
```diff
@@ -95,12 +95,12 @@ export const JamSessionInfoContextProvider = ({
     setRescanInfo({
```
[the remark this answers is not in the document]
kishore08-07: Out of scope for this PR.

src/components/sweep/SweepPage.tsx
```diff
@@ -138,41 +169,30 @@ export const SweepPage = ({ walletFileName }: SweepPageProps) => {
...
-      })
-
-      if (result.error !== undefined) {
-        if (result.response.status === 404) {
-          return null
-        }
-        throw new Error(getErrorReason(result.error, 'Failed to load schedule'))
+    retry: false,
+    select: (data: TumblerPlanResponse): { schedule: Schedule } | null => {
+      if (data.status.toLowerCase() === 'pending') {
```
parrth20: the status select only filters 'pending', so terminal/stale/errored plans render as a running schedule. (PLAUSIBLE)

The old getScheduleQuery returned null on 404 and only produced a schedule when getschedule actually returned a schedule array. The new select returns a schedule for every status except 'pending':

```
select: (data) => {
  if (data.status.toLowerCase() === 'pending') return null
  return { schedule: toSchedule(data) }
}
```

src/components/sweep/SweepPage.tsx
```diff
@@ -138,41 +169,30 @@ export const SweepPage = ({ walletFileName }: SweepPageProps) => {
...
-
-  const stopScheduleQuery = useQuery({
-    ...stopScheduleQueryOptions,
-    enabled: false,
+  const startTumblerPlanMutation = useMutation({
+    ...tumblerstartMutation({ client }),
     retry: false,
-    staleTime: 1,
-    gcTime: 1,
   })
```
parrth20: small cleanup `createTumblerPlanMutation` and `startTumblerPlanMutation` are useMutation wrappers whose state/reset are never used.

src/components/sweep/SweepPage.tsx
```diff
@@ -181,18 +201,31 @@ export const SweepPage = ({ walletFileName }: SweepPageProps) => {
...
+        body: {
+          destinations: args.body.destinations,
+          parameters: args.body.parameters,
+          force: true,
+        },
+      })
+
+      return await startTumblerPlanMutation.mutateAsync({
+        path: args.path,
+      })
```
parrth20, line 213: start is now a non-atomic two-step (plan then start). 
So If tumblerstart fails after tumblerplan succeeds, a plan is left created on the backend and onError surfaces the failure. The force: true on the plan call mitigates this on retry, but it's a semantic change from the old single runschedule call worth a sentence in the PR description.

parrth20 (commented):
migration changes (new) API returns FastAPI errors shaped { detail: ... } , but `src/lib/errorReason.ts` `normalizeAppError` (line ~31) only reads .message/.error_description. So all error reasons now fall back to "Unknown error". extend it to parse detail.

src/components/sweep/SweepPage.tsx
```diff
@@ -206,27 +239,20 @@ export const SweepPage = ({ walletFileName }: SweepPageProps) => {
...
     },
   })
 
-  const sessionSchedule = isScheduleValue(jmSession?.schedule) ? jmSession.schedule : undefined
-  const queriedSchedule = isScheduleValue(getScheduleQuery.data?.schedule) ? getScheduleQuery.data.schedule : undefined
-  const currentSchedule = sessionSchedule ?? queriedSchedule ?? localSchedule
+  const currentSchedule = isScheduleValue(getScheduleQuery.data?.schedule) ? getScheduleQuery.data.schedule : undefined
 
   const schedulerRunning = jmSession?.coinjoin_in_process === true && currentSchedule !== undefined
   const singleCoinJoinRunning = jmSession?.coinjoin_in_process === true && !schedulerRunning
```
parrth20: `currentSchedule` now comes only from the `tumblerstatus` poll the old optimistic `localSchedule` and `jmSession.schedule` are both gone.

In the gap right after starting a sweep (`coinjoin_in_process === true`, but the plan is still `pending` or the first poll hasn't resolved), `currentSchedule` is `undefined`, so:

- `schedulerRunning` → `false`
- `singleCoinJoinRunning` → `true`

…and we render the "coinjoin already running" warning instead of the sweep progress bar.

Can we keep an optimistic schedule (or treat `pending` as "scheduler starting") until the poll catches up, so the wrong alert doesn't flash?

parrth20 (dismissed):
Nice migration the mechanical parts (import renames, `unknown` error typing, null-guards) look clean.

One blocker before merge: with the new FastAPI `{ detail }` errors, `normalizeAppError` doesn't read `detail`, so every reason degrades to "Unknown error" (see inline comment). Worth a test with a `{ detail }` error.

Also a minor `SweepPage` alert flash + a couple inline nits. Happy to re-review once the error handling is sorted 🙏

src/components/sweep/SweepPage.tsx
kishore08-07: Thanks, I've added a note in the PR description.

src/components/sweep/SweepPage.tsx
kishore08-07: Good catch! I've updated the select logic to return null for any status that is not `running` or if the plan has been flagged as stale (`stale === true`).
Now this correctly resets the UI to the normal sweep form when a schedule is no longer active.

src/components/sweep/SweepPage.tsx
kishore08-07: Done. I've removed the intermediate `useMutation` hooks and now directly invoke the `tumblerplan` and `tumblerstart` from the sdk with `throwOnError: true` inside the main mutation.

kishore08-07:
> migration changes (new) API returns FastAPI errors shaped { detail: ... } , but `src/lib/errorReason.ts` `normalizeAppError` (line ~31) only reads .message/.error_description. So all error reasons now fall back to "Unknown error". extend it to parse detail.

Fixed, I have extended `normalizeAppError` to parse the fastAPI `{ detail }` field. It supports both string details and array-based validation errors, also added test cases in errorReason.test.ts to cover these shapes and verify fallback precedence.

src/components/sweep/SweepPage.tsx
kishore08-07: Nice suggestion. I have added guards for both `isWaitingSchedulerStart` and `waitingForTumblerStatus` (checks if the status query `isPending` while a coinjoin is starting) to the `singleCoinJoinRunning` check.
This prevents the warning alert from flashing and shows only the loading state until the first status poll resolves.

src/components/earn/report/hooks/useQueryYieldgenReport.ts
```diff
@@ -11,23 +10,15 @@ export type { EarnReportEntry } from '@/components/earn/report/hooks/earnReportP
...
-        } else {
-          lines = response.data ?? []
-        }
+        const data = response.data
+        const lines = Array.isArray(data) ? data : Array.isArray(data?.yigen_data) ? data.yigen_data : []
 
-        // API returns string[] (CSV lines with header)
         return yieldgenReportToEarnReportEntries(lines)
       } catch (error: unknown) {
         // 404 is returned until the maker is started at least once
```
theborakompanioni, line 23: Is this still true for joinmarket-ng?

src/components/earn/report/hooks/useQueryYieldgenReport.ts
```diff
@@ -11,23 +10,15 @@ export type { EarnReportEntry } from '@/components/earn/report/hooks/earnReportP
...
-        // Wrong type generated by API! Try to parse actual format `{ yigen_data: [] }`!
-        // see: https://github.com/JoinMarket-Org/joinmarket-clientserver/pull/1836 (last checked: 2026-03-09)
-        const data = response.data as unknown
-        if (data && typeof data === 'object' && 'yigen_data' in data && Array.isArray(data.yigen_data)) {
-          lines = (data.yigen_data ?? []) as string[]
-        } else {
-          lines = response.data ?? []
-        }
+        const data = response.data
+        const lines = Array.isArray(data) ? data : Array.isArray(data?.yigen_data) ? data.yigen_data : []
```
theborakompanioni: This was a legacy workaround before, now the line is a little bit strange without any explanation. e.g. why can `data` be an array, when the type says otherwise?
Either we remove the workaround (as joinmarket-ng is the only supported backend for v2) or add a comment.

src/components/create/CreateStepWalletDetails.tsx
```diff
@@ -14,7 +14,7 @@ type CreateStepWalletDetailsProps = ComponentProps<typeof CreateWalletForm> & {
 export const CreateStepWalletDetails = ({ sessionInfo, ...createFormProps }: CreateStepWalletDetailsProps) => {
   return (
     <div className="space-y-2">
-      {sessionInfo?.session === true ? (
+      {sessionInfo?.session === true && sessionInfo.wallet_name ? (
```
theborakompanioni: Use `session == true` as the only source of truth if a session is active.
If no wallet name is provided, use a placeholder or different phrasing without wallet name (e.g. "Another wallet is active")

src/components/earn/report/hooks/useQueryYieldgenReport.ts
kishore08-07: Yes, that check was a legacy workaround to support the legacy clientserver backend (which returned a plain array) alongside the new ng backend.

src/components/create/OtherWalletActiveAlert.tsx
```diff
@@ -12,15 +12,18 @@ type OtherWalletActiveAlertProps = {
 }
 
 export const OtherWalletActiveAlert = ({ walletFileName, linkTarget }: OtherWalletActiveAlertProps) => {
+  const { t } = useTranslation()
+  const walletName = walletFileName
+    ? walletDisplayName(walletFileName)
+    : t('create_wallet.another_wallet', 'another wallet')
```
theborakompanioni: This approach does not work for all languages.

src/components/earn/report/hooks/useQueryYieldgenReport.ts
kishore08-07: Resolved in a8e15e8 by removing the legacy workaround.

src/components/create/CreateStepWalletDetails.tsx
kishore08-07: Resolved in [d57c3f7](https://github.com/joinmarket-webui/jam/pull/1305/commits/d57c3f77b2850c2623a919923ab85d5f8b61d061).
 Refactored the session verification logic across the create and import flow steps to use `session === true` as the sole source of truth for checking active sessions.

src/components/create/OtherWalletActiveAlert.tsx
kishore08-07: Resolved in [a025c4a](https://github.com/joinmarket-webui/jam/pull/1305/commits/a025c4aa2734aac62977bd4b52d2a8bde07511aa).

src/components/earn/report/hooks/useQueryYieldgenReport.ts
kishore08-07: Yes, it is true for joinmarket-ng.
https://github.com/joinmarket-ng/joinmarket-ng/blob/main/jmwalletd/src/jmwalletd/errors.py

theborakompanioni (approved):
lgtm

PR #1325, opened 2026-07-08
chore: add csp meta tag

theborakompanioni:
Related to https://github.com/joinmarket-webui/jam-docker/pull/203

Adds a Content-Security-Policy as in the standalone/ui-only image.

> When both HTTP headers and meta tags are present, browsers will combine both policies using the most restrictive rules from each.

### DIFF
--- a/index.html
+++ b/index.html
@@ -2,7 +2,7 @@
 <html lang="en" suppressHydrationWarning>
   <head>
     <meta charset="utf-8" />
-    <meta http-equiv="Content-Security-Policy" content="default-src 'self'; style-src 'self' 'unsafe-inline'; img-src 'self' data:; frame-src 'none'; object-src 'none';">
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -2,11 +2,12 @@
 <html lang="en" suppressHydrationWarning>
   <head>
     <meta charset="utf-8" />
+    <meta http-equiv="Content-Security-Policy" content="default-src 'self'; style-src 'self' 'unsafe-inline'; img-src 'self' data:; frame-src 'none'; object-src 'none';">
```
kishore08-07:
```suggestion
    <meta http-equiv="Content-Security-Policy" content="default-src 'self'; style-src 'self' 'unsafe-inline'; img-src 'self' data:; frame-src 'none'; object-src 'none'; base-uri 'self'; form-action 'self';">
```

theborakompanioni: Isn't this already covered by `default-src 'self'`?

kishore08-07: no, `default-src` only acts as a fallback for CSP fetch directives,
Both `base-uri` and `form-action` are non-fetch CSP directives, so they do not inherit from `default-src`.

kishore08-07 (approved):

PR #1326, opened 2026-07-08
chore(ui): add fonts

theborakompanioni:
Adds [Inter](https://github.com/rsms/inter) and [JetBrains Mono](https://github.com/JetBrains/JetBrainsMono) and sets them as default fonts (for sans-serif and monospace respectively).

This is just a concept and open for debate.

## :camera_flash:  Before vs After

image
image

image
image

image
image

image
image


kishore08-07 (approved):
We can optimize the bundle size in a follow-up by switching Inter to use the `InterVariable.woff2` variable font. That would help us remove the static Inter woff2 font files, reducing the font folder size without any visual quality loss.

PR #1337, opened 2026-07-11
feat: display alert and reload balance on tx message via websocket

### LINKED ISSUE #1307: feat: display alert and reload balance on tx message via websocket
On incoming tx message via websocket, the utxo information should be reloaded.
This way the balance stays in sync, when the tx was not made via the ui (e.g. during earn, or just an external deposit).
An info alert via sonner can be shown briefly (e.g. "New Transaction xyz").

parrth20:
Closes [#1307](https://github.com/joinmarket-webui/jam/issues/1307)

On incoming tx messages via websocket, the wallet info is now reloaded (new AFTER_TX_MESSAGE trigger in WalletInfoAutoReload, same pattern as the existing triggers) and a brief info toast "New transaction: xxxx…xxxx" is shown.

The toast is only shown for txids not already present in jmTxStore transactions made via the UI are added to the store before the websocket echoes them, so they don't produce a duplicate alert. This also dedupes repeated notifications for the same tx.

Verified on regtest by triggering a direct-send via the API (outside the UI): toast appears and balance updates without user interaction. 

Note: the backend currently only emits tx messages from the coinjoin/direct-send paths, not for external deposits the frontend handles those messages whenever the backend starts sending them.
image

### DIFF
--- a/src/App.test.tsx
+++ b/src/App.test.tsx
@@ -67,6 +67,10 @@ vi.mock('./store/jmSessionStore', () => ({
   jmSessionStore: { getState: () => ({ state: holders.jmSession }) },
 }))
 
+vi.mock('./store/jmTxStore', () => ({
+  jmTxStore: { getState: () => ({ state: {} }) },
+}))
+
 vi.mock('@/store/jamSettingsStore', () => ({
   jamSettingsStore: { getState: () => ({ state: { developerMode: holders.developerMode } }) },
   useDeveloperMode: () => ({ enabled: holders.developerMode }),
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -55,6 +55,7 @@ import { useJamWalletInfoContext } from './context/JamWalletInfoContext'
 import { JmWebsocketContextProvider } from './context/JmWebsocketContextProvider'
 import { getErrorReason } from './lib/errorReason'
 import { jmSessionStore } from './store/jmSessionStore'
+import { jmTxStore, type JmTxInfos } from './store/jmTxStore'
 import type { Milliseconds } from './types/global'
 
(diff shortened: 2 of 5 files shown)


parrth20:
ping @theborakompanioni

theborakompanioni (approved):
ACK. However, it seems the backend does not send messages when a new tx is registered during an active maker session. Can you verify?
If that is the case, please open an issue on the joinmarket-ng repository.
This feature is only viable if works for transactions that are not initiated by the user (otherwise the UI already knows to reload information).

Additionally, it seems there is also no message after successful collaborative transactions that **are indeed** initiated by the user. Right?

parrth20:
> ACK. However, it seems the backend does not send messages when a new tx is registered during an active maker session. Can you verify? If that is the case, please open an issue on the joinmarket-ng repository. This feature is only viable if works for transactions that are not initiated by the user (otherwise the UI already knows to reload information).
> 
> Additionally, it seems there is also no message after successful collaborative transactions that **are indeed** initiated by the user. Right?

Verified you're right on both counts. In joinmarket-ng's jmwalletd, the only {txid, txdetails} broadcast is inline in the direct-send endpoint (routers/coinjoin.py). There is no wallet transaction-monitor callback, so no websocket message is sent for txs registered during an active maker session (verified on regtest: an incoming external tx produces no message, neither unconfirmed nor on confirmation), and none after successful collaborative transactions either the taker task only broadcasts coinjoin_state on completion.

For comparison, joinmarket-clientserver registers sendTxNotification as a WalletService callback (jmclient/wallet_rpc.py → websocketserver.py), so it notifies on every wallet tx that's the behavior this PR's frontend side was built against (and it works end-to-end wherever the backend does send the message, e.g. direct-send).

PR #1339, opened 2026-07-11
feat: plan scheduled sweep

theborakompanioni:
- adds options to sweep
- improve schedule entries view
- ability to preview schedule plan before start

# :camera_flash: 

image
image

image

image


src/components/send/SendPage.tsx
```diff
@@ -516,15 +520,17 @@ export const SendPage = ({ walletFileName }: SendPageProps) => {
...
-              <div>
-                <Button
-                  variant="outline"
-                  onClick={() => setShowAbortCoinjoinDialog(true)}
-                  disabled={isWaitingCoinjoinStop}
-                >
-                  {t('global.abort')}
-                </Button>
-              </div>
+              {takerRunning && !schedulerRunning ? (
```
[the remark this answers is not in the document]
theborakompanioni: While this is true, we want to be explicit here.

src/components/sweep/SweepFormSchema.ts
```diff
@@ -1,25 +1,70 @@
...
 }
 
 export type SweepResolverContext = {
   addressSummary: AddressSummary
 }
 
+export const MIN_MIN_NUMBER_OF_COLLABORATORS = Math.max(1, JM_MINIMUM_MAKERS_DEFAULT)
+export const MAX_MAX_NUMBER_OF_COLLABORATORS =
+  MIN_MIN_NUMBER_OF_COLLABORATORS +
+  pseudoRandomInteger(MIN_MIN_NUMBER_OF_COLLABORATORS * 4, MIN_MIN_NUMBER_OF_COLLABORATORS * 5 + 1)
```
parrth20: Would you mind capping this at the taker’s supported maximum of 20 and mirroring that bound in the schema? This expression currently produces 20-25. The planner accepts those values, but a phase with `counterparty_count > 20` fails TakerConfig validation only when that phase executes, and retrying preserves the same count. Keeping both slider and schema at or below 20 would prevent plans that cannot complete.

src/components/sweep/preconditions.ts
```diff
@@ -16,11 +17,11 @@ export interface SweepPreconditionSummary {
 
 const DEFAULT_OPTIONS: SweepPreconditionOptions = {
   minNumberOfUtxos: 1,
-  minConfirmations: 5,
+  minConfirmations: JM_TUMBLER_MIN_CONFIRMATIONS_BETWEEN_PHASES,
```
parrth20: Could the initial preflight use `JM_TAKER_UTXO_AGE` instead? min_confirmations_between_phases applies after a tumbler broadcast, while initial taker input selection uses the five-confirmation taker age. At exactly five confirmations the backend can start, but this guard blocks the UI for another block.

src/components/sweep/SweepPage.tsx
```diff
@@ -279,57 +225,24 @@ export const SweepPage = ({ walletFileName }: SweepPageProps) => {
   const isOperationDisabled =
     feeConfigValidation.maxFeesConfigMissing ||
     collaborativeOperationRunning ||
-    jmSession?.rescanning ||
+    rescanInfo.rescanning ||
     !preconditionSummary.isFulfilled
 
   const isStartDisabled =
```
parrth20, line 232: Could we revalidate the pending plan’s destinations against the current `addressSummary` before enabling/confirming Start? An own-wallet address can be new during planning and become used while the plan is pending; the form is then unmounted and the backend starts the persisted plan without JAM’s reuse policy. Re-running `buildDestinationErrors` here would preserve the previous flow’s safety.

src/components/sweep/SweepPage.tsx
```diff
@@ -279,57 +225,24 @@ export const SweepPage = ({ walletFileName }: SweepPageProps) => {
...
-  const onOpenScheduleConfirm = async () => {
-    if (isOperationDisabled || isWaitingSchedulerStart || isWaitingSchedulerStop) {
-      return
-    }
-
-    await form.handleSubmit(() => {
-      setShowScheduleConfirmDialog(true)
-    })()
+    await startScheduleMutationMutateAsync({
+      path: { walletname: walletFileName },
```
parrth20, line 245: Could we invalidate or rebuild a pending plan when its source UTXO set changes before Start????

src/components/sweep/SweepPage.tsx
```diff
@@ -232,29 +144,63 @@ export const SweepPage = ({ walletFileName }: SweepPageProps) => {
...
       setAlertMessage(message)
       toast.error(message)
     },
   })
 
-  const currentSchedule = isScheduleValue(getScheduleQuery.data?.schedule) ? getScheduleQuery.data.schedule : undefined
+  const currentSchedule = useMemo(() => {
+    if (getScheduleQuery.error) return
+    if (getScheduleQuery.data === undefined) return
+    if (getScheduleQuery.data.stale === true) return
```
parrth20, line 192: Could we keep this response visible instead of dropping it? cos the the backend’s first post-restart response is `stale:true`, `status:'failed`, with the restart error. Dropping it exposes the fresh-plan form, whose `force:true` request can overwrite the failed plan before the user sees or acknowledges it.

src/components/sweep/scheduleUtils.ts
```diff
@@ -1,251 +1,326 @@
...
 
-  if (state === 1) {
-    return 'confirmed'
+  const totalEstimatedDurationBetweenPhasesInSeconds: Seconds = Math.ceil(
+    Math.max(
+      1,
+      entries.slice(0, Math.max(0, entries.length - 1)).reduce((acc, entry) => {
+        const fixedRuntimeInSeconds = Math.max(
+          MIN_ESTIMATE_ACTION_RUNTIME_IN_SECONDS,
+          Math.max(entry.durationSeconds ?? 0, entry.idleTimeoutSeconds ?? 0),
```
parrth20: the ETA does not follow runner semantics. Maker sessions end at the duration deadline or idle timeout, whichever occurs first, but this uses max. With shipped defaults it counts 48 hours instead of the 12-hour hard duration for every maker phase.
Could we make this estimate follow the runner’s ordering?

src/components/sweep/SweepForm.tsx
```diff
@@ -0,0 +1,316 @@
...
+            ? JAM_SWEEP_DESTINATION_ADDRESSES_MIN_COUNT
+            : JAM_SWEEP_DESTINATION_ADDRESSES_DEFAULT_COUNT
+        }
+        register={register}
+        setValue={setValue}
+        formState={formState}
+        fields={destinationsFieldArray.fields}
+        disabled={disabled}
+        onClickAppend={() => {
+          destinationsFieldArray.append(buildSweepDestinationValues(1), { shouldFocus: false })
```
parrth20, line 171: can we cap this field array and enforce the same maximum in the schema at the usable stage-2 mixdepth count??

parrth20 (dismissed):
Everything can be done in a follow-up, ACK !!

parrth20:
@theborakompanioni lint errors !!

src/components/sweep/scheduleUtils.ts
theborakompanioni: Fixed in 95b3426b6277e784eea79b69a95ccafbb7878b7e

src/components/sweep/SweepPage.tsx
theborakompanioni: Will be done in a follow-up :muscle:

theborakompanioni: See https://github.com/joinmarket-webui/jam/issues/1346

src/components/sweep/SweepPage.tsx
theborakompanioni: Good idea. Follow-up! See https://github.com/joinmarket-webui/jam/issues/1347

src/components/sweep/SweepPage.tsx
theborakompanioni: Maybe this is something that the backend should enforce also, what do you think?
We might not always have this information. Part of it can be done in https://github.com/joinmarket-webui/jam/issues/1347 (when Jam triggered plan creation and knows about the present utxos)

src/components/sweep/preconditions.ts
theborakompanioni: Done!

PR #1343, opened 2026-07-12
chore(ui): improve tx tags colors

theborakompanioni:
## :camera_flash:  Before / After
image

image

Most imporantly: Change outputs should not use a "success"-style color.


parrth20 (approved):
LGTM !

PR #1350, opened 2026-07-25
refactor: improve balance summary calculation

theborakompanioni:
Calculate balances by looping through UTXO array only once.


parrth20 (approved):
LGTM 🚀 tACK

PR #1367, opened 2026-07-28
fix(receive): block new addresses during rescan

### LINKED ISSUE #1361: fix: prevent loading new addresses while rescanning
Prevent loading new addresses while rescanning.
During rescan, the backend might provide addresses that have already been used as "new".
This might only be an issue the first time rescanning, but it's best to play it safe.

parrth20:
### Summary

- Prevent new address requests while a wallet rescan is active.
- Disable the reveal and refresh address buttons during rescanning.
- Add regression coverage for the rescan state.

Closes #1361

### DIFF
--- a/src/components/receive/ReceivePage.test.tsx
+++ b/src/components/receive/ReceivePage.test.tsx
@@ -34,6 +34,7 @@ const mocks = vi.hoisted(() => {
 
   return {
     developerMode: false,
+    rescanning: false,
     getAddress: vi.fn(),
     defaultJars,
     jars: defaultJars,
@@ -107,6 +108,12 @@ vi.mock('@/context/JamWalletInfoContext', () => ({
   }),
 }))
 
+vi.mock('@/context/JamSessionInfoContext', () => ({
+  useRescanStatus: () => ({
+    rescanInfo: { rescanning: mocks.rescanning },
+  }),
+}))
+
 vi.mock('@/hooks/useApiClient', () => ({
   useApiClient: () => ({}),
 }))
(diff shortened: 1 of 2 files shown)


src/components/receive/ReceivePage.test.tsx
```diff
@@ -175,6 +183,16 @@ describe('ReceivePage', () => {
...
   })
 
+  it('prevents loading addresses while rescanning', () => {
+    mocks.rescanning = true
+
+    render(<ReceivePage walletFileName="wallet.jmdat" />)
+
+    expect(screen.getByRole('button', { name: 'receive.button_reveal_address' })).toBeDisabled()
+    expect(screen.getByRole('button', { name: 'receive.button_new_address' })).toBeDisabled()
+    expect(mocks.getAddress).not.toHaveBeenCalled()
```
theborakompanioni, line 194: Of course it's good to verify that `getAddress` was not called, but I guess the button - even if disabled - should actually be clicked for the assumption to have a chance of catching a bug.

src/components/receive/ReceivePage.tsx
```diff
@@ -107,6 +109,7 @@ export const ReceivePage = ({ walletFileName }: ReceivePageProps) => {
   }
 
   const fetchNewAddress = async () => {
+    if (rescanInfo.rescanning) return
```
theborakompanioni, line 112: Can this be removed or externalized into a variable (e.g. `!operationEnabled`) so there is a single place to change, should there be further restrictions?

e.g. imho this might also be true if the maker is running, since collaborative tx outputs come from purpose "receive"  (`m/84'/0'/{jarIndex}'/0/x`) instead of the "change" (`m/84'/0'/{jarIndex}'/1/x`).

Is that taken care of in joinmarket-ng (maker does not use address already shown to the user and vice versa) @m0wer ?

theborakompanioni (dismissed):
Only NITs -> ACK.

src/components/receive/ReceivePage.test.tsx
parrth20: Agreed I’ll click the disabled button in the test before verifying that `getAddress` was not called.

theborakompanioni (approved):
reACK.

PR #1369, opened 2026-07-28
fix(receive): save sourceJarIndex before async operation

theborakompanioni:
See title. Since the mixdepth is not included in the response, this makes sure the `sourceIndexJar` is actually the one the address is requested from.


kishore08-07 (dismissed):
LGTM 🚀

PR #1370, opened 2026-07-28
chore(send): jars with frozen funds are selectable

### LINKED ISSUE #1362: ui(send): click on selected jar should open utxo dialog
See title.

### LINKED ISSUE #1363: ui(send): all jars containing frozen utxos should be selectable
If a user deselects all utxos during "utxo selection" he ends up having an invalid form state (your jar does not have any spendable utxos) and must navigate to the Jar Details Overlay to resolve the issue. Or a user wants to spend from a Jar that initially has no spendable balance, but some frozen utxos.  

It should be possible to select a jar that has zero spendable balance, but frozen utxos.
Maybe the utxo selection can be automatically opened if a jar in such a state is selected.

theborakompanioni:
Resolves #1363.
Resolves #1362.

Makes jars with ("available") frozen funds selectable and opens Jar Selection Dialog on click.


src/components/send/SendForm.tsx
```diff
@@ -298,41 +302,58 @@ export function SendForm({
...
-                      }
-                    }}
-                    disabled={disabled || jar.balanceSummary.calculatedAvailableBalanceInSats <= 0}
-                  />
-                ))}
+              <div className="flex flex-1 flex-row flex-wrap items-start justify-around gap-8">
+                {jars.map((jar, index) => {
+                  const noFunds =
+                    jar.balanceSummary.calculatedAvailableBalanceInSats <= 0 &&
+                    jar.balanceSummary.calculatedAvailableFrozenBalanceInSats <= 0
```
parrth20, line 317: `calculatedAvailableFrozenBalanceInSats` also includes frozen expired fidelity bonds, but the UTXO selector excludes all fidelity bonds. A jar containing only such a bond becomes selectable, then asks the user to select UTXOs while offering nothing selectable. Could this condition use frozen non-fidelity UTXOs so every enabled jar can actually be repaired through the dialog?

theborakompanioni: Good point! I guess the issue is with the jar selector then? It should be able to unfreeze expired fidelity bonds
However, an expired fidelity bond output cannot be spent in a collaborative transaction - this must be taken care of.
Okay for you if this is done in a follow-up?

parrth20 (approved):
LGTM

PR #1379, opened 2026-07-29
fix: disable freeze/unfreeze when maker/taker/rescan is active

### LINKED ISSUE #1366: fix: freeze/unfreeze should be disabled while rescanning
See title.

At best a wallet is not used during rescan (e.g. supposedly empty addresses shown might end up being already used; wrong labels, etc.).
Can be revisited, since it should be safe to freeze/unfreeze utxos, when they are not moved anyway (not send operation, no maker session, etc.)-

theborakompanioni:
Resolves #1366.

See title.


parrth20 (approved):
ACK

PR #1389, opened 2026-07-30
chore(ci): pin ci actions

theborakompanioni:
See title.
Pins workflow action dependencies by hash.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -41,7 +41,9 @@ jobs:
 
     steps:
     - name: Checkout repository
-      uses: actions/checkout@v5
+      uses: actions/checkout@v7
+      with:
+        persist-credentials: false
 
     # Initializes the CodeQL tools for scanning.
     - name: Initialize CodeQL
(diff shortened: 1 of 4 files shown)


.github/workflows/codeql-analysis.yml
```diff
@@ -41,7 +41,9 @@ jobs:
 
     steps:
     - name: Checkout repository
-      uses: actions/checkout@v5
+      uses: actions/checkout@v7
```
kishore08-07, line 44:
```suggestion
      uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
```

kishore08-07 (approved):

.github/workflows/codeql-analysis.yml
kishore08-07, line 50: a few deps in this file could be pinned? can do in a follow up...

theborakompanioni: I did not change codeql-analysis. I think it is disabled  (https://github.com/joinmarket-webui/jam/actions/workflows/codeql-analysis.yml) because it is overruled by the repository settings, guess the action was present before it because integrated into GH? Let's see what the best approach is. Can imagine that it will be just deleted..

.github/workflows/codeql-analysis.yml
theborakompanioni: Wont do. See below.

PR #1392, opened 2026-07-30
fix(wallet): keep jars visible and clickable during background refetch

### LINKED ISSUE #1382: fix(ui): ability to view jar details while refetching display data
Show a different loading indicator while `/display` request is refetched.

Currently, the user cannot access the Jar Details Overlay while loading (even if data is present, but stale).
Image

Loading should be indicated in a way such as that a user can still view jar details.

brian-soltani:
The jars section in MainWalletPage was gated on isFetching, which is true on every background refetch, not just the initial load. That meant any time wallet data refreshed in the background, the jars grid, and the Jar Details overlay it opens, would get replaced by a spinner even though the previously loaded jars were still valid to show and click. The balance display right above it already used isLoading for this same purpose, so this brings the jars section in line with that pattern.

I also updated the test that covered the old spinner behavior. It now asserts that a jar stays visible and clickable while isFetching is true, and that clicking refresh still triggers a refetch, instead of asserting a spinner appears.

Closes #1382.

### DIFF
--- a/src/components/MainWalletPage.test.tsx
+++ b/src/components/MainWalletPage.test.tsx
@@ -125,10 +125,10 @@ describe('MainWalletPage', () => {
     expect(refetch).toHaveBeenCalled()
   })
 
-  it('shows a jars spinner while fetching and refreshes on click', () => {
+  it('keeps jars visible and clickable while fetching in the background', () => {
     walletInfo = { ...walletInfo, isFetching: true }
     render(<MainWalletPage walletFileName={walletFileName} />)
-    expect(screen.getByTestId('spinner')).toBeInTheDocument()
+    expect(screen.getByText(jars[0].name)).toBeInTheDocument()
     fireEvent.click(screen.getByText('global.refresh'))
     expect(refetch).toHaveBeenCalled()
   })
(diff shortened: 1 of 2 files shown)


parrth20:
Hey @brisolt thanks for contributing !!

src/components/MainWalletPage.test.tsx
```diff
@@ -125,10 +125,10 @@ describe('MainWalletPage', () => {
     expect(refetch).toHaveBeenCalled()
   })
 
-  it('shows a jars spinner while fetching and refreshes on click', () => {
+  it('keeps jars visible and clickable while fetching in the background', () => {
     walletInfo = { ...walletInfo, isFetching: true }
     render(<MainWalletPage walletFileName={walletFileName} />)
-    expect(screen.getByTestId('spinner')).toBeInTheDocument()
+    expect(screen.getByText(jars[0].name)).toBeInTheDocument()
```
parrth20, line 131: The test name says the jar remains clickable, but it only verifies visibility and clicks Refresh. Could we click the jar here and assert that `WalletJarsDetailsOverlay` opens? That would directly cover the regression described in #1382.

parrth20 (dismissed):
Otherwise LGTM !!! 
@brisolt Thanks !

src/components/MainWalletPage.test.tsx
brian-soltani: Good catch, done. The test now clicks the jar and asserts WalletJarsDetailsOverlay opens with the right selected index, in addition to the refresh check.

parrth20 (approved):

PR #1394, opened 2026-07-31
fix(send): disallow send from a jar with unfrozen fidelity bond

### LINKED ISSUE #1378: fix: prevent maker/taker operation if only fb output present
- Prevent maker/taker operations, if only fb outputs are present (e.g. expired non-frozen).

Currently, a maker operation can be started, but will never load and stays stuck.
Image

Since the UI is stuck also think about:
- Ability to stop the maker even if the offer is not (yet) present/loaded (?)

How to reproduce:
Freeze all coins except one expired fb output and try starting the maker.

Counter test: Have one expired non-frozen fb output, and non-fb output, this should work (?) and the maximum should not exceed the non-fb-output. Please verify.

### LINKED ISSUE #1387: non-frozen expired fidelity bonds should prevent collaborative transactions
non-frozen expired fidelity bonds handling needs some care.
They cannot be spent in collaborative transactions, and Jam should make the user aware of that.

One method is to freeze any and all fidelity bonds, regardless of their expiry date, and make the user go through "unlock" or "renew".

Original comment by @parrth20 :

`calculatedAvailableFrozenBalanceInSats` also includes frozen expired fidelity bonds, but the UTXO selector excludes all fidelity bonds. A jar containing only such a bond becomes selectable, then asks the user to select UTXOs while offering nothing selectable. Could this condition use frozen non-fidelity UTXOs so every enabled jar can actually be repaired through the dialog?

_Originally posted by @parrth20 in https://github.com/joinmarket-webui/jam/pull/1370#discussion_r3677013394_

theborakompanioni:
Resolves #1378.
Resolves #1387.

- Prompts the user to freeze any Fidelity Bond before sending a transaction.
- Tries to freeze FB outputs after creating or renewing
(However, never "automatically" attempt to do so - only from internal operations "create" or "renew").

If a user imports a wallet with an existing FB, it might be that the backend does not automatically freeze it - Jam should not do it either.


parrth20 (commented):
the direction looks good to me!!! Perhaps the remaining precondition work could show unfrozen Fidelity Bonds as a distinct reason with a clear message, and it may also be worth checking the maker-start path for #1378 Please continue 👍

PR #1355, opened 2026-07-25
fix(sweep): validate destination addresses against wallet network

Ashish-Kumar-Dash:
## What does this PR do?

The sweep/scheduler form only validated destination addresses with `isValidAddress`, which is network-agnostic, a valid testnet/regtest address was accepted on mainnet and would only fail once the tumbler schedule was already running. The send form already guards against this via `destinationAddressField`'s `isAddressOnNetwork` check; the sweep schema had a `// TODO: use formValidation#destinationAddressField ?` marking the gap.

This adds the same network check to `sweepFormSchema`, wired from the existing `useDetectNetwork()` hook (already used by the send form), with a new `scheduler.feedback_destination_network_mismatch` string reusing the send form's error copy.

## Visual Demo (For contributors especially)

No UI change

PR #1395, opened 2026-08-01
fix(validation): distinguish testnet and regtest for bech32 addresses

Ashish-Kumar-Dash:
## What does this PR do?

Follow-up to #1355. That PR added wallet-network validation for destination addresses, and to avoid a regtest wallet rejecting its own legacy (base58) addresses, isAddressOnNetwork treated testnet and regtest as interchangeable. As flagged in review, that fallback was too broad: it also relaxed bech32 addresses, so a bcrt1… (regtest) address wrongly passed on a testnet wallet, and vice versa.

This tightens the fallback to only the ambiguous case:

Base58 (p2pkh/p2sh): testnet and regtest share version bytes, so bitcoin-address-validation always labels them testnet, these stay interchangeable.
Bech32/bech32m (tb1… vs bcrt1…): the HRP already identifies the network, so getAddressInfo returns the correct distinct network and these are now validated strictly (gated on !bech32).
Net effect: mainnet validation is unchanged (never ambiguous); only the wrong test-network variant of a segwit/taproot address is now correctly rejected.

### DIFF
--- a/src/lib/formValidation.test.ts
+++ b/src/lib/formValidation.test.ts
@@ -13,6 +13,9 @@ const mainnetAddress = '1BitcoinEaterAddressDontSend8MUo1T'
 const testnetAddress = 'mipcBbFg9gMiCh81Kj8tqqdgoZub1ZJRfn'
 const regtestBech32Address = 'bcrt1q6rz28mcfaxtmd6v789l9rrlrusdprr9pz3cppk'
 const testnetBech32Address = 'tb1qw508d6qejxtdg4y5r3zarvary0c5xw7kxpjzsx'
+// bech32m/taproot pair - same witness program, differing only by HRP (tb1p vs bcrt1p)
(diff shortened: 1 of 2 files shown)


parrth20:
@Ashish-Kumar-Dash, thanks for contributing!!!

review the copilot suggestions...If you think they should be implemented, feel free to make the changes o/w, please reply in the relevant threads with your proper reasoning.
ping me when you’re done, and I’ll take another look.

Ashish-Kumar-Dash:
> review the copilot suggestions...If you think they should be implemented, feel free to make the changes o/w, please reply in the relevant threads with your proper reasoning. ping me when you’re done, and I’ll take another look.

nope, copilot was right, I implemented the proposed fix!

parrth20:
@Ashish-Kumar-Dash would it be worth adding a tb1p…/bcrt1p… taproot pair here as well? Since this is intended to support both Bech32 and Bech32m, it could help lock that behaviour down.

Ashish-Kumar-Dash:
added!

parrth20 (approved):
LGTM 👍

Issue #1397, opened 2026-08-01
Fidelity bond UTXO stays permanently unfrozen if renew/move-to-jar sweep fails after unfreezing it


CapThunder19:
## Summary

`useFidelityBondSweep`  the hook shared by the **Renew Bond** and **Move to Jar** dialogs  unfreezes a fidelity bond's UTXO before sweeping it to its new destination. If the sweep transaction fails *after* the unfreeze step but before/during broadcast, the error-handling rollback restores only the *other* UTXOs it temporarily froze. It never re-freezes the bond UTXO itself. The bond is left unprotected, with no clear indication to the user that this happened.

---

## Location

[`src/components/earn/fidelity-bond/useFidelityBondSweep.ts`](https://github.com/joinmarket-webui/jam/blob/devel/src/components/earn/fidelity-bond/useFidelityBondSweep.ts), lines 58–137

Used by:
- `src/components/earn/RenewBondDialog.tsx`
- `src/components/earn/MoveToJarDialog.tsx`

---

## Steps to Reproduce

1. Have a matured (time-unlocked) fidelity bond that is currently frozen.
2. Open **Renew Bond** (or **Move to Jar**) and submit.
3. Internally: other UTXOs in the jar get frozen  the bond UTXO gets unfrozen  `directSend` is called to sweep the bond.
4. Have `directSend` fail at this point (network error, backend rejects the tx, insufficient fee, node restart, etc.).
5. Observe: the `catch` block re-freezes the other jar UTXOs from step 3, but **does not** re-freeze the bond UTXO unfrozen in step 3.

---

## Code Evidence

```ts
const frozen: Utxo[] = []
try {
  for (const u of utxosToFreeze) {
    await freezeUtxo.mutateAsync({ ... })
    frozen.push(u)
  }

  if (utxo.frozen) {
    await unfreezeUtxo.mutateAsync({          // bond unfrozen here — not tracked for rollback
      path: { walletname: walletFileName },
      body: { 'utxo-string': utxo.utxo, freeze: false },
    })
  }

  const result = await directSend.mutateAsync({ ... })   // if this throws...
  ...
} catch (_ignoredOnPurpose: unknown) {
  // Best-effort rollback — unfreeze UTXOs that were frozen before the error
  for (const u of frozen) {                    // only restores the OTHER utxos
    await unfreezeUtxo.mutateAsync({ ... })
  }
  return undefined                              // bond UTXO left unfrozen
}
```

## Expected Behavior
[...]

PR #1399, opened 2026-08-02
fix(fb): re-freeze fidelity bond utxo when renew/move sweep fails after unfreezing it

CapThunder19:
## Summary

`useFidelityBondSweep` — shared by the Renew Bond and Move to Jar dialogs freezes the jar's other UTXOs, unfreezes the target fidelity bond, then sweeps the whole jar. If the sweep fails *after* the bond was unfrozen, the error-handling rollback restored only the other, temporarily-frozen UTXOs , it never re-froze the bond UTXO itself. A failed renew/move could silently leave a matured fidelity bond permanently unfrozen and unprotected, with no indication to the user beyond a generic error toast.

This adds a `bondWasUnfrozen` flag, set only when the bond's unfreeze call actually succeeds, and refreezes the bond in the `catch` block when that flag is set mirroring the existing best-effort rollback pattern already used for the other UTXOs in this file.

## Why

This directly undermines the auto-freeze protections recently added for fidelity bonds, and the downstream mitigations that were built on top of them:

- #1394 disallows sending from a jar with an unfrozen bond
- #1387 blocks collaborative transactions with a non-frozen expired bond

Both react to a bond *already being* unfrozen but don't address how it can end up that way through this exact rollback gap. This PR fixes the root cause instead of relying solely on those downstream guards.

## Changes

- `src/components/earn/fidelity-bond/useFidelityBondSweep.ts`: track whether the bond was unfrozen during the sweep attempt and re-freeze it on failure.
- `src/components/earn/fidelity-bond/useFidelityBondSweep.test.ts` (new): this hook previously had no test coverage. Added three tests:
  - refreezes the bond when a failed sweep had unfrozen it (the regression case)
  - does not attempt to re-freeze the bond when it was never unfrozen (guards against over-eager rollback)
  - happy path: sweep succeeds, other jar UTXOs get frozen then restored

## Test plan

- [x] `npm run build` passes, no type errors
- [x] `npm run test:unit`  full suite passes for all fidelity-bond/renew/move-to-jar tests (the only 5 pre-existing failures are unrelated: locale dependent number formatting in `Balance.test.tsx`, unaffected by this change)
- [x] New tests specifically cover the failure path this PR fixes

solves #1397

### DIFF
--- a/src/components/earn/fidelity-bond/useFidelityBondSweep.ts
+++ b/src/components/earn/fidelity-bond/useFidelityBondSweep.ts
@@ -154,7 +154,8 @@ export function useFidelityBondSweep({
 
         return undefined
       } finally {
-        
+        // refetch only if the sweep broadcast, and don't let a refetch error
+        // get treated as a sweep failure
         if (sweepBroadcasted) {
           try {
             await walletInfoRefetch()
--- a/src/components/earn/fidelity-bond/useFidelityBondSweep.ts
+++ b/src/components/earn/fidelity-bond/useFidelityBondSweep.ts
@@ -56,6 +56,7 @@ export function useFidelityBondSweep({
       setError(undefined)
 
       const frozen: Utxo[] = []
+      let bondWasUnfrozen = false
       try {
         // Freeze other UTXOs in the source jar so only the FB gets swept
         for (const u of utxosToFreeze) {
@@ -71,6 +72,7 @@ export function useFidelityBondSweep({
             path: { walletname: walletFileName },
             body: { 'utxo-string': utxo.utxo, freeze: false },
           })
+          bondWasUnfrozen = true
         }
 
         const result = await directSend.mutateAsync({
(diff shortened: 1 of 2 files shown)


theborakompanioni:
Nice one! Thanks @CapThunder19 :pray:

src/components/earn/fidelity-bond/useFidelityBondSweep.ts
```diff
@@ -133,6 +135,20 @@ export function useFidelityBondSweep({
             console.debug('Error while unfreezing previously frozen UTXO in error handling.')
           }
         }
+
+        // Best-effort rollback — re-freeze the bond if we unfroze it before the error
+        if (bondWasUnfrozen) {
+          try {
+            await freezeUtxo.mutateAsync({
+              path: { walletname: walletFileName },
```
parrth20, line 146: could we avoid re-freezing after `directSend` succeeds? If the final refetch fails, this catch runs even though the bond is already spent perhaps guard it with a `sweepBroadcasted` flag and add a test for this case.

theborakompanioni: Maybe the refetch should be moved outside?

parrth20: Yes, this should be moved outside try/catch block!

theborakompanioni: This (still within a new try/catch)! 

On a side note: Jam could be much smarter in its reloading mechanism, but that is for some other time (not here).

CapThunder19:
pls review it i have put refetch outside and  inside the finally block.

theborakompanioni (approved):
utACK

PR #1400, opened 2026-08-02
fix: pin toLocaleString() calls to en-US locale

### LINKED ISSUE #1398: Unpinned toLocaleString() calls cause locale-dependent number formatting and non-deterministic test failures
**Expected behavior**

Numbers displayed in the app (block heights, fee rates, offer/fidelity-bond sizes, scheduled-transaction counts, etc.) should format consistently, and the test suite (`npm run check` / `npx vitest run`) should pass regardless of the machine's OS/browser locale setting.

**Actual behavior**

Roughly 15+ call sites across `src/` use `.toLocaleString()` without pinning an explicit locale, so output silently follows the host machine's default locale instead of a fixed one. This causes:

1. Inconsistent number formatting in production for users on non-US locales (e.g. block heights, fee rates, offer sizes render with different digit grouping/separators).
2. Non-deterministic test failures: on a machine set to `en-IN` locale, `npx vitest run --project unit` fails 17/795 tests, all traceable to this one root cause.

Examples:
- `Balance.test.tsx` (8 tests): expects `'12,345,600,000'`, gets `'12,34,56,00,000'` — via `formatSats` → `src/lib/utils.ts:208`
- `RescanChainPage.test.tsx`: expects `'From block #481,824'`, gets `'From block #4,81,824'` — raw `SEGWIT_ACTIVATION_BLOCK.toLocaleString()` at `src/components/settings/RescanChainPage.tsx:155` (also lines 115, 136)
- `utils.test.ts`, `JamDisplayContextProvider.test.tsx`: same root cause

Other unpinned production sites that will vary by locale: `src/components/layout/AppFooter.tsx:97` (block height), `src/components/orderbook/OrderbookChart.tsx:30-31` / `OrderbookContent.tsx:90` (fidelity-bond/offer sizes), `src/components/import/ImportStepConfirm.tsx:130,137` / `ImportDetailsForm.tsx:90-112` (blockheight/gaplimit on wallet import), `src/components/earn/EarnForm.tsx:155-163` (min offer size), `src/components/sweep/ScheduleEntryItem.tsx` / `SweepScheduleProgress.tsx`, `src/lib/feeEstimate.ts:67-77` (fee rate sat/vB).

This is distinct from the intentional locale-varying behavior of `formatBtc`/`formatSats` (fixed for a crash in #1289/PR #1290, which deliberately accepted locale-varying output for that one formatter — see the dedicated test at `utils.test.ts:580-593`). The ~15 other call sites were never brought in line with that decision.

1. Set your OS/browser locale to something other than en-US (e.g. en-IN or it-IT).
2. Clone the repo, `npm install`.
3. Run `npx vitest run --project unit`.
4. Observe 17 test failures related to number formatting (Balance.test.tsx, RescanChainPage.test.tsx, utils.test.ts, JamDisplayContextProvider.test.tsx) — expected en-US-formatted strings don't match actual locale-formatted output.
5. Alternatively, run the app itself under a non-US locale and observe block heights/fee rates/offer sizes rendered with unexpected digit grouping.

**Additional context**

`npm run build` and `npx eslint .` both pass cleanly — this is isolated to number formatting, not a broader regression. No privacy or fund-safety impact; this is purely a display/test-determinism issue.

**Proposed fix direction**

Either (a) pin all remaining `toLocaleString()`/`Intl.NumberFormat` calls to an explicit locale constant (e.g. `en-US`), or (b) if locale-following is intended everywhere, update the affected tests to mock/pin locale the way `withRuntimeLocale()` already does in `utils.test.ts:580`.

GuTS805:
Closes #1398

Pins the remaining unpinned `toLocaleString()` calls (block heights, fee rates, offer/fidelity-bond sizes, scheduled-transaction counts) to `en-US`, so number formatting is consistent regardless of the host machine's locale.

Also fixed `ImportDetailsForm.test.tsx`, which built its expected string using an unpinned `toLocaleString()` and would otherwise mismatch the now-pinned component output.

Tested: `npx vitest run --project unit` — 264/264 relevant tests pass. `tsc -b` build clean.

### DIFF
--- a/src/components/earn/EarnForm.tsx
+++ b/src/components/earn/EarnForm.tsx
@@ -152,15 +152,15 @@ export function EarnForm({
             .min(
               JAM.OFFER_MINSIZE_MIN,
               t('earn.feedback_invalid_min_amount_range', {
-                minAmountMin: JAM.OFFER_MINSIZE_MIN.toLocaleString(),
-                minAmountMax: offerMinsizeMax.toLocaleString(),
+                minAmountMin: JAM.OFFER_MINSIZE_MIN.toLocaleString('en-US'),
+                minAmountMax: offerMinsizeMax.toLocaleString('en-US'),
               }),
             )
             .max(
               offerMinsizeMax,
               t('earn.feedback_invalid_min_amount_range', {
-                minAmountMin: JAM.OFFER_MINSIZE_MIN.toLocaleString(),
-                minAmountMax: offerMinsizeMax.toLocaleString(),
+                minAmountMin: JAM.OFFER_MINSIZE_MIN.toLocaleString('en-US'),
+                minAmountMax: offerMinsizeMax.toLocaleString('en-US'),
               }),
             )
             .required(t('earn.feedback_invalid_min_amount')),
--- a/src/components/earn/EarnForm.tsx
+++ b/src/components/earn/EarnForm.tsx
@@ -152,15 +152,15 @@ export function EarnForm({
             .min(
               JAM.OFFER_MINSIZE_MIN,
               t('earn.feedback_invalid_min_amount_range', {
-                minAmountMin: JAM.OFFER_MINSIZE_MIN.toLocaleString('en-US'),
-                minAmountMax: offerMinsizeMax.toLocaleString('en-US'),
+                minAmountMin: JAM.OFFER_MINSIZE_MIN.toLocaleString(),
+                minAmountMax: offerMinsizeMax.toLocaleString(),
               }),
             )
             .max(
               offerMinsizeMax,
               t('earn.feedback_invalid_min_amount_range', {
-                minAmountMin: JAM.OFFER_MINSIZE_MIN.toLocaleString('en-US'),
-                minAmountMax: offerMinsizeMax.toLocaleString('en-US'),
+                minAmountMin: JAM.OFFER_MINSIZE_MIN.toLocaleString(),
+                minAmountMax: offerMinsizeMax.toLocaleString(),
               }),
             )
             .required(t('earn.feedback_invalid_min_amount')),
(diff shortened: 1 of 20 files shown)


theborakompanioni:
Hey @GuTS805! Thanks for the ticket and fix. Can you take a look at https://github.com/GuTS805/jam/blob/986aea014c5643e223584ba091ff1a71fa0582fb/src/lib/utils.test.ts#L40 ? Maybe this approach can be generalized. I don't think a fix should include hardcoding the locale in components, but rather only adapt the tests to setup everything correctly (i.e. setup various locales). What do you think?

kishore08-07 (commented):
Thanks for the PR! While this fixes the test failures, but hardcoding 'en-US' directly in the ui components has a few drawbacks:
- breaks i18n support,
- could lead to inconsistent ui: balances (btc/sats) will still format using the user's browser locale, creating a mix of formatting styles on the same screen.

GuTS805:
Thanks both for the feedback. good points on hardcoding the locale in components. I've reworked this:

- Reverted all component/source changes back to their original form (no hardcoded locale, so i18n / browser-locale-following behavior is preserved exactly as before).
- Generalized the `withRuntimeLocale` helper from `utils.test.ts` into a shared `src/test/withRuntimeLocale.ts` (also extended it to mock `Intl.NumberFormat` directly, since `getBtcParts` doesn't go through `Number.prototype.toLocaleString`).
- Used it to wrap the locale-dependent assertions across the previously-failing tests (`Balance.test.tsx`, `RescanChainPage.test.tsx`, `JamDisplayContextProvider.test.tsx`, `ImportDetailsForm.test.tsx`), plus a few pre-existing unwrapped cases in `utils.test.ts` itself that had the same issue.

Tests now pass deterministically regardless of the machine's locale, and the UI behavior is unchanged. Let me know if this matches what you had in mind.

theborakompanioni (approved):
utACK.

CI would need to run in a different locale than en-US to be sure this does not break again. Good follow-up?

GuTS805:
Good point. Happy to look into adding a non-en-US locale to the CI test matrix as a follow-up (either a separate PR or a tracking issue, whichever you'd prefer).

kishore08-07 (approved):

PR #1403, opened 2026-08-02
chore: print console warning

theborakompanioni:
Prints the following message in browser console:

image

Just a concept. Should this be done?


parrth20:
> Prints the following message in browser console:
> 
> image
> Just a concept. Should this be done?

Yep this is good idea we should implement this !!

parrth20 (approved):
ACK

PR #1404, opened 2026-08-02
refactor: use specific address type in ambiguity check

theborakompanioni:
Small follow-up to #1395.

Define ambiguity checks for specific address types.


parrth20 (approved):
LGTM

PR #1418, opened 2026-08-08
fix(send): correctly handle null amount on direct sweep sends

### LINKED ISSUE #1411: Direct sweep sends always fail with "Invalid amount given for sweep" (null/undefined mismatch in buildNonCollaborativeSendRequest)
## Description

When CoinJoin is disabled ("Send directly") and the user sweeps a jar, the transaction never gets built the app throws immediately instead of sending.

## Root cause

SendForm.schema.ts:115-120 transforms amount.amount to null when isSweep is true this is the real shape react-hook-form hands to onSubmit in production (confirmed by SendForm.schema.test.ts:251-259).

But collaborativeSend.ts:33-35 guards with:

if (data.amount.isSweep === true && data.amount.amount !== undefined) {
  throw new Error('Cannot trigger non-collaborative transaction: Invalid amount given for sweep.')
}
null !== undefined in JS, so this fires on every legitimate sweep, not just the malformed input it was meant to catch.

## Steps to reproduce

Send  toggle CoinJoin off ("Send directly")
Choose Sweep
Submit
Result: Error: Cannot trigger non-collaborative transaction: Invalid amount given for sweep.  no transaction sent.
Impact

Direct (non-CoinJoin) sweeps are completely non-functional for every user who tries them. The CoinJoin sweep path (buildCollaborativeSendRequest) is unaffected.

## Why CI didn't catch it

The existing test at collaborativeSend.test.ts:146-158 hand-constructs amount: undefined for the sweep case, instead of the null the schema actually produces, so the mismatch was never exercised.

## Suggested fix

In buildNonCollaborativeSendRequest, treat both null and undefined as "no amount set" (e.g. data.amount.amount != null).
Update the existing tests to use amount: null.
Add a regression test that feeds the schema's real sweep output into the builder.

CapThunder19:
Direct (non-CoinJoin) sweep sends throw an error and never complete. SendForm.schema.ts sets amount.amount to null when sweeping, but buildNonCollaborativeSendRequest only checked for undefined, so it rejected every real sweep submission.

Changed the check to catch both null and undefined. Added a test covering the actual null shape the form produces, alongside the existing undefined case.

Tested: npx vitest run src/components/send/collaborativeSend.test.ts (14/14 pass), eslint and prettier clean on changed files.

fixes #1411

### DIFF
--- a/src/components/send/collaborativeSend.test.ts
+++ b/src/components/send/collaborativeSend.test.ts
@@ -157,6 +157,22 @@ describe('buildNonCollaborativeSendRequest', () => {
     expect(request.amount_sats).toBe(0)
   })
 
+  it('maps sweep direct sends to amount_sats=0 when amount is null (real form-schema shape)', () => {
+    // SendForm.schema.ts transforms amount.amount to `null` (not `undefined`) once isSweep is true,
+    // which is what react-hook-form actually submits in production
+    const request = buildNonCollaborativeSendRequest({
+      ...baseValues(),
+      isCoinJoin: false,
+      amount: {
+        isSweep: true,
+        amount: null as unknown as undefined,
+        sweepAmount: 500_000,
+      },
+    })
+
+    expect(request.amount_sats).toBe(0)
+  })
+
   it('passes direct-send txfee values', () => {
     expect(
       buildNonCollaborativeSendRequest({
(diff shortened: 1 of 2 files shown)


theborakompanioni:
Hey @CapThunder19 I fail to reproduce the problem. Seeing no error when sending non-collaborative sweeps. Can you elaborate?

CapThunder19:
Sorry my mistake, 
 found the bug by calling the yup schema's validate() directly, which transforms amount.amount to null when sweeping. I fed that output straight into buildNonCollaborativeSendRequest and it threw, so I assumed that's what happens in production.
 
 I shouldve dug deeper, i did this time 
 
 Closing this PR and issue  nothing to fix here.

PR #1420, opened 2026-08-09
feat(earn): show own offer status in local orderbook

### LINKED ISSUE #1377: feat(earn): ability to see if own offer is found in local orderbook
As a user I want to immediately know whether my order is visible in my local orderbook (without opening the orderbook overlay).
Also display info about the bond (bond value; which one [if there a multiple]) in the orderbook (bonded offer) if a Fidelity Bond exists.

Image

parrth20:
Show the current maker offer’s local orderbook status directly on the Earn page.

- poll the existing orderbook query while the maker is running
- match the current offer using the nickname and offer ID
- show checking, visible, not found, and error states
- display the advertised Fidelity Bond value, amount, and locktime
- add tests for offer visibility and bond details

Closes #1377

### DIFF
--- a/src/components/earn/EarnPage.tsx
+++ b/src/components/earn/EarnPage.tsx
@@ -96,6 +96,7 @@ export const EarnPage = ({ walletFileName }: EarnPageProps) => {
       delayBefore: 1_000,
     }),
     enabled: makerRunning && !!currentOffer && !!jmSession?.nickname,
+    staleTime: Number.POSITIVE_INFINITY,
     refetchInterval: (query) =>
       query.state.error || !query.state.data?.offers.some((offer) => isCurrentOrderbookOffer(offer))
         ? JAM.WAIT_FOR_UPDATE_ORDERBOOK_POLLING_INTERVAL
--- a/src/components/earn/EarnPage.tsx
+++ b/src/components/earn/EarnPage.tsx
@@ -82,16 +82,19 @@ export const EarnPage = ({ walletFileName }: EarnPageProps) => {
 
   const isCurrentOfferAvailable = jmSession?.offer_list && jmSession.offer_list.length > 0
   const currentOffer = jmSession?.offer_list?.[0]
+  const isCurrentOrderbookOffer = (offer: OrderbookApi.OrderbookOffer) =>
+    offer.counterparty === jmSession?.nickname && String(offer.oid) === String(currentOffer?.oid)
 
   const orderbookQuery = useQuery({
     queryKey: ['orderbook'],
     queryFn: withQueryDelay(OrderbookApi.fetchOrderbook, { delayBefore: 1_000 }),
     enabled: makerRunning && !!currentOffer && !!jmSession?.nickname,
-    refetchInterval: JAM.WAIT_FOR_UPDATE_SESSION_POLLING_INTERVAL,
+    refetchInterval: (query) =>
+      query.state.error || !query.state.data?.offers.some((offer) => isCurrentOrderbookOffer(offer))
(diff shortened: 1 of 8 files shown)


src/components/earn/OfferCard.tsx
```diff
@@ -109,6 +149,27 @@ export function OfferCard({ className, value, nickname, children }: PropsWithChi
...
         )}
+        {bondValue > 0 && (
+          <div className="flex min-w-0 items-start gap-4 sm:col-span-2">
+            <ShieldCheckIcon className="mt-0.5 shrink-0" />
+            <div className="min-w-0 flex-1">
+              <Label className="font-semibold">{t('earn.current.text_fidelity_bond')}</Label>
+              <div className="text-muted-foreground flex flex-wrap gap-x-4 text-sm">
+                <span>
+                  {t('earn.current.text_bond_value')}: {Math.floor(bondValue).toLocaleString('en-US')}
+                </span>
```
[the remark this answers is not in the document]
parrth20: fixed in 252ee143

src/components/earn/EarnPage.tsx
```diff
@@ -80,6 +81,29 @@ export const EarnPage = ({ walletFileName }: EarnPageProps) => {
   const [showFeeConfigDialog, setShowFeeConfigDialog] = useState(false)
 
   const isCurrentOfferAvailable = jmSession?.offer_list && jmSession.offer_list.length > 0
+  const currentOffer = jmSession?.offer_list?.[0]
+
+  const orderbookQuery = useQuery({
+    queryKey: ['orderbook'],
+    queryFn: withQueryDelay(OrderbookApi.fetchOrderbook, { throttle: 210 }),
```
theborakompanioni:
```suggestion
    queryFn: withQueryDelay(OrderbookApi.fetchOrderbook, { 
      /* avoid flickering in visible/error/checking phases using isLoading/isFetching */
      delayBefore: 1_000
    }),
```

e.g. this happens on errors because `isLoading` is always true till first non-error response arrives.

src/components/earn/EarnPage.tsx
```diff
@@ -80,6 +81,29 @@ export const EarnPage = ({ walletFileName }: EarnPageProps) => {
...
+    enabled: makerRunning && !!currentOffer && !!jmSession?.nickname,
+    refetchInterval: JAM.WAIT_FOR_UPDATE_SESSION_POLLING_INTERVAL,
```
theborakompanioni: A couple of thoughts:

- The interval value should have its own config params(`WAIT_FOR_UPDATE_ORDERBOOK_POLLING_INTERVAL` or similar)
- The interval itself feels way too short. Maybe refetch more often while offer is not found in local orderbook by providing a reftech interval function?

Personally I think fetching every 15 seconds while 'missing' or 'error' is totally fine as default (the backend itself has a default of 30s anyway! A lower value  is okay -> feedback that something is happening in the background). When the order is 'visible', rechecking every ~2min is also fine.

src/components/earn/EarnPage.tsx
parrth20: fixed in 0353e745. I replaced the throttle with delayBefore: 1_000, which avoids flickering and keeps the checking state correct when the first request fails.

src/components/earn/EarnPage.tsx
parrth20: Done in 0e792df5. I added separate configurable intervals: 15s while the offer is missing/error and 2 minutes once it is visible. The query switches between them dynamically, and I added a test covering these states.

theborakompanioni (approved):

PR #1424, opened 2026-08-10
perf(context): optimize combinedUtxosHash from O(n²) to O(n)

### LINKED ISSUE #1421: Performance: combinedUtxosHash uses O(n²) array concatenation
# Performance: `combinedUtxosHash` uses an O(n²) array concatenation pattern

## Description

The `combinedUtxosHash` implementation in `src/context/JamWalletInfoContextProvider.tsx` currently combines arrays using a spread operation inside `reduce()`:

```ts
const combined = arrays.reduce(
  (acc, current) => [...acc, ...current],
  []
);
```

This creates a new array on every iteration and copies all previously accumulated elements into it again.

As the number of UTXO-related arrays/elements grows, the total amount of copying can become **O(n²)** instead of **O(n)**.

## Why this matters

`combinedUtxosHash` is used in the wallet information context and is recalculated when the relevant UTXO data changes.

The current implementation therefore performs unnecessary allocations and repeated copying. For larger UTXO sets or frequent updates, this can increase:

* CPU usage
* temporary memory allocations
* garbage-collection pressure
* time spent recalculating the hash

This is particularly relevant because the calculation occurs in the central wallet state provider.

## Evidence

The problematic pattern is:

```ts
[...acc, ...current]
```

inside a `reduce()`.

For example, conceptually:

```text
Iteration 1 → copy 1 element
Iteration 2 → copy previous elements + new elements
Iteration 3 → copy all previous elements + new elements
...
```

Therefore, previously accumulated data is repeatedly copied instead of being appended to a single buffer.

## Suggested improvement

The concatenation can be implemented using a pre-allocated `Uint8Array` and `set()` operations, or another linear-time buffer construction approach.

For example, conceptually:

```ts
const result = new Uint8Array(totalLength);

let offset = 0;

for (const current of arrays) {
  result.set(current, offset);
  offset += current.length;
}
```

This allows the data to be copied once into the final buffer, resulting in **O(n)** total copying.

The exact implementation should preserve the current ordering, byte representation, and hashing behavior.

## Expected result

The optimized implementation should produce exactly the same hash as the current implementation while avoiding repeated intermediate array allocations.

A regression test should verify that the resulting hash remains identical for:

* empty input
* a single UTXO/input
* multiple inputs
* larger inputs

A benchmark would also be useful to demonstrate the performance difference for larger inputs.

## Location

`src/context/JamWalletInfoContextProvider.tsx`

Approximately lines **132–136**.

## Duplicate check

I searched the repository's GitHub Issues and Pull Requests using multiple related terms, including:

* `combinedUtxosHash`
* UTXO performance
* UTXO hash
* spread reduce
* O(n²)
* performance

I did not find an existing issue or PR addressing this specific implementation.

## Proposed scope

This should be a small, self-contained performance improvement with no intended API changes.

The main requirements are:

1. Replace the repeated spread/`reduce()` concatenation.
2. Preserve the exact resulting byte sequence.
3. Preserve the existing SHA-256/hash output.
4. Add or update tests to verify identical behavior.
5. Optionally add a benchmark to demonstrate the improvement.

Amanyadav0556:
Closes #1421

This PR optimizes `combinedUtxosHash` in `src/context/JamWalletInfoContextProvider.tsx` by replacing the existing spread-inside-reduce concatenation with a pre-allocated `Uint8Array` and `Uint8Array.prototype.set()`.

The change eliminates repeated copying and intermediate array allocations while preserving the exact byte sequence and SHA-256 output.

## Problem

The previous implementation used:

```ts
const combinedUtxoIds = new Uint8Array(
  utxoIds.reduce((acc, current) => [...acc, ...current], [] as number[])
)
```

The `[...acc, ...current]` operation creates a new array on every iteration and re-copies all previously accumulated bytes.

For a total of `n` bytes, this can result in **O(n²) total copying** and creates unnecessary intermediate allocations, increasing memory usage and GC pressure.

## Solution

The new implementation:

1. Calculates the total required byte length.
2. Allocates one `Uint8Array` of the exact required size.
3. Copies each UTXO byte array once using `Uint8Array.set()`.

```ts
const totalLength = utxoIds.reduce(
  (sum, current) => sum + current.length,
  0
)

const combinedUtxoIds = new Uint8Array(totalLength)

let offset = 0

for (const current of utxoIds) {
  combinedUtxoIds.set(current, offset)
  offset += current.length
}
```

This reduces the concatenation from **O(n²) copying to O(n) copying** and removes the repeated intermediate array allocations.

## Behaviour / Compatibility

There is **no intended behaviour change**.

The optimized implementation preserves:

* UTXO ordering
* Byte ordering
* Byte values
* Combined byte sequence
* Final SHA-256 digest

Therefore, the resulting hash remains identical to the previous implementation.

## Tests

Added regression coverage for:

* Empty input
* Single UTXO
* Multiple UTXOs
* Varied byte values
* Larger input with 50 UTXOs

The tests verify that the optimized implementation produces the same SHA-256 digest as the previous implementation.

## Validation

* `git diff --check` ✅
* Hash equivalence verified by regression tests ✅
* Import order fixed according to `.prettierrc.json` ✅
* Only 2 relevant files changed ✅
* No new dependencies introduced ✅
* `@noble/hashes` already exists in production dependencies ✅

### Local CI Note

The repository requires Node.js `>=26.3.0`, while the local environment currently has Node.js `24.11.1`. Because of this version difference, the complete CI suite could not be run locally.

CI should run using the repository's supported Node.js version.

## Merge Request

This is a focused performance optimization with no intended functional or API changes. The implementation is covered by regression tests and preserves the existing hash output.

**Requesting review and merge once CI checks pass.**

### DIFF
--- a/src/context/JamWalletInfoContextProvider.test.tsx
+++ b/src/context/JamWalletInfoContextProvider.test.tsx
@@ -278,7 +278,6 @@ describe('<JamWalletInfoContextProvider />', () => {
  */
 const combinedUtxosHashReference = (utxoList: Utxo[]): string => {
   const utxoIds = utxoList.map((it) => hexToBytes(it.utxo.split(':', 1)[0]))
-  // eslint-disable-next-line unicorn/no-array-reduce
   const combinedUtxoIds = new Uint8Array(utxoIds.reduce((acc, current) => [...acc, ...current], [] as number[]))
   return bytesToHex(sha256(combinedUtxoIds))
 }
@@ -298,7 +297,7 @@ describe('combinedUtxosHash — optimized vs reference (regression)', () => {
     let captured = ''
     render(
       <JamWalletInfoContextProvider walletFileName="testing.jmdat">
-        <CaptureWalletInfo onContext={(ctx) => (captured = ctx.utxosHashHex)} />
+        <CaptureWalletInfo onContext={(context) => (captured = context.utxosHashHex)} />
       </JamWalletInfoContextProvider>,
     )
     return captured
--- a/src/context/JamWalletInfoContextProvider.test.tsx
+++ b/src/context/JamWalletInfoContextProvider.test.tsx
@@ -305,12 +305,29 @@ describe('combinedUtxosHash — optimized vs reference (regression)', () => {
 
   it('produces the same hash for an empty UTXO list', () => {
     const utxoList: Utxo[] = []
-    expect(captureHash(utxoList)).toBe(combinedUtxosHashReference(utxoList))
+    const actual = captureHash(utxoList)
+    expect(actual).toBe(combinedUtxosHashReference(utxoList))
+    expect(actual).toBe('e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855')
   })
 
   it('produces the same hash for a single UTXO', () => {
     const utxoList = [utxo({ utxo: `${txid('a')}:0` })]
-    expect(captureHash(utxoList)).toBe(combinedUtxosHashReference(utxoList))
+    const actual = captureHash(utxoList)
+    expect(actual).toBe(combinedUtxosHashReference(utxoList))
+    expect(actual).toBe('e0e77a507412b120f6ede61f62295b1a7b2ff19d3dcc8f7253e51663470c888e')
+  })
+
+  it('is okay for it to produce the same hash for different output in same txid', () => {
+    const utxoList0 = [utxo({ utxo: `${txid('a')}:0` })]
+    const actual0 = captureHash(utxoList0)
+    expect(actual0).toBe(combinedUtxosHashReference(utxoList0))
+    expect(actual0).toBe('e0e77a507412b120f6ede61f62295b1a7b2ff19d3dcc8f7253e51663470c888e')
+
+    const utxoList1 = [utxo({ utxo: `${txid('a')}:1` })]
+    const actual1 = captureHash(utxoList1)
+    expect(actual1).toBe(combinedUtxosHashReference(utxoList1))
+    expect(actual1).toBe('e0e77a507412b120f6ede61f62295b1a7b2ff19d3dcc8f7253e51663470c888e')
+    expect(actual1).toBe(actual0)
   })
 
   it('produces the same hash for multiple UTXOs', () => {
@@ -319,7 +336,10 @@ describe('combinedUtxosHash — optimized vs reference (regression)', () => {
       utxo({ utxo: `${txid('b')}:1` }),
       utxo({ utxo: `${txid('c')}:2` }),
     ]
-    expect(captureHash(utxoList)).toBe(combinedUtxosHashReference(utxoList))
+
+    const actual = captureHash(utxoList)
+    expect(actual).toBe(combinedUtxosHashReference(utxoList))
+    expect(actual).toBe('6bdaf8651d78f983ab7ce864414ace095b2942cbe5ff43deade48d51aa34b2cd')
   })
 
   it('produces the same hash for UTXOs whose txid bytes vary across the full byte range', () => {
--- a/src/context/JamWalletInfoContextProvider.test.tsx
+++ b/src/context/JamWalletInfoContextProvider.test.tsx
@@ -1,3 +1,5 @@
+import { sha256 } from '@noble/hashes/sha2.js'
+import { bytesToHex, hexToBytes } from '@noble/hashes/utils.js'
 import { render, screen, waitFor } from '@testing-library/react'
 import { Network } from 'bitcoin-address-validation'
 import { beforeEach, describe, expect, it, vi } from 'vitest'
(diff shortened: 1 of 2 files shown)


Amanyadav0556:
Hi, the CI checks are now passing successfully.

The lint issue has been fixed, and the performance optimization preserves the existing hash output.

Could you please review this PR and merge it if everything looks good?

Related issue: #1421

src/context/JamWalletInfoContextProvider.test.tsx
```diff
@@ -269,3 +271,72 @@ describe('<JamWalletInfoContextProvider />', () => {
...
+      <JamWalletInfoContextProvider walletFileName="testing.jmdat">
+        <CaptureWalletInfo onContext={(context) => (captured = context.utxosHashHex)} />
+      </JamWalletInfoContextProvider>,
+    )
+    return captured
+  }
+
+  it('produces the same hash for an empty UTXO list', () => {
+    const utxoList: Utxo[] = []
+    expect(captureHash(utxoList)).toBe(combinedUtxosHashReference(utxoList))
```
theborakompanioni:
```suggestion
    const actual = captureHash(utxoList)
    expect(actual).toBe(combinedUtxosHashReference(utxoList))
    expect(actual).toBe('e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855')
```

src/context/JamWalletInfoContextProvider.test.tsx
```diff
@@ -269,3 +271,72 @@ describe('<JamWalletInfoContextProvider />', () => {
...
+  })
+
+  it('produces the same hash for a single UTXO', () => {
+    const utxoList = [utxo({ utxo: `${txid('a')}:0` })]
+    expect(captureHash(utxoList)).toBe(combinedUtxosHashReference(utxoList))
```
theborakompanioni:
```suggestion
    const actual = captureHash(utxoList)
    expect(actual).toBe(combinedUtxosHashReference(utxoList))
    expect(actual).toBe('e0e77a507412b120f6ede61f62295b1a7b2ff19d3dcc8f7253e51663470c888e')
```

src/context/JamWalletInfoContextProvider.test.tsx
```diff
@@ -269,3 +271,72 @@ describe('<JamWalletInfoContextProvider />', () => {
...
+  })
+
+  it('produces the same hash for multiple UTXOs', () => {
+    const utxoList = [
+      utxo({ utxo: `${txid('a')}:0` }),
+      utxo({ utxo: `${txid('b')}:1` }),
+      utxo({ utxo: `${txid('c')}:2` }),
+    ]
+    expect(captureHash(utxoList)).toBe(combinedUtxosHashReference(utxoList))
```
theborakompanioni:
```suggestion
    const actual = captureHash(utxoList)
    expect(actual).toBe(combinedUtxosHashReference(utxoList))
    expect(actual).toBe('6bdaf8651d78f983ab7ce864414ace095b2942cbe5ff43deade48d51aa34b2cd')
```

src/context/JamWalletInfoContextProvider.tsx
```diff
@@ -131,7 +131,13 @@ interface JamWalletInfoContextProviderProps {
 
 const combinedUtxosHash = (utxos: Utxo[]) => {
```
theborakompanioni, line 132: Method can also move to utils, so the test setup can be simplified.
Can be done in a follow-up 🙏

src/context/JamWalletInfoContextProvider.tsx
```diff
@@ -131,7 +131,13 @@ interface JamWalletInfoContextProviderProps {
...
   const utxoIds = utxos.map((it) => hexToBytes(it.utxo.split(':', 1)[0]))
-  const combinedUtxoIds = new Uint8Array(utxoIds.reduce((acc, current) => [...acc, ...current], [] as number[]))
+  const totalLength = utxoIds.reduce((sum, current) => sum + current.length, 0)
+  const combinedUtxoIds = new Uint8Array(totalLength)
+  let offset = 0
+  for (const current of utxoIds) {
+    combinedUtxoIds.set(current, offset)
+    offset += current.length
+  }
```
theborakompanioni:
```suggestion
  const combinedUtxoIds = concatBytes(...utxoIds)
```

Can we use `concatBytes` from `@noble/hashes/utils.js`?
```
import { concatBytes } from '@noble/hashes/utils.js'
```

src/context/JamWalletInfoContextProvider.test.tsx
```diff
@@ -269,3 +271,72 @@ describe('<JamWalletInfoContextProvider />', () => {
...
+    return captured
+  }
+
+  it('produces the same hash for an empty UTXO list', () => {
+    const utxoList: Utxo[] = []
+    expect(captureHash(utxoList)).toBe(combinedUtxosHashReference(utxoList))
+  })
+
+  it('produces the same hash for a single UTXO', () => {
+    const utxoList = [utxo({ utxo: `${txid('a')}:0` })]
```
theborakompanioni, line 314: Maybe also add a test that the output index is currently not taken into account?
e.g. `${txid('a')}:0` and `${txid('a')}:1` will produce the same hash.

theborakompanioni:
Looks nice overall! Way better.
Maybe simplify with `concatBytes` and externalize function to utils?
Otherwise just NITs. :muscle:

src/context/JamWalletInfoContextProvider.tsx
theborakompanioni: Done!

src/context/JamWalletInfoContextProvider.test.tsx
theborakompanioni: Done.

src/context/JamWalletInfoContextProvider.test.tsx
theborakompanioni: Done.

src/context/JamWalletInfoContextProvider.test.tsx
theborakompanioni: Done.

src/context/JamWalletInfoContextProvider.test.tsx
theborakompanioni: Done.

theborakompanioni (approved):

Amanyadav0556:
Hi Thebora,
Thank you so much for reviewing and merging my PR #1424.
I really appreciate the detailed feedback and guidance you provided during the review. It helped me improve the implementation and, more importantly, understand the project's standards and expectations better.
I’ve genuinely enjoyed working on this contribution and learning from the review process. I’m looking forward to contributing more to the project and continuing to learn from the community.
Thanks again for your time, feedback, and support. I truly appreciate the opportunity to contribute!
Best regards,
Aman Yadav

________________________________

PR #1426, opened 2026-08-11
fix(earn): export complete report as CSV

### LINKED ISSUE #1425: Earn report CSV export only contains the current table page
## Expected behavior

Hitting "Download" on the Earn Report should give me my whole report.

## Actual behavior

It only exports the page I'm currently looking at 25 rows by default. The rest are
just missing, with no warning and a filename that looks like a full export. And since
the table is sorted newest first, you keep the 25 newest and lose all the older ones,
which is the opposite of what you want if you're exporting for your records.

1. Get more than 25 entries in the Earn Report (developer mode's "Generate demo report"
   works).
2. Open the report — the summary and chart correctly show the full count.
3. Leave the page size at 25 and click "Download".
4. Count the rows in the CSV. You get 25.

Picking "Show all" before downloading works around it, which is probably why nobody
noticed.

## Specifications

- Version: 2.0.0-beta.2 (`devel` @ c832b80)
- Platform / Browser: any

## Suggested fix

Export from `table.getSortedRowModel().rows` instead, which keeps the filter and sort
but ignores pagination. Leave `visibleRows` for the table body, and add a test with more
than 25 entries that actually reads the blob and counts rows.

Guflly:
Closes #1425

The Earn Report CSV now uses the filtered and sorted rows before pagination, so downloading does not discard entries outside the current page. The table remains paginated.

A regression test covers a 30-entry report and verifies that the 25-row page exports all 30 rows in the current sort order.

Tests:
- `vitest run --project unit src/components/earn/report/EarnReportContent.test.tsx`
- `tsc -b --pretty false`
- focused ESLint and Prettier checks

### DIFF
--- a/src/components/earn/report/EarnReportContent.test.tsx
+++ b/src/components/earn/report/EarnReportContent.test.tsx
@@ -209,6 +209,35 @@ describe('EarnReportContent', () => {
     createElement.mockRestore()
   })
 
+  it('exports all filtered rows regardless of pagination', async () => {
+    mocks.entries = Array.from({ length: 30 }, (_, index) =>
+      entry({
+        notes: `maker note ${index + 1}`,
+        timestamp: new Date(Date.UTC(2026, 0, index + 1, 12)),
+      }),
+    )
+
+    render(<EarnReportContent enabled />)
+
+    expect(screen.getByText('pagination:30')).toBeInTheDocument()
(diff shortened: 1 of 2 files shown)


parrth20:
Hey @Guflly thanks for contributing !!

parrth20 (approved):
Looks good this keeps filtering and sorting while exporting all rows regardless of pagination. The regression test covers the issue well. Approved!!!!!

PR #1430, opened 2026-08-11
fix(a11y): make settings rows keyboard-accessible

### LINKED ISSUE #1429: Seed phrase and xpubs can't be reached without a mouse
## Expected behavior

I should be able to Tab to a Settings row and open it with Enter or Space.

## Actual behavior

The Settings rows aren't real buttons — they're divs with an onClick and a
`cursor-pointer` style, so they only respond to a mouse. Tab skips straight past
them and Enter does nothing.

For most rows there's another way in, so it's just annoying — rescan and logs are in
the sidebar, lock wallet is a proper button in the navbar. But two rows have no
other route at all:

- Show seed phrase
- Show extended public keys

Those dialogs open from `SettingsPage.tsx:146` and `:153` and nowhere else. So if you
use a keyboard, you see your seed once while creating the wallet and can never get
back to it. If you imported the wallet, you never see it at all.

1. Unlock a wallet and open Settings.
2. Press Tab over and over — focus only ever hits the toggle switches, never the
   seed or xpubs rows.
3. There's no way to open either one without a mouse.

## Suggested fix

Make the row a real `<button type="button">` when it has an action, using the same
style reset as `Balance.tsx` after #1412. That covers every row in one go.

One thing to sort out: `SettingSwitch` puts a `<Switch>` inside the row, and a button
inside a button isn't valid HTML — so that case probably needs to stay a plain
wrapper, or become a `<label>`.

`SettingsItem.tsx` is also the only file in `src/components/settings/` without a test.

CapThunder19:
Settings rows only worked with a mouse.

`SettingItem` rendered a `div` with an `onClick` and `cursor-pointer` styling, so the rows
looked and felt like buttons  but had no `role`, `tabIndex` or key handler. Tab skipped
straight past them and Enter/Space did nothing. A plain `div` can't take programmatic focus
either, so the rows couldn't be focused at all, not just skipped in the tab order.

Most rows had another way in, so this was mostly an annoyance  rescan and logs are in the
sidebar, lock wallet is a real button in the navbar. But **show seed phrase** and **show
xpubs** open from the settings page and nowhere else, so keyboard users couldn't reach their
seed backup at all. You'd see it once during wallet creation and never again, or never at
all if you imported the wallet.

While fixing it I noticed the three `displayToggle={false}` switches privacy mode,
currency unit and theme render no toggle, so they had no focusable child either and were
equally unreachable. Those are covered now too.

### Approach

`SettingItem` renders a real `button` when the row is itself the control. That way focus,
Enter/Space and the role come from the browser instead of being hand-rolled, and `disabled`
uses the native attribute rather than styling alone.

Rows whose children already provide a focusable control keep a plain wrapper, so a button is
never nested inside a button. `SettingSwitch` derives that from its existing `displayToggle`
prop, so there's no new decision at any call site.

Same pattern as the balance toggle fix in #1412.

### Tests

`SettingsItem.tsx` was the only file in `src/components/settings/` without a test, so this
adds one covering tab order, Enter, Space, click, the disabled state and both switch
variants. `SettingsPage.test.tsx` passes unchanged  its `.closest('div')` helper still
works because the click bubbles to the new button.

Full unit suite green (881 tests), eslint/prettier/tsc clean.

fixes #1429

### DIFF
--- a/src/components/settings/SettingsItem.tsx
+++ b/src/components/settings/SettingsItem.tsx
@@ -28,10 +28,16 @@ export const SettingItem = ({
   hasInteractiveChild = false,
   children,
 }: SettingsItemProps) => {
-  const className = cx('flex w-full min-w-0 items-center justify-between gap-2 py-2 text-left', {
-    'hover:bg-muted/50 -mx-2 cursor-pointer rounded-md px-2': !disabled,
+  const rowClassName = cx('flex min-w-0 items-center justify-between gap-2 py-2', {
+    'hover:bg-muted/50 cursor-pointer rounded-md px-2': !disabled,
     'cursor-not-allowed opacity-60': disabled,
   })
+  // The row bleeds into the card padding so the hover background spans the full width.
+  // A block `div` with `width: auto` gets this from the negative margin alone, but a
+  // `button` does not stretch to its parent, and once `w-full` pins the width the
+  // negative margin no longer widens it. Keeping the bleed on a wrapper makes `w-full`
+  // resolve against the widened box, so both variants end up the same size.
+  const bleedClassName = cx({ '-mx-2': !disabled })
 
   const content = (
     <>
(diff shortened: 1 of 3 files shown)


src/components/settings/SettingsItem.tsx
```diff
@@ -19,16 +25,16 @@ export const SettingItem = ({
   title,
   action,
   disabled = false,
+  hasInteractiveChild = false,
   children,
 }: SettingsItemProps) => {
+  const className = cx('flex w-full min-w-0 items-center justify-between gap-2 py-2 text-left', {
```
theborakompanioni: This somehow changes the width of the item.
See the hover effect before and after.

### 📸  Before/After

Image
Image

CapThunder19:
pls review it

theborakompanioni (approved):
lgtm 🚀

PR #1439, opened 2026-08-15
feat(slider): add opt-in companion numeric input

### LINKED ISSUE #1371: ui: sliders should have additional input method
All sliders should also have a companion input element, so users can change values manually

Image

e.g.

Image

chiliec:
## What

Closes #1371.

Sliders can now render a companion numeric input, so users can type an exact value instead of only dragging — matching the screenshots in the issue.

This is done via a new **opt-in** `withInput` prop on the base `Slider` component (`src/components/ui/slider.tsx`), so it's fully backward compatible: any existing `<Slider>` with no `withInput` renders exactly as before.

## Behaviour

When `withInput` is set on a **single-thumb** slider, a small number `<Input>` is rendered beside it that:

- reflects the current value,
- lets the user type a value,
- clamps to `[min, max]`,
- snaps to `step` (when provided),
- calls the same `onValueChange([n])` the slider uses, so there is a single source of truth,
- inherits `disabled`,
- ignores empty / non-numeric input,
- carries the slider's `aria-label`.

Multi-thumb range sliders are guarded (`_values.length === 1`) — a single companion input can't represent two thumbs, so the input is simply not rendered there.

Two small supporting changes in the base component:
- `step` is now forwarded to the Radix `Root` (it was previously dropped).
- a `SliderProps` type is exported.

## Where it's adopted

- `SendForm` — the number-of-collaborators slider.
- `SweepForm` — the transactions-per-jar and rounding-chance-percent sliders.
- **Left unchanged:** the SweepForm min/max-collaborators slider — it's a multi-thumb range slider, so a single companion input doesn't apply.

(If you'd prefer a different set of adoption sites, or the input on every single-thumb slider automatically, happy to adjust.)

## Tests / validation

- New `src/components/ui/slider.test.tsx` — 9 Vitest cases: no input by default, reflects value, typed→`onValueChange`, clamp above/below, step snap, ignore empty, disabled, and multi-thumb renders no input.
- New `WithInput` Storybook story.

Ran locally (Node 26.3.0 per `engine-strict`):
- `npx tsc --noEmit -p tsconfig.app.json` → passes
- `vitest run src/components/ui/slider.test.tsx` → **9 passed**
- `vitest run src/components/send src/components/sweep` → **160 passed (23 files)**, no regressions
- `eslint` (changed files) → 0 problems
- `npm run build` → succeeds

## Note

First-time contributor here — happy to adjust naming, scope, or styling to fit the project's conventions.

### DIFF
--- a/src/components/ui/jam/SliderWithInput.tsx
+++ b/src/components/ui/jam/SliderWithInput.tsx
@@ -1,0 +1,83 @@
+import * as React from 'react'
+import * as SliderPrimitive from '@radix-ui/react-slider'
+import { cn } from '@/lib/utils'
+import { Input } from '../input'
+import { Slider } from '../slider'
+
+type SliderWithInputProps = React.ComponentProps<typeof SliderPrimitive.Root> & {
+  /** Optional class name applied to the companion input. */
+  inputClassName?: string
+}
+
+/**
+ * A single-thumb {@link Slider} paired with a companion numeric input, so users
+ * can type an exact value instead of dragging. Keeps the base `ui/slider`
+ * component a clean Radix wrapper; the typing behaviour lives here.
+ *
+ * Only meaningful for single-thumb sliders (a single `value`/`defaultValue`).
+ */
+function SliderWithInput({
+  className,
+  inputClassName,
+  defaultValue,
+  value,
+  min = 0,
+  max = 100,
+  step,
+  ...props
+}: SliderWithInputProps) {
+  const _values = React.useMemo(
+    () => (Array.isArray(value) ? value : Array.isArray(defaultValue) ? defaultValue : [min, max]),
+    [value, defaultValue, min, max],
+  )
+
+  const currentValue = _values[0]
+  const disabled = props.disabled
+
+  const clamp = (raw: number) => {
+    let next = Math.min(max, Math.max(min, raw))
+    if (typeof step === 'number' && step > 0) {
+      // Snap to the nearest step relative to `min`, then re-clamp to `max`.
+      next = min + Math.round((next - min) / step) * step
+      next = Math.min(max, Math.max(min, next))
+    }
+    return next
+  }
+
+  const commitValue = (rawText: string) => {
+    if (rawText.trim() === '') return
+    const parsed = Number(rawText)
+    if (Number.isNaN(parsed)) return
+    props.onValueChange?.([clamp(parsed)])
+  }
+
+  return (
+    <div data-slot="slider-with-input" className="flex w-full items-center gap-3">
+      <Slider
+        className={className}
+        defaultValue={defaultValue}
+        value={value}
+        min={min}
+        max={max}
+        step={step}
+        {...props}
+      />
+      <Input
+        data-slot="slider-input"
+        type="number"
+        inputMode="numeric"
+        min={min}
+        max={max}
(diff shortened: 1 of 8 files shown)


src/components/ui/slider.tsx
```diff
@@ -1,27 +1,45 @@
 import * as React from 'react'
 import * as SliderPrimitive from '@radix-ui/react-slider'
 import { cn } from '@/lib/utils'
+import { Input } from './input'
+
+type SliderProps = React.ComponentProps<typeof SliderPrimitive.Root> & {
```
theborakompanioni: Hey @chiliec! Awesome. :muscle: 
Would it be possible to wrap the behaviour in a separate component, e.g. `SliderWithInput` in the `ui/jam` dir, and keep `ui/slider.tsx` untouched?
This way radix slider component can be updated more easily.

Also, is it possible to add it to multi-value slider?

Image

Otherwise: LGTM!

theborakompanioni (changes requested):

src/components/ui/slider.tsx
theborakompanioni: Any thoughts @chiliec?

chiliec:
Thanks @theborakompanioni — done! I've extracted the behaviour into a dedicated `SliderWithInput` component under `ui/jam/` and reverted `ui/slider.tsx` (and its story) back to the untouched base Radix wrapper.

- `ui/slider.tsx` is now identical to `master` again — no `withInput` prop on the base component.
- New `ui/jam/SliderWithInput.tsx` wraps `Slider` and adds the companion numeric input (clamp / step-snap / disabled handling), with its tests moved alongside it and a new `stories/jam/SliderWithInput.stories.tsx`.
- `SendForm` / `SweepForm` use `SliderWithInput` where the numeric input is wanted; the multi-thumb range slider in `SweepForm` keeps using the base `Slider`.

Verified locally: `tsc --noEmit` clean, ESLint clean, and the 8 unit tests pass.

src/components/ui/jam/SliderWithInput.tsx
```diff
@@ -0,0 +1,83 @@
...
+        inputMode="numeric"
+        min={min}
+        max={max}
+        step={step}
+        value={currentValue ?? ''}
+        disabled={disabled}
+        aria-label={props['aria-label']}
+        onChange={(event) => commitValue(event.target.value)}
+        className={cn('h-8 w-20 shrink-0 text-center', inputClassName)}
+      />
```
theborakompanioni:
```suggestion
      <Input
        data-slot="slider-input"
        type="number"
        inputMode="numeric"
        min={min}
        max={max}
        step={step}
        value={currentValue ?? ''}
        disabled={disabled}
        aria-label={props['aria-label']}
        onChange={(event) => commitValue(event.target.value)}
        className={cn('h-8 w-20 shrink-0 text-center', inputClassName)}
      />
      <Slider
        className={className}
        defaultValue={defaultValue}
        value={value}
        min={min}
        max={max}
        step={step}
        {...props}
      />
```

theborakompanioni (approved):
Looks nice! 

Small change added: rendering the input first, slider second.

Thank you so much :muscle: 

Maybe multi-value slider can be done in a follow-up, but only if you want to :pray:

PR #1443, opened 2026-08-16
fix(fb): don't treat a failed post-broadcast refetch as a bond creati…

### LINKED ISSUE #1442: Create Fidelity Bond: a successful bond transaction is reported as a failure when the wallet refetch afterwards fails
While reading through the fidelity bond flows I noticed the create wizard can tell you bond creation failed *after* the transaction has already been broadcast.

`handleCreateFidelityBond` in `useCreateFidelityBondWizard.ts` wraps everything in one `try`, including the final wallet refresh:

```js
const result = await directSend.mutateAsync({ ...amount_sats: 0 /* sweep */ })

// ...unfreeze loop and freeze-new-bond call, both individually try/caught...

setTxResult(result)
setStep('success')
toast.success(...)

await walletInfo.refetch()      // <-- not guarded, still inside try
} catch {
  ...
  setStep('freeze_utxos')       // <-- undoes 'success' after a real broadcast
}
```

Every other post-broadcast step in there is individually try/caught. `refetch()` is the only one that isn't and it does throw: `walletInfo.refetch` is a mutation's `mutateAsync` whose `mutationFn` calls `.refetch({ throwOnError: true })` on both wallet queries with `retry: false`, so any transient backend hiccup rejects.

**What the user gets:** a "Success" toast, then a silent bounce back a step. No error message at all  `setError` is only wired to the freeze/unfreeze/directSend mutations, not to this. The txid is still sitting in `txResult`, but the success screen only renders at `step === 'success'`, so they never see it. Balances stay stale because the failed refetch isn't retried. Their coins are locked and they have no way to know it worked.

The window isn't a tight race either there's a deliberate ~1s `delayedPromise` plus one unfreeze round-trip per frozen UTXO before the refetch runs.

**This is already fixed in the sibling path.** `useFidelityBondSweep` (renew / move-to-jar) does the right thing since #1399:

```js
} finally {
  // refetch only if the sweep broadcast, and don't let a refetch error
  // get treated as a sweep failure
  if (sweepBroadcasted) {
    try { await walletInfoRefetch() } catch (error) { console.warn(...) }
  }
}
```

...plus a `sweepBroadcasted` flag so nothing post-broadcast can reach the rollback. The create wizard has neither. There's even a `// TODO: refactor: use the sweep functionality from useFidelityBondSweep here` directly above the affected block.

**Reproduced** by driving the real hook with only the network boundary mocked (same approach as `useFidelityBondSweep.test.ts`) `directSend` resolves, `refetch` rejects:

```
step        : freeze_utxos
txResult    : {"txinfo":{"txid":"created-tx"}}
error shown : undefined
toast.success called: 1
```

The identical test with a resolving refetch ends on `success`, so the refetch is the only variable.

**Suggested fix:** set a `broadcasted` flag once `directSend` resolves, gate the catch rollback on `!broadcasted`, and move `refetch()` into a `finally` with its own try/catch mirroring `useFidelityBondSweep`.

Related: #1397 / #1399 (same bug class in renew/move-to-jar), #1435 (same class in the import flow).

CapThunder19:
Fixes #1442.

`handleCreateFidelityBond` awaited `walletInfo.refetch()` inside the same `try` that guards the sweep. That refetch is a mutation's `mutateAsync`, and its `mutationFn` calls `.refetch({ throwOnError: true })` on both wallet queries with `retry: false`, so any transient backend failure rejects.

When that happened after the bond transaction had already broadcast, the catch ran and reset the wizard back to `freeze_utxos`. The user got a success toast, then got bounced back a step with no error message at all (nothing calls `setError` on that path), and the txid was left stranded in state because the success screen only renders at `step === 'success'`. Balances stayed stale too, since the failed refetch is never retried.

Every other post-broadcast step in that function was already individually try/caught. The refetch was the only one that wasn't.

## What changed

The same guard `useFidelityBondSweep` got in #1399, adapted for the create flow:

* track `broadcasted`, set right after `directSend` resolves
* bail out of the catch when `broadcasted` is true, so nothing after the broadcast can rewind the wizard
* move `walletInfo.refetch()` into a `finally` with its own try/catch

The `bondWasUnfrozen` half of #1399 has no equivalent here. The create flow never unfreezes a bond, it freezes the sibling utxos and sweeps a normal utxo into a fresh timelock address.

## Tests

Added a `bond creation failures` block to the existing suite, which had no failure path coverage at all (no `mockRejected` anywhere in the file):

* `stays on the success step when the wallet refetch after the broadcast fails`, which fails on devel with `expected 'freeze_utxos' to be 'success'` and passes here
* `rolls back to the freeze step when the send itself fails`, confirming genuine failures still roll back and still skip the refetch

The 25 existing tests in that file pass both with and without the fix, so the new one is the only test that moves.

Locally green: 27 tests in that file, 164 across `src/components/earn`, plus eslint, prettier and `tsc -b`.

### DIFF
--- a/src/components/earn/CreateFidelityBondDialog/useCreateFidelityBondWizard.test.ts
+++ b/src/components/earn/CreateFidelityBondDialog/useCreateFidelityBondWizard.test.ts
@@ -101,6 +101,18 @@ const minLockdateOption = (isDeveloperMode = false) => {
   return lockdate
 }
 
+const submitFromReview = async (result: { current: ReturnType<typeof useCreateFidelityBondWizard> }) => {
+  act(() => result.current.setSelectedLockdate(minLockdateOption()))
+  act(() => result.current.setSelectedJarIndex(0))
+  act(() => result.current.toggleUtxoSelection(result.current.availableUtxos[0]))
+  act(() => result.current.setStep('freeze_utxos'))
+  await act(async () => result.current.handleNext())
+  expect(result.current.step).toBe('review')
+
+  act(() => result.current.setConfirmationChecked(true))
+  await act(async () => result.current.handleNext())
+}
+
 describe('useCreateFidelityBondWizard', () => {
   beforeEach(() => {
     vi.clearAllMocks()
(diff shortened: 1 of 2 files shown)


theborakompanioni (approved):

theborakompanioni:
Thank you @CapThunder19 :pray:

PR #1445, opened 2026-08-17
feat(earn): Fixed Fee Quantization Bands

### LINKED ISSUE #1427: feat: Fee Quantization Bands
Add quick selection for common maker fees settings on the Earn page.
Explain the reason and encourage users to use these (e.g. small info box when custom value is provided).

```
QUANT_REL = [2e-5, 5e-5, 1e-4, 2e-4, 5e-4, 1e-3, 2e-3, 5e-3, 1e-2, 2e-2, 5e-2, 1e-1]
                  # in percent: 0.002, 0.005, 0.01, 0.02, 0.05, 0.1, 0.2, 0.5, 1, 2, 5, 10
QUANT_ABS = [100, 200, 500, 1000, 2000, 5000, 10000]   # sats per slot
```

See "Fee Quantization Bands" on https://joinmarket-ng.sgn.space/
Full discussion: https://github.com/joinmarket-ng/joinmarket-ng/issues/508

theborakompanioni:
Closes #1427 

- disables fees for unbonded makers
- allows only predefined fee bands by default

image

image

image

- custom values can be enabled in an "advanced" section on the settings page
image

Various other small improvements:
- improve color of chunked addresses
- improve color of non-significant balance digits
- `PageTitle`  supports child components
- balanced text for page description 
- subtitles for `SettingsItem` component
- earn: move min amount input in options sections (similar accordion as in other forms)


src/components/orderbook/OrderbookChart.tsx
```diff
@@ -14,24 +15,19 @@ interface FeeBucket {
...
-  relative: [0.00002, 0.00005, 0.0001, 0.0002, 0.0005, 0.001, 0.002, 0.005, 0.01, 0.02, 0.05, 0.1],
-  absolute: [0, 100, 200, 500, 1_000, 2_000, 5_000, 10_000],
-}
-
 const bucketByQuantizationBand = (entries: OrderTableEntry[], mode: FeeMode) => {
   const offersByMaker = new Map<string, OrderTableEntry>()
 
   for (const entry of entries) {
-    if ((mode === 'absolute' ? !entry.type.isAbsolute : !entry.type.isRelative) || entry.bondValue.value <= 0) continue
+    if (mode === 'absolute' ? !entry.type.isAbsolute : !entry.type.isRelative) continue
```
parrth20: this removes the existing bondValue <= 0 filter, so unbonded makers are counted as bonded and included in the quantization buckets. pls preserve the bond-value filter while reusing OFFER_FEE_BANDS.

src/components/earn/EarnForm.tsx
```diff
@@ -31,32 +32,40 @@ const FieldPrefixSatSymbol = (
...
 
   return yup
     .object({
-      offerType: yup.string<OfferType>().default(FORM_INPUT_DEFAULT_VALUES.offerType).required(),
+      offerType: yup
+        .string<OfferType | '__free'>()
+        .default(FORM_INPUT_DEFAULT_VALUES.offerType)
+        .required()
+        .test('valid-offer-type-test', t('earn.feedback_invalid_offer_type_bondless_maker'), (value) => {
+          if (fidelityBonds.length === 0) {
```
parrth20, line 64: `fbOutputs` also contains expired bonds, so checking only `length>0` allows paid offers with no active bond. Could we require at least one currently locked bond and add an expired-only regression test?

src/components/ui/jam/Balance.tsx
```diff
@@ -96,10 +91,15 @@ const BitcoinBalance = ({
...
-          [styles.bitcoinAmountColor]: highlightSignificantDigits,
-        })}
+        className={cn(
+          'slashed-zero tabular-nums select-all',
+          styles.bitcoinAmount,
+          {
+            [styles.bitcoinAmountSpacing]: fractionalPartSpacing,
+            [styles.bitcoinAmountColor]: highlightSignificantDigits,
+          },
+          className,
```
parrth20: `className` previously styled the complete balance wrapper. Applying it only to the numeric span makes callers such as `FidelityBondAmount` render a large amount with a normal-sized symbol. Could we preserve the wrapper-level behavior?

src/components/settings/SettingsPage.tsx
```diff
@@ -39,23 +40,55 @@ import { useFeeConfigValidation } from '@/hooks/useFeeConfigValidation'
...
+  const { t } = useTranslation()
+
+  const [tab, setTab] = useState<SettingsTab>(initialTab)
+
+  return (
+    <div className="mx-auto max-w-4xl space-y-3 p-4">
+      <PageTitle title={t('navbar.menu_mobile_settings')}>
+        <Tabs value={tab} onValueChange={(value) => setTab(value as SettingsTab)}>
+          <TabsList>
+            <TabsTrigger value="basic">Basic</TabsTrigger>
```
parrth20: The new Basic/Advanced tabs and expert-setting labels are hardcoded English. Could we move this new user-facing copy into the i18n resources???

src/components/earn/EarnForm.tsx
theborakompanioni: Expired bonds are still valuable. I think this is okay for now, what do you think?
When the FB really drops to zero value, people will notice by nobody picking up their offer.
Not that his can happen to any maker running longer than the expiry date of their bond.

src/components/orderbook/OrderbookChart.tsx
theborakompanioni: Hmm.. I removed it since this is just the chart displaying information from the table, right? Shouldn't it include all offers regardless of the bond? (genuine question - happy to remove it if you think this is more suitable)

src/components/ui/jam/Balance.tsx
theborakompanioni: Good point, thanks. Fixed.

src/components/settings/SettingsPage.tsx
theborakompanioni: Done. 🙏

src/components/earn/EarnForm.test.tsx
```diff
@@ -26,77 +32,153 @@ describe('EarnForm', () => {
...
+      expect.objectContaining({
+        offerAbsoluteFee: OFFER_FEE_ABS_DEFAULT,
+        offerMinAmount: OFFER_MINSIZE_DEFAULT,
+        offerType: '__free',
+      }),
+      expect.anything(),
+    )
+  })
+
+  it('switches to absolute fees and submits the edited values', async () => {
```
parrth20, line 79: Could we add a focused test confirming that a paid offer with fidelityBonds={[]} is rejected? The current tests cover free-without-bond and paid-with-bond, but not this new rejection path.

src/components/earn/EarnForm.tsx
parrth20: You’re right expiry does not immediately reduce a bond’s value to zero it decays afterward. I’m okay with keeping this as-is for now.

src/components/orderbook/OrderbookChart.tsx
parrth20: I’d still keep the bond value filter...the table can show all offers, but this chart and the linked analysis are specifically about bonded makers and explicitly ignore bondless offers, could you restore the filter and keep zero-bond entries excluded in the test?

theborakompanioni: Agreed. :muscle:

src/components/earn/EarnForm.test.tsx
theborakompanioni: Absolutely! Done.

parrth20 (approved):
LGTM !!!!

PR #1448, opened 2026-08-20
fix: skip initial UTXO refetch

### LINKED ISSUE #1447: Bug: AFTER_UTXO_CHANGE triggers redundant wallet refetches after UTXO updates
# Bug: `AFTER_UTXO_CHANGE` causes redundant wallet refetches

## Expected behavior

A single wallet synchronization event should refresh the wallet information once.

When that refresh already updates the UTXO state, the resulting `utxosHashHex` change should **not** immediately trigger the same `refetchWalletBalance()` operation again.

## Actual behavior

`refetchWalletBalance()` already fetches both the UTXOs and wallet display information. The resulting UTXO update changes `utxosHashHex`, which is also a dependency of the `AFTER_UTXO_CHANGE` effect.

This causes the effect to schedule another complete wallet refresh after the configured **210ms delay**.

The resulting flow is:

```text
AFTER_TX_MESSAGE
      ↓
refetchWalletBalance()
      ↓
/utxos + /displaywallet
      ↓
utxosHashHex changes
      ↓
AFTER_UTXO_CHANGE
      ↓
210ms delay
      ↓
refetchWalletBalance() again
      ↓
/utxos + /displaywallet again
```

## 1. The first refresh already fetches both resources

**File:** `src/context/JamWalletInfoContextProvider.tsx`

```ts
utxosQueryResult
  .refetch({ throwOnError: true })
  .then((utxos) =>
    displayWalletQueryResult
      .refetch({ throwOnError: true })
      .then(() => utxos)
  )
```

So one `refetchWalletBalance()` already performs both:

```text
/utxos
/displaywallet
```

## 2. The resulting UTXO change triggers another refresh

**File:** `src/context/JamWalletInfoContextProvider.tsx`

```ts
const utxosHashHex = useMemo(
  () => bytesToHex(combinedUtxosHash(utxos)),
  [utxos]
)
```

**File:** `src/App.tsx`

```ts
useEffect(
  function refetchWalletInfoAfterUtxoChange() {
    const delayBefore =
      RELOAD_WALLET_INFO_DELAY.AFTER_UTXO_CHANGE

    // ...
    refetchWalletBalance({ delayBefore, signal })
  },
  [refetchWalletBalance, utxosHashHex]
)
```

The important part is that the first wallet refresh changes the value that the second effect is watching:

```text
refetchWalletBalance()
        ↓
utxos updated
        ↓
utxosHashHex changes
        ↓
useEffect runs
        ↓
refetchWalletBalance() again
```

This means the refresh itself can indirectly trigger another refresh.

## 3. Reproduction

1. Open the application with the Network tab enabled.
2. Unlock a wallet or trigger an event such as `AFTER_TX_MESSAGE`.
3. Observe the initial `/utxos` and `/displaywallet` requests.
4. Wait for the UTXO response to update `utxosHashHex`.
5. Observe another pair of requests approximately **210ms later**.

### Observed

```text
First refresh
  /utxos
  /displaywallet

~210ms later

Second refresh
  /utxos
  /displaywallet
```

As a control, disabling only the `AFTER_UTXO_CHANGE` effect results in one request to each endpoint for the same synchronization event.

## Evidence

| `AFTER_UTXO_CHANGE` | Enabled | Disabled |
| ------------------- | ------: | -------: |
| `/utxos`            |       2 |        1 |
| `/displaywallet`    |       2 |        1 |

This behavior was reproduced across multiple synchronization events.

The important observation is that enabling `AFTER_UTXO_CHANGE` changes the request count from:

```text
1 × /utxos
1 × /displaywallet
```

to:

```text
2 × /utxos
2 × /displaywallet
```

for the same synchronization event.

## Impact

The same wallet synchronization can perform a second complete refresh unnecessarily, resulting in:

* additional `/utxos` requests
* additional `/displaywallet` requests
* unnecessary backend work
* redundant frontend refetch/loading activity
* additional network traffic during wallet synchronization

While this may not cause incorrect wallet state, it introduces redundant work into the synchronization flow.

## Possible fix

The `AFTER_UTXO_CHANGE` flow could be adjusted so that a UTXO update produced by `refetchWalletBalance()` does not immediately trigger another full wallet refresh.

One possible approach is to make the display-wallet query explicitly depend on the relevant UTXO state through its `queryKey`, then remove the manual refetch cascade if the existing synchronization behavior is preserved.

Another possible approach would be to distinguish between:

1. UTXO changes caused by an external synchronization event, and
2. UTXO changes that are already the result of `refetchWalletBalance()`.

This would prevent the second effect from treating the UTXO update produced by the first refresh as a new synchronization event.

I'd be happy to investigate or implement a fix if this behavior is considered unintended.

Amanyadav0556:
## What does this PR do?

This PR fixes **Issue #1447** by preventing an unnecessary wallet refetch when `WalletInfoAutoReload` mounts before any UTXO state has actually changed.

Previously, the `AFTER_UTXO_CHANGE` effect could execute during the initial mount and trigger an unnecessary `refetchWalletBalance()` call. This was especially observable under React `StrictMode`, where mount/effect behavior can be invoked more than once during development.

### Changes

* Replaced the initial-mount boolean flag in `WalletInfoAutoReload` with a `useRef` that tracks the previous `utxosHashHex`.
* Added a strict hash comparison so the refetch is skipped when the current UTXO hash has not changed.
* Preserved the existing refetch behavior when the UTXO hash genuinely changes.
* Updated `src/App.test.tsx` with `StrictMode` coverage.
* Stabilized the `jmTxStore` mock to prevent unnecessary render loops in the test suite.
* Added integration coverage for initial mount, subsequent UTXO hash changes, and unchanged hash values.

### Expected behavior

* Initial mount → no unnecessary wallet refetch.
* `EMPTY_HASH → ACTUAL_HASH` → wallet refetch occurs.
* Subsequent UTXO hash change → wallet refetch occurs.
* Same UTXO hash → no additional refetch.
* React `StrictMode` → no duplicate initial refetch.

### Validation

The following checks pass successfully:

* `npx vitest run src/App.test.tsx` — 6 tests passed

* `npx tsc --noEmit` — passed

* `npm run lint` — passed

* Fixes #1447

## Visual Demo

No visual UI changes were introduced by this PR. This is a behavior/lifecycle fix related to wallet UTXO refetching, so a visual demo is not applicable.

### DIFF
--- a/src/App.test.tsx
+++ b/src/App.test.tsx
@@ -268,7 +268,7 @@ describe('WalletInfoAutoReload', () => {
     const { rerender } = render(
       <StrictMode>
         <App />
-      </StrictMode>
+      </StrictMode>,
     )
 
     // 1. Verify refetchWalletBalance is NOT called because of the initial mount
@@ -280,7 +280,7 @@ describe('WalletInfoAutoReload', () => {
     rerender(
       <StrictMode>
         <App />
-      </StrictMode>
+      </StrictMode>,
     )
     await waitFor(() => expect(refetchWalletBalance).toHaveBeenCalledTimes(1))
 
@@ -289,24 +289,26 @@ describe('WalletInfoAutoReload', () => {
     rerender(
       <StrictMode>
         <App />
-      </StrictMode>
+      </StrictMode>,
     )
     await waitFor(() => expect(refetchWalletBalance).toHaveBeenCalledTimes(2))
 
-    // 4. Test that the existing refetch behavior arguments remain unchanged 
+    // 4. Test that the existing refetch behavior arguments remain unchanged
     // It should be called with { delayBefore: 210, signal: <AbortSignal> }
-    expect(refetchWalletBalance).toHaveBeenLastCalledWith(expect.objectContaining({
-      delayBefore: 210,
-      signal: expect.any(AbortSignal) as unknown as AbortSignal
-    }))
+    expect(refetchWalletBalance).toHaveBeenLastCalledWith(
+      expect.objectContaining({
+        delayBefore: 210,
+        signal: expect.any(AbortSignal) as unknown as AbortSignal,
+      }),
+    )
 
     // 6. Test that same hash doesn't trigger another refetch
     rerender(
       <StrictMode>
         <App />
-      </StrictMode>
+      </StrictMode>,
     )
-    await new Promise(r => setTimeout(r, 50))
+    await new Promise((r) => setTimeout(r, 50))
     expect(refetchWalletBalance).toHaveBeenCalledTimes(2)
   })
 })
--- a/src/App.test.tsx
+++ b/src/App.test.tsx
@@ -256,7 +256,6 @@ describe('App', () => {
 describe('WalletInfoAutoReload', () => {
   beforeEach(() => {
     vi.clearAllMocks()
-    window.history.pushState({}, '', '/')
     holders.walletFileName = 'wallet.jmdat'
     holders.token = 'tok'
     holders.refreshToken = 'refresh'
--- a/src/App.test.tsx
+++ b/src/App.test.tsx
@@ -272,8 +272,6 @@ describe('WalletInfoAutoReload', () => {
     )
 
     // 1. Verify refetchWalletBalance is NOT called because of the initial mount
-    // Yield to the event loop so the synchronous effects execute
-    await new Promise((r) => setTimeout(r, 0))
     expect(refetchWalletBalance).not.toHaveBeenCalled()
 
     // 2. A subsequent utxosHashHex change DOES call refetchWalletBalance
@@ -309,7 +307,6 @@ describe('WalletInfoAutoReload', () => {
         <WalletInfoAutoReload />
       </StrictMode>,
     )
-    await new Promise((r) => setTimeout(r, 50))
     expect(refetchWalletBalance).toHaveBeenCalledTimes(2)
   })
 })
--- a/src/App.test.tsx
+++ b/src/App.test.tsx
@@ -1,7 +1,7 @@
 import { type ReactNode, StrictMode } from 'react'
 import { fireEvent, render, screen, waitFor } from '@testing-library/react'
 import { describe, expect, it, vi, beforeEach } from 'vitest'
-import App from './App'
+import App, { WalletInfoAutoReload } from './App'
 
 type Holders = {
   walletFileName?: string
@@ -267,19 +267,20 @@ describe('WalletInfoAutoReload', () => {
   it('skips refetch on initial mount but refetches on subsequent utxosHashHex changes', async () => {
     const { rerender } = render(
       <StrictMode>
-        <App />
+        <WalletInfoAutoReload />
       </StrictMode>,
     )
 
     // 1. Verify refetchWalletBalance is NOT called because of the initial mount
-    await waitFor(() => expect(screen.getByText('main-wallet-page')).toBeInTheDocument())
+    // Yield to the event loop so the synchronous effects execute
+    await new Promise((r) => setTimeout(r, 0))
     expect(refetchWalletBalance).not.toHaveBeenCalled()
 
     // 2. A subsequent utxosHashHex change DOES call refetchWalletBalance
     holders.utxosHashHex = 'changed-hash-1'
     rerender(
       <StrictMode>
-        <App />
+        <WalletInfoAutoReload />
       </StrictMode>,
     )
     await waitFor(() => expect(refetchWalletBalance).toHaveBeenCalledTimes(1))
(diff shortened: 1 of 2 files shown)


Amanyadav0556:
Hi @parrth20 

Both of my checks have passed successfully .

Could you please review the PR when you get a chance? If everything looks good, please merge it.

Thanks!

src/App.test.tsx
```diff
@@ -266,3 +252,64 @@ describe('App', () => {
...
+  it('skips refetch on initial mount but refetches on subsequent utxosHashHex changes', async () => {
+    const { rerender } = render(
+      <StrictMode>
+        <WalletInfoAutoReload />
+      </StrictMode>,
+    )
+
+    // 1. Verify refetchWalletBalance is NOT called because of the initial mount
+    // Yield to the event loop so the synchronous effects execute
+    await new Promise((r) => setTimeout(r, 0))
```
parrth20: these real-time waits seem unnecessary because render and rerender already flush synchronous effects...Could we remove this wait and the 50ms wait on line 312?? the assertions pass immediately without them, and removing them avoids adding timing-related flakiness.

theborakompanioni: Done in [0da8982](https://github.com/joinmarket-webui/jam/pull/1448/commits/0da89825e542e498bc006cf5b815fbcee6ff7bf3).

Amanyadav0556: @parrth20 Done, I removed both the 0ms and 50ms waits as you suggested. All checks are passing now. Please have a look when you get time. Thanks!

GuTS805:
@theborakompanioni quick question before this gets merged. Am I reading this right that it doesn't actually close #1447?

My understanding: utxosHashHex is already a dependency of the effect, and refetchWalletBalance is a stable reference (mutateAsync from useMutation, bound once on the observer). So the effect can only re-run when the hash has genuinely changed, which means previousUtxosHashHexRef.current is always holding the previous hash and never matches the current one. If that's correct, the guard can only short-circuit on the very first run. functionally a "skip the first effect run" flag.
The redundant refetch in the issue is a genuine hash change though, so it would slip right past it:

AFTER_TX_MESSAGE → refetchWalletBalance() → /utxos returns a new set
→ utxosHashHex changes → AFTER_UTXO_CHANGE fires → refetchWalletBalance() again
which would still be 2× /utxos + 2× /displaywallet per sync event mid-session, matching the evidence table in the issue. The new test seems to encode this too — step 2 asserts a refetch does happen on a subsequent hash change.

It does remove one refetch at mount (and the StrictMode double, though that's dev-only), so it's a real improvement either way. I just want to make sure the mid-session case isn't being closed off prematurely. Happy to be corrected if I'm missing something.

@Amanyadav0556 if the above holds, one option would be the first approach from the issue. Let react-query handle it. useQueryDisplayWallet already receives utxosHashHex, but it's passed as meta.cacheBuster, and meta isn't part of the generated createQueryKey (it only uses _id/baseUrl/body/headers/path/query), so it isn't busting anything today. If that value were part of the query key instead, displaywallet would refetch on utxo changes on its own and the AFTER_UTXO_CHANGE effect could be dropped entirely rather than guarded.

src/App.test.tsx
```diff
@@ -266,3 +252,61 @@ describe('App', () => {
     await waitFor(() => expect(screen.getByText('main-wallet-page')).toBeInTheDocument())
   })
 })
+
+describe('WalletInfoAutoReload', () => {
+  beforeEach(() => {
+    vi.clearAllMocks()
+    window.history.pushState({}, '', '/')
```
theborakompanioni:
```suggestion
```

theborakompanioni:
> @theborakompanioni quick question before this gets merged. Am I reading this right that it doesn't actually close #1447?

> @Amanyadav0556 if the above holds, one option would be the first approach from the issue. Let react-query handle it. useQueryDisplayWallet already receives utxosHashHex, but it's passed as meta.cacheBuster, and meta isn't part of the generated createQueryKey (it only uses _id/baseUrl/body/headers/path/query), so it isn't busting anything today. If that value were part of the query key instead, displaywallet would refetch on utxo changes on its own and the AFTER_UTXO_CHANGE effect could be dropped entirely rather than guarded.

Agreed. `WalletInfoAutoReload` component was migrated from the legacy branch that did not use react-query. The whole approach can probably entirely replaced and simplified by using query keys.

Would not mind this being merged (including the test) and if somebody wants to pick it up, improve in a follow-up PR.

GuTS805:
@Amanyadav0556 since you already have the context here, the follow-up is yours if you want it. Replacing the manual AFTER_UTXO_CHANGE effect with a proper query key on the displaywallet query.

Amanyadav0556:
@theborakompanioni Yeah sure, I can pick this up. I’ll work on it in a separate PR once this one gets merged.

theborakompanioni (approved):

PR #1454, opened 2026-08-24
fix: always show balances in orderbook

### LINKED ISSUE #1453: fix: always show balances in orderbook
Orderbook view hides balances when in private mode.
This is not necessary for the values in the local orderbook.

Image

Georgefifth:
## What does this PR do?

Problem:
- In private mode the orderbook view masks fee, min/max order size and market summary values. These are public market data from the local orderbook, not wallet balances, so hiding them is unnecessary and makes the orderbook unreadable.

Solution:
- Pass `showBalance={true}` to those `Balance` components (table: fee, min/max size, miner fee contribution; content: market summary). Same pattern already used in the bond value tooltip in the same table, so no new mechanism.

Result:
- `src/components/orderbook/OrderbookTable.privateMode.test.tsx` renders the table with private mode enabled and asserts the values render unmasked; it fails without the fix.
- Full unit suite: 903/903 passing (`npm test` with NODE_OPTIONS=--no-webstorage). `tsc -b`, eslint and prettier clean.

- Fixes #1453

## Visual Demo

Regtest environment not available here, so covered by the private-mode rendering test instead. Before: values show as the `*****` mask in private mode. After: plain sats values.

### DIFF
--- a/src/components/orderbook/OrderbookContent.tsx
+++ b/src/components/orderbook/OrderbookContent.tsx
@@ -263,6 +263,7 @@ export const OrderbookContent = ({ enabled, className }: OrderbookContentProps) 
                       <span>{t('orderbook.market_summary_median_abs_fee_label')}:</span>
                       <Balance
                         valueString={String(Math.round(marketSummary.medianAbsFee))}
+                        showBalance={true}
                         enableVisibilityToggle={false}
                       />
                     </p>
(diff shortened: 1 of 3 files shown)


theborakompanioni (approved):
@Georgefifth Amazing. Thank you 🙏

PR #1457, opened 2026-08-25
fix(orderbook): count each maker once in total liquidity

### LINKED ISSUE #1456: Total liquidity in the orderbook counts the same maker's coins more than once
Was poking around the Market Summary and the "Total liquidity" number looked higher than it
should be.

It sums `maxsize` across every offer row (`OrderbookContent.tsx:211`), but makers often advertise
several offers backed by the same coins either `--dual-offers`, or a yield generator splitting
one balance into contiguous size bands. A taker can only fill one of them, so those sats end up
counted more than once.

Clearest example in the current mainnet orderbook one maker, two offers, same size range, same
bond UTXO:

```json
{"counterparty":"J54JcU8wv1JqxsKw","oid":0,"ordertype":"sw0reloffer","minsize":27300,"maxsize":8388608,
 "cjfee":"0.001","fidelity_bond_data":{"utxo_txid":"f0da1772...bed52d0"}}
{"counterparty":"J54JcU8wv1JqxsKw","oid":1,"ordertype":"sw0absoffer","minsize":27300,"maxsize":8388608,
 "cjfee":250,"fidelity_bond_data":{"utxo_txid":"f0da1772...bed52d0"}}
```

Those 8,388,608 sats get counted twice. Of the 9 makers currently advertising more than one offer,
all 9 have every offer on a single bond UTXO. One of them (`J59SWfLvRkZHt2KE`) tiles its balance
across 21 bands, so its real ceiling of 41,643,514 shows up as 235,336,811.

Across the whole snapshot (1506 offers, 1473 makers) it comes out at 2757.55 BTC displayed vs
2686.20 actual about 2.7% high. Not dramatic today, but it grows with however many makers split
their offers.

What makes me think it's an oversight rather than intentional: `bondedMakers`, nine lines up in
the same `useMemo`, already dedupes by counterparty, and the fee bands chart does too. And in
#1103 a reviewer raised exactly this about the bonded/unbonded count *"since a counterparty can
have multiple offers, this will over-count"*  which got fixed, and is why `counterpartyBonds`
exists. Total liquidity just never got the same treatment.

Fix seems small: keep a map of each maker's largest `maxsize` and sum that, same shape as
`counterpartyBonds` right above it.

Unless "total liquidity" is meant as "everything advertised", in which case maybe it's the label
that needs changing instead. 

Happy to open a PR either way.

CapThunder19:
Closes #1456.

`totalLiquidity` was summing `maxsize` over every offer row, so a maker advertising several offers
on the same coins got counted more than once. Now it keeps each maker's largest offer in a map and
sums that — same shape as the `counterpartyBonds` map just above it.

Left the rest of `marketSummary` alone. `maxSizes` stays since `maxOfferSize` really is per-offer.

Added a test file for `OrderbookContent` (there wasn't one). It renders the component and reads the
value that ends up in the DOM: distinct makers as a control, a dual-offer maker with the same range
twice, one tiling a balance across several bands, and a dual-offer maker with a split range. I
reverted the fix to check they actually catch it — the three dedupe cases fail, the control passes.

The medians are still per-offer. The difference is tiny right now (5141.5 vs 5220) since hardly any
maker publishes two offers of the same type, so I kept this to the reported bug. Can do it
separately if you want.

### DIFF
--- a/src/components/orderbook/OrderbookContent.test.tsx
+++ b/src/components/orderbook/OrderbookContent.test.tsx
@@ -1,0 +1,104 @@
+import type React from 'react'
+import { QueryClient, QueryClientProvider } from '@tanstack/react-query'
+import { render, screen, waitFor } from '@testing-library/react'
+import { describe, expect, it, vi } from 'vitest'
+import type { OrderbookOffer } from '@/lib/api/orderbook'
+import { OrderbookContent } from './OrderbookContent'
+
+vi.mock('react-i18next', () => ({
+  useTranslation: () => ({
+    t: (key: string) => key,
+    i18n: { t: (key: string) => key, language: 'en', resolvedLanguage: 'en' },
+  }),
+}))
+
+vi.mock('@/components/ui/tooltip', () => ({
+  Tooltip: ({ children }: { children: React.ReactNode }) => <span>{children}</span>,
+  TooltipContent: ({ children }: { children: React.ReactNode }) => <span>{children}</span>,
+  TooltipTrigger: ({ children }: { children: React.ReactNode }) => <>{children}</>,
+  TooltipProvider: ({ children }: { children: React.ReactNode }) => <>{children}</>,
+}))
+
+vi.mock('@/components/ui/jam/Balance', () => ({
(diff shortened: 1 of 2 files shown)


theborakompanioni (approved):
Nice 🚀

PR #1461, opened 2026-08-27
build(deps): update to joinmarket-api-ts v0.37.1

theborakompanioni:
Switches from joinmarket-ng-api-ts v1.0.0 to joinmarket-api-ts v0.37.1 which is now compatible with joinmarket-ng. Client regenerated and namespace changed to original.
The version of `joinmarket-api-ts` will now be sychronized with the version of `joinmarket-ng`.


parrth20 (approved):
LGTM !!

PR #1463, opened 2026-08-27
Send: pin exact utxos when sweeping instead of re-deriving at broadcast time

### LINKED ISSUE #1422: Send: sweep amount is snapshotted at click time and never re-synced, so the confirmation dialog can show a different amount than what is actually swept
**Expected behavior**

The amount shown in the payment confirmation dialog for a sweep should match the amount that will actually be swept, or otherwise stay in sync with the source jar's available balance while the form is open.

**Actual behavior**

Clicking **Sweep** snapshots the source jar's available balance into the form field `amount.sweepAmount`. That snapshot is never recomputed when the jar's available balance changes. Including when the change is caused by the user's own action on the same page (the UTXO selection dialog, which freezes/unfreezes UTXOs and then refetches wallet info).

The request actually sent to the backend is `amount_sats: 0`. A real sweep, resolved server-side against whatever is spendable at broadcast time. So the number the user reads and approves, and the amount the wallet actually moves, are two independent values that can diverge in either direction.

**1. The value is captured once, at click time** — `src/components/send/SendForm.tsx:503-509`:

```tsx
onClick={() => {
  setValue('amount.isSweep', true, { shouldValidate: true })
  setValue('amount.sweepAmount', sourceJar?.balanceSummary.calculatedAvailableBalanceInSats, {
    shouldValidate: true,
  })
  setValue('amount.amount', undefined, { shouldValidate: true })
}}
```

**2. Nothing re-syncs it.** A repo-wide grep for `sweepAmount` shows the only non-test writes are `SendForm.tsx:345` (reset on *jar change*), `SendForm.tsx:505` (the capture above) and `SendForm.tsx:547` (reset on *clear sweep*). There is no write on balance change. The one effect that does react to balance changes only re-validates the jar field — `src/components/send/SendForm.tsx:202-209`:

```tsx
useEffect(() => {
  if (formState.dirtyFields.source?.fromJar) {
    console.debug('Validating `source.fromJar` as wallet balance summary changed (e.g. frozen/unfrozen utxos, new transaction, etc.)')
    void trigger('source.fromJar')
  }
}, [walletBalanceSummary, trigger, formState.dirtyFields.source?.fromJar])
```

The comment explicitly names "frozen/unfrozen utxos" as a trigger, but `amount.sweepAmount` is not touched.

**3. Validation cannot catch it.** `SendForm.schema.ts:112-120` validates `sweepAmount` only as `integer().min(1).max(MAX_SEND_AMOUNT)`, and the balance guard at `SendForm.schema.ts:195-197` explicitly opts out for sweeps (`if (root.amount.isSweep) return true`).

**4. The stale value is what the user approves.** `src/components/send/PaymentConfirmDialog.tsx:144` renders `values.amount?.sweepAmount` verbatim, and `PaymentConfirmDialog.tsx:29` also derives the "Collaborator fee ≤ X (Y%)" figure from it.

**5. The transaction ignores it.** `src/components/send/collaborativeSend.ts:40` and `:63` both send `amount_sats: 0` for a sweep.

**6. Internal inconsistency confirming the oversight.** In the same file, the *inline* fee estimate reads the **live** balance for sweeps — `src/components/send/SendForm.tsx:225-230`:

```tsx
const amountForFeeEstimate = useMemo(() => {
  if (values.amount?.isSweep === true) {
    return sourceJar?.balanceSummary.calculatedAvailableBalanceInSats   // live
  }
  return values.amount?.amount
}, [sourceJar, values.amount?.amount, values.amount?.isSweep])
```

while the confirm dialog reads the snapshot. Two different numbers for the same concept, in the same flow.

1. Open the Send page and select a source jar that has some available balance and some frozen UTXOs (e.g. 100,000 sats available, 900,000 sats frozen).
2. Click **Sweep**. The form shows "100,000 sats".
3. Click the **Select UTXOs** button next to the source-jar label. It stays enabled during a sweep — `useUtxoSelectionDialog.ts:245` gates it only on `isSubmitting || sourceJar === undefined || sourceJar.utxos.length === 0`, with no `isSweep` check.
4. Tick the frozen UTXOs and confirm. `useUtxoSelectionDialog.ts:157` computes `utxosToUnfreeze`, unfreezes them, and `:110` chains `walletInfoRefetch()`. The jar's available balance is now 1,000,000 sats.
5. Submit. The confirmation dialog still shows **"Sweep 100,000 sats"**. Confirm it.
6. `amount_sats: 0` goes to the backend and **1,000,000 sats** leave the jar.

**Additional context**

The confirmation dialog is the last thing between the user and an irreversible on-chain transaction, and here it shows a concrete sats figure that the code path taken afterwards does not use. In the unfreeze direction the wallet moves more than the user approved; in the freeze direction it moves less, and the coinjoin can fall below the maker's `minsize` and fail. The same stale figure also feeds `estimateMaxCollaboratorFee`, so the "Collaborator fee ≤ X (Y%)" line and its percentage are wrong too.

I'm aware of the existing tooltip `send.confirm_send_modal.text_sweep_info_popover` ("The exact transaction amount can only be calculated by JoinMarket at the point when the transaction is made."). That hedge reads as covering mining-fee deduction — a small delta — rather than an arbitrary difference caused by a different UTXO set. The fact that `amountForFeeEstimate` already uses the live balance suggests the live value is available and intended to be used; the confirm dialog just wasn't updated.

This is not attacker-exploitable — it requires the user's own freeze/unfreeze action — so I'm filing it publicly rather than reporting it privately. Happy to move it if maintainers prefer otherwise.

**Proposed fix direction**

Stop storing a snapshot: keep only `amount.isSweep` in the form and derive the displayed amount from the live source jar wherever it is rendered (`SendForm.tsx:529`, `PaymentConfirmDialog.tsx:144`, `PaymentConfirmDialog.tsx:29`), the way `amountForFeeEstimate` already does. If the field must stay for schema reasons, add an effect that re-runs `setValue('amount.sweepAmount', ...)` whenever the available balance changes while `isSweep` is true, and/or have `PaymentConfirmDialog` re-prompt if the captured value no longer matches `meta.sourceJar.balanceSummary.calculatedAvailableBalanceInSats`.

GuTS805:
## What does this PR do?

Sweep amount was captured once when the button was clicked, but the actual utxos to spend were only decided by the backend at broadcast time. If the jar's contents changed in between (another tx confirms, a utxo gets frozen, etc), the swept amount could silently drift from what the backend actually sends.

This pins the exact utxos backing the jar's available balance at the moment "Sweep" is clicked, and sends them as `input_utxos` so the backend sweeps precisely those utxos instead of re-deriving "everything unfrozen in the mixdepth" later.

- Fixes #1422

### DIFF
--- a/src/components/send/SendForm.tsx
+++ b/src/components/send/SendForm.tsx
@@ -18,6 +18,7 @@ import { useQueryOrderbook } from '@/hooks/useQueryOrderbook'
 import type { BalanceSummary } from '@/lib/balanceSummary'
 import { parseBip21Uri, type Bip21ParseResult } from '@/lib/bip21'
 import type { JamFeeConfigValues } from '@/lib/feeConfig'
+import * as fb from '@/lib/fidelityBondUtils'
 import { cn, delayedPromise, factorToPercentage, type WalletFileName } from '@/lib/utils'
 import type { BitcoinAddress, JarIndex } from '@/types/global'
 import { DevBadge } from '../dev/DevBadge'
@@ -94,6 +95,13 @@ const AddressFromJarSelectorDialog = ({
     />
   )
 }
+
+// Same "available" definition as `toBalanceSummary`: not frozen and not a
+// timelocked fidelity bond. Captured at the moment "Sweep" is clicked so the
+// exact utxos can be pinned via `input_utxos`, instead of the backend
+// re-deriving "everything unfrozen in the mixdepth" at broadcast time.
+const availableUtxosForSweep = (jar: Jar | undefined) =>
+  (jar?.utxos ?? []).filter((utxo) => !utxo.frozen && !fb.utxo.isLocked(utxo)).map((utxo) => utxo.utxo)
 
 const FieldPrefixSatSymbol = (
   <SatSymbol
@@ -355,6 +363,7 @@ export function SendForm({
                         if (isSweep === true) {
                           setValue('amount.isSweep', false, { shouldValidate: true })
                           setValue('amount.sweepAmount', undefined, { shouldValidate: true })
+                          setValue('amount.sweepUtxos', undefined, { shouldValidate: true })
                           setValue('amount.amount', undefined, { shouldValidate: true })
                         }
                         if (destinationJarIndex === jar.jarIndex) {
@@ -520,6 +529,7 @@ export function SendForm({
                     setValue('amount.sweepAmount', sourceJar?.balanceSummary.calculatedAvailableBalanceInSats, {
                       shouldValidate: true,
                     })
+                    setValue('amount.sweepUtxos', availableUtxosForSweep(sourceJar), { shouldValidate: true })
                     setValue('amount.amount', undefined, { shouldValidate: true })
                   }}
                 >
@@ -560,6 +570,7 @@ export function SendForm({
                   onClick={() => {
                     setValue('amount.isSweep', false, { shouldValidate: true })
                     setValue('amount.sweepAmount', undefined, { shouldValidate: true })
+                    setValue('amount.sweepUtxos', undefined, { shouldValidate: true })
                     setValue('amount.amount', undefined, { shouldValidate: true })
                   }}
                 >
@@ -570,6 +581,7 @@ export function SendForm({
               {errors.amount?.sweepAmount?.message ? (
                 <FieldError>{errors.amount.sweepAmount.message}</FieldError>
               ) : null}
+              {errors.amount?.sweepUtxos?.message ? <FieldError>{errors.amount.sweepUtxos.message}</FieldError> : null}
               {errors.amount?.isSweep?.message ? <FieldError>{errors.amount.isSweep.message}</FieldError> : null}
             </Field>
           </div>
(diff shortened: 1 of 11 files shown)


src/components/send/SendForm.tsx
```diff
@@ -95,6 +96,13 @@ const AddressFromJarSelectorDialog = ({
   )
 }
 
+// Same "available" definition as `toBalanceSummary`: not frozen and not a
+// timelocked fidelity bond. Captured at the moment "Sweep" is clicked so the
+// exact utxos can be pinned via `input_utxos`, instead of the backend
+// re-deriving "everything unfrozen in the mixdepth" at broadcast time.
+const availableUtxosForSweep = (jar: Jar | undefined) =>
```
theborakompanioni, line 103: Off-topic: Makes me wonder if sweeping expired fidelity bonds should even be possible from the Send page.. 
With jm-ng as backend, they can be unfrozen while still time-locked.. if they expire and users are not careful, they are swept.

Maybe something for a follow-up.

theborakompanioni (dismissed):

src/components/send/SendForm.tsx
GuTS805: Yeah, that's a real gap. isLocked goes false the moment the locktime passes, so an expired bond sitting unfrozen just lands in the available set and gets swept with everything else.

Worth noting SendPage already filters bonds out of availableUtxosForPayment for non-coinjoin sends, so the intent exists, it just doesn't reach the sweep path.

Happy to take it as a follow-up.

theborakompanioni (approved):

PR #1466, opened 2026-08-28
fix: use utxos hash in displaywallet query key

Amanyadav0556:
## Summary

Follow-up to #1448.

This PR replaces the manual `AFTER_UTXO_CHANGE` wallet refetch mechanism with a React Query query-key dependency.

## Changes

- Make `utxosHashHex` part of the `displaywallet` query identity.
- Remove the `meta.cacheBuster` dependency for this behavior.
- Remove the manual `AFTER_UTXO_CHANGE` refetch path.
- Simplify the wallet auto-reload flow.
- Add/update tests for the query-key based behavior.

## Result

Previous flow:

UTXO change
? AFTER_UTXO_CHANGE
? manual refetch

New flow:

UTXO change
? utxosHashHex changes
? displaywallet query key changes
? React Query handles the fetch

## Validation

- Focused tests: PASS
- Typecheck: PASS
- Lint: PASS
- Build: PASS
- git diff --check: PASS

This is the follow-up approach discussed during review of #1448.

### DIFF
--- a/src/App.test.tsx
+++ b/src/App.test.tsx
@@ -97,6 +97,7 @@ vi.mock('@tanstack/react-query', () => ({
 
 vi.mock('@joinmarket-webui/joinmarket-api-ts/@tanstack/react-query', () => ({
   lockwalletOptions: () => ({ queryKey: ['lock'] }),
+  displaywalletOptions: () => ({ queryKey: [{ _id: 'displaywallet' }], queryFn: vi.fn() }),
 }))
 
 vi.mock('@joinmarket-webui/joinmarket-api-ts/jm', () => ({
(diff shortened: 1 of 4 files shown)


theborakompanioni:
> I gone through the pr, keepPreviousData looks like the right call here. One thing I noticed: I removed that line locally and the suite still passed 914/914. The test only grabs options.queryKey, so placeholderData isn't really covered.
> 
> useQueryUtxos.test.ts does queryMock.mockImplementation((options) => options) and asserts on what comes back (uses it for enabled). Might work here too, up to you though.

@Amanyadav0556 Do you plan to address this comment? Share your thoughts! :slightly_smiling_face:

PR #1467, opened 2026-08-30
fix(earn): pin the reviewed utxos when creating a fidelity bond

### LINKED ISSUE #1460: Create Fidelity Bond: UTXOs arriving in the jar after the freeze step get swept into the time-locked bond
**Expected behavior**

Only the UTXOs the user selected (and reviewed on the confirmation screen) should end up in the time-locked fidelity bond address.

**Actual behavior**

The create-bond wizard works around the fact that `direct-send` only takes a `mixdepth` (not a UTXO list) by freezing every other UTXO in the jar, then sweeping the whole mixdepth. That freeze is a one-time snapshot, and nothing re-checks it before broadcasting.

`useCreateFidelityBondWizard.ts:135-140` builds `utxosToFreeze` as a `useMemo` over the jar's UTXOs at that moment:

```ts
const utxosToFreeze = useMemo(() => {
  if (selectedJar === undefined) return []
  return selectedJar.utxos.filter(
    (utxo) => !utxo.frozen && !fb.utxo.isFidelityBond(utxo) && !selectedUtxos.some((it) => it.utxo === utxo.utxo),
  )
}, [selectedJar, selectedUtxos])
```

`handleFreezeUtxos` (`:247-272`) freezes that list, then moves to the `review` step. `handleCreateFidelityBond` (`:274-288`) then sweeps the whole mixdepth with no re-check:

```ts
const result = await directSend.mutateAsync({
  path: { walletname: walletFileName },
  body: {
    mixdepth: selectedJar.jarIndex,
    amount_sats: 0, // 0 := sweep!
    destination: address,
  },
})
```

Between those two points sits the `review` step, which requires the user to read a warning and manually tick a confirmation checkbox (`CreateFidelityBondDialogSteps.tsx:303-307`) before clicking through — so the gap is arbitrary in length and entirely user-paced.

Anything unfrozen that lands in that jar during the gap is swept into the bond along with the selected funds. The review screen shows `totalAmount`, which is the sum of the *selected* UTXOs only (`:147`), so the amount the user confirms can be less than what actually gets locked.

1. Have a jar with one spendable UTXO (e.g. 0.01 BTC).
2. Open Create Fidelity Bond, pick a lock date, pick that jar, select the UTXO, continue to the review screen. It shows "Amount: 0.01 BTC".
3. Before ticking the checkbox and confirming, have a new unfrozen UTXO arrive in that same jar — a deposit to a receive address in that jar, or coinjoin change if a maker is running in that mixdepth.
4. Tick the checkbox and confirm.
5. The broadcast sweeps the whole mixdepth: both the reviewed 0.01 BTC and the newly-arrived UTXO get locked into the timelock address, for the full bond duration.

**Additional context**

I don't think this is intentional — the UI explicitly promises the opposite. The freeze step is titled "Freeze Remaining UTXOs" / "Protect your privacy by freezing unselected UTXOs", and its description says *"They will be frozen to remain in Jar {{ jar }}"* (`translation.json:648-657`). The freeze in `useFidelityBondSweep.ts:35` carries the comment *"Freeze other UTXOs in the source jar so only the FB gets swept"*. Both state a guarantee the snapshot approach can't actually provide.

Scope note: `useFidelityBondSweep.ts` (used by renew / move-to-jar) has the same snapshot shape, but there the freeze loop and the sweep run back-to-back inside a single mutation with the user confirmation happening before all of it (`RenewBondDialog.tsx:110-115`), so the window is just network latency rather than user-paced. The create wizard is the one with a genuinely open-ended gap.

Prior fidelity-bond fixes in this area (#1397, #1399, #1048, #1442) all deal with failure/rollback or post-success refetch handling — none covers this snapshot-vs-live-state gap. #1346/#1347 track the same bug class but scoped to the sweep scheduler.

This is a fund-safety issue rather than a display one — locked funds are inaccessible for the chosen duration, and the UI supports multi-year lockdates. It's self-contained though (only affects the wallet owner's own funds, not exploitable by anyone else), which is why I'm filing it publicly rather than reporting privately.

**Proposed fix direction**

Short term: immediately before `directSend`, re-fetch the jar's current UTXO set and either freeze anything new that isn't selected, or abort and send the user back to review if new funds appeared.

Longer term this workaround can go away entirely — `direct-send` now accepts an explicit `input_utxos` list on the backend (joinmarket-ng#588, which closes out what #772 asked for), so the sweep could just name the selected UTXOs instead of freezing everything else. That path needs the client package release that #1422 and #1437 are also waiting on.

GuTS805:
## What does this PR do?

The create-bond wizard froze every other utxo in the jar and then swept the whole mixdepth, because `direct-send` only took a `mixdepth`. That freeze is a snapshot taken before the review step, and nothing re-checked it before broadcasting — so anything unfrozen landing in the jar while the user read the warning and ticked the confirmation checkbox got locked into the bond too, for the full bond duration. The review screen shows the sum of the *selected* utxos, so the amount the user confirms could be less than what actually got locked.

`direct-send` now accepts an explicit `input_utxos` list, so the sweep names the reviewed utxos directly and spends exactly those, regardless of what else is sitting unfrozen in the mixdepth at broadcast time. It also bails out instead of broadcasting if the reviewed utxo is no longer in the jar, since an empty `input_utxos` is rejected by the API and the old code would have swept the whole mixdepth in that case.

- Fixes #1460

### Note on the freeze step

I left the `utxosToFreeze` machinery and the "Freeze Remaining UTXOs" step in place. It's redundant now that the sweep names its inputs, but removing it touches the wizard steps, the i18n keys and a fair amount of test setup, so it felt better as a separate change than bundled into a fund-safety fix. Happy to do it here instead if you'd prefer.

### Tests

Two new tests in `useCreateFidelityBondWizard.test.ts`:

- a utxo arriving in the jar between the review step and confirmation is not swept into the bond
- no broadcast happens if the reviewed utxo is gone from the jar by the time the user confirms

I checked both actually pin the behaviour down — reverting the fix makes them fail (along with the existing direct-send assertion), rather than passing regardless.

## Visual Demo

No UI changes — this only affects which utxos the broadcast spends.

### DIFF
--- a/src/components/earn/CreateFidelityBondDialog/useCreateFidelityBondWizard.test.ts
+++ b/src/components/earn/CreateFidelityBondDialog/useCreateFidelityBondWizard.test.ts
@@ -191,12 +191,64 @@ describe('useCreateFidelityBondWizard', () => {
         mixdepth: 0,
         amount_sats: 0,
         destination: 'bcrt1qtimelockdestination',
+        input_utxos: ['only:0'],
       },
     })
     expect(result.current.step).toBe('success')
     expect(result.current.txResult).toEqual({ txid: 'created-tx' })
     expect(mocks.walletInfoRefetch).toHaveBeenCalled()
     expect(mocks.toastSuccess).toHaveBeenCalledWith('earn.fidelity_bond.create_fidelity_bond.success_text')
+  })
+
+  it('sweeps only the reviewed utxos when new funds arrive in the jar before confirming', async () => {
+    mocks.walletInfo.jars = [jar(0, [utxo({ utxo: 'reviewed:0', value: 25_000 })])]
+    const { result, rerender } = renderHook(() => useCreateFidelityBondWizard(true, vi.fn(), 'wallet.jmdat'))
+
+    act(() => result.current.setSelectedLockdate(minLockdateOption()))
+    await act(async () => result.current.handleNext())
+    act(() => result.current.toggleUtxoSelection(result.current.availableUtxos[0]))
+    await act(async () => result.current.handleNext())
+
+    expect(result.current.step).toBe('review')
+    expect(result.current.totalAmount).toBe(25_000)
+
+    mocks.walletInfo.jars = [
+      jar(0, [utxo({ utxo: 'reviewed:0', value: 25_000 }), utxo({ utxo: 'arrived-late:0', value: 80_000 })]),
+    ]
+    rerender()
+
+    act(() => result.current.setConfirmationChecked(true))
+    await act(async () => result.current.handleNext())
+
+    expect(mocks.directSendMutateAsync).toHaveBeenCalledWith({
+      path: { walletname: 'wallet.jmdat' },
+      body: {
+        mixdepth: 0,
+        amount_sats: 0,
+        destination: 'bcrt1qtimelockdestination',
+        input_utxos: ['reviewed:0'],
+      },
+    })
+  })
+
+  it('does not broadcast when the reviewed utxo is no longer in the jar', async () => {
+    mocks.walletInfo.jars = [jar(0, [utxo({ utxo: 'reviewed:0', value: 25_000 })])]
+    const { result, rerender } = renderHook(() => useCreateFidelityBondWizard(true, vi.fn(), 'wallet.jmdat'))
+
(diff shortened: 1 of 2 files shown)


theborakompanioni (approved):
tACK. Beautiful 💪

PR #1468, opened 2026-08-30
feat(jar-details): freeze and unfreeze selected utxos in one request

### LINKED ISSUE #1437: feat: freeze/unfreeze multiple utxos at once
Currently, freezing/unfreezing `n` utxos requires `n` network requests.
It would be nice to freeze/unfreeze multiple utxos with one request.

This requires a backend change!

GuTS805:
## What does this PR do?

Freezing or unfreezing n selected utxos in the jar details view fired n parallel requests through `Promise.allSettled`. That also meant a batch could land half-applied, so the UI had to count fulfilled vs rejected and show a third "some of them worked" state.

The wallet API now has a `freeze-batch` endpoint that applies every entry or none of them, so this sends a single request and both handlers collapse to a plain success or error path.

- Closes #1437

### Behaviour change

Partial outcomes are no longer possible. Previously a batch could end up with some utxos frozen and some not, and the toast reported the split. Now it either all lands or nothing does, and the toast is success or error. No i18n keys are orphaned by this, both `toast_*_error` keys are still used on the error path.

One small fix came along with it: the error toast counted `selectedUtxos` while the request actually operated on `eligibleUtxos`, which differ whenever same-address utxos get auto-included. Both counts now use `eligibleUtxos`.

### Scope

I left the freeze loops in `useCreateFidelityBondWizard.ts` and `useFidelityBondSweep.ts` alone for now. There's an open question on #1467 about whether the "Freeze Remaining UTXOs" step should go away entirely, and batching loops that might be deleted seemed like wasted churn. Happy to follow up once that's settled.

### Tests

`WalletJarsDetailsContent.test.tsx` now asserts a single batch request goes out with the right entries and freeze flag, plus a new test that a rejected batch surfaces an error and does not refetch wallet info. I checked the assertions actually bite by hardcoding the freeze flag, which fails the unfreeze test rather than passing regardless.

## Visual Demo

No UI changes. Same buttons, same toasts, only the request shape and the partial-state case differ.

### DIFF
--- a/src/components/wallet/WalletJarsDetailsContent.tsx
+++ b/src/components/wallet/WalletJarsDetailsContent.tsx
@@ -1,5 +1,5 @@
 import { useCallback, useEffect, useMemo, useState } from 'react'
-import { freezeMutation } from '@joinmarket-webui/joinmarket-api-ts/@tanstack/react-query'
+import { freezebatchMutation } from '@joinmarket-webui/joinmarket-api-ts/@tanstack/react-query'
 import { useMutation } from '@tanstack/react-query'
 import type { RowSelectionState } from '@tanstack/react-table'
 import type { TFunction } from 'i18next'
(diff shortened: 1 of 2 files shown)


theborakompanioni (approved):
tACK. Awesome 🚀

src/components/wallet/WalletJarsDetailsContent.tsx
```diff
@@ -182,21 +160,10 @@ export const UtxosContent = ({ enabled, walletFileName, addressSummary, jar }: U
...
-          toast.warning(errorMessage)
-        } else {
-          toast.error(errorMessage)
-        }
-      }
+      await unfreezeUtxos.mutateAsync(eligibleUtxos)
+      toast.success(t('jar_details.utxo_list.toast_unfreeze_success', { count: eligibleUtxos.length }))
     } catch (_ignoredOnPurpose) {
-      toast.error(t('jar_details.utxo_list.toast_unfreeze_error', { count: selectedUtxos.length }))
+      toast.error(t('jar_details.utxo_list.toast_unfreeze_error', { count: eligibleUtxos.length }))
```
theborakompanioni, line 166: NIT: Should the `toast_unfreeze_error_other` translation be adapted?
Currently it says "Some selected UTXOs could not be unfrozen. Please try again."..

Simply "Selected UTXOS could not be unfrozen. Please try again."?
Something different?

theborakompanioni: Ahh.. still used in UTXO selection dialog of send flow. Should/can be done when batch is used everywhere :pray:

Issue #1465, opened 2026-08-28
Uppercase bech32 addresses are rejected, and slip past the reuse / duplicate-destination checks


CapThunder19:
Bech32 addresses are case-insensitive BIP-173 says decoders MUST accept both the all-lowercase and all-uppercase form, and it actually recommends uppercase inside QR codes because it packs more densely in alphanumeric mode. Jam never normalizes case anywhere, and that shows up in a few places.

The visible one: a valid uppercase Taproot address is rejected as invalid.

```ts
const P2TR = 'bc1p5cyxnuxmeuwuvkwfem96lqzszd02n6xdcjrs20cac6yqjjwudpxqkedrcr'

isValidAddress(P2TR)                 // true
isValidAddress(P2TR.toUpperCase())   // false  <- same address
```

The cause is upstream in `bitcoin-address-validation`: it chooses bech32m vs bech32 with a case-sensitive `address.startsWith('bc1p')`, so `BC1P…` gets checksummed with the wrong variant and fails. Uppercase `BC1Q…` works fine, which makes it feel arbitrary from the user's side same paste, same scanner, works for SegWit v0 and fails for Taproot with a bare "invalid address".

Two quieter ones, both because addresses are compared as raw strings:

- `isReusedAddress` (`src/lib/formValidation.ts`) does a plain `addressSummary[value]` lookup, and that map is keyed by the lowercase addresses the API returns. An uppercase form of one of the wallet's own used addresses just misses, so the reuse warning never fires.
- The sweep scheduler's duplicate-destination check (`src/components/sweep/destinationValidation.ts`) counts exact trimmed strings, so `bc1q…` in one field and `BC1Q…` in another aren't seen as the same destination — two tumbler phases can end up paying the same address.

I think one small helper fixes all three: normalize at the input boundary, lowercasing **only** when the string is all-uppercase and starts with a bech32 HRP (`BC1`/`TB1`/`BCRT1`). Base58 addresses are case-sensitive, so those must be left alone. Applying it in `parseBip21Uri`, `destinationAddressField`'s transform, and the sweep's `normalizeAddress` should cover it.

Unrelated but in the same file `src/lib/bip21.ts:28` uses `split('?', 2)`, and JS truncates on a split limit instead of keeping the remainder, so anything after a second `?` in the URI is silently dropped.

Happy to open a PR if this looks right.

PR #1470, opened 2026-08-31
fix(validation): accept uppercase bech32 addresses

CapThunder19:
## What does this PR do?

Bech32 addresses are case-insensitive BIP-173 requires decoders to accept the all-uppercase form, and recommends it inside QR codes since alphanumeric mode encodes it more densely. Jam compared addresses as raw strings everywhere, which broke three things:

- **An uppercase taproot address is rejected as invalid.** `bitcoin-address-validation` selects bech32m vs bech32 with a case-sensitive `startsWith('bc1p')`, so `BC1P…` gets checksummed with the wrong variant. Uppercase `BC1Q…` works fine, which makes it look arbitrary same paste, works for segwit v0, fails for taproot.
- **The reuse warning silently misse.** `isReusedAddress` does a plain `addressSummary[value]` lookup, and that map is keyed by the lowercase addresses the API returns.
- **The sweep duplicate-destination check misses too** It counts exact strings, so `bc1q…` in one field and `BC1Q…` in another aren't seen as the same destination, and two tumbler phases could pay one address.

Adds `normalizeBitcoinAddress()` to `formValidation.ts`, which lowercases only when the string is all-uppercase *and* carries a bech32 HRP base58 is case-sensitive so it's left untouched, and mixed case stays invalid per BIP-173. It's applied at the input boundaries so the canonical form is what the form stores and submits, plus in the shared predicates as a backstop.

Two smaller things in `bip21.ts` while I was in there:

- Parameter names are read case-insensitively now. A wallet that uppercases a whole QR URI uppercases `AMOUNT=` too, and the old code dropped the amount while still filling in the address. **This goes slightly beyond BIP21, which only defines lowercase names** .
- `split('?', 2)` truncated the query instead of keeping the remainder, so anything after a second `?` was lost.

Tests cover uppercase forms of all six bech32 flavours, base58 and mixed-case safety, case-insensitive reuse detection, uppercase BIP21 URIs, and case-differing sweep duplicates.

#1465

### DIFF
--- a/src/components/sweep/destinationValidation.ts
+++ b/src/components/sweep/destinationValidation.ts
@@ -1,8 +1,8 @@
 import type { TFunction } from 'i18next'
 import type { AddressSummary } from '@/context/JamWalletInfoContext'
-import { isReusedAddress, isValidAddress } from '@/lib/formValidation'
+import { isReusedAddress, isValidAddress, normalizeBitcoinAddress } from '@/lib/formValidation'
 
-const normalizeAddress = (value: string): string => value.trim()
+const normalizeAddress = (value: string): string => normalizeBitcoinAddress(value)
 
 export const buildDestinationErrors = (
   addresses: string[],
--- a/src/components/sweep/destinationValidation.ts
+++ b/src/components/sweep/destinationValidation.ts
@@ -2,14 +2,12 @@ import type { TFunction } from 'i18next'
 import type { AddressSummary } from '@/context/JamWalletInfoContext'
 import { isReusedAddress, isValidAddress, normalizeBitcoinAddress } from '@/lib/formValidation'
 
-const normalizeAddress = (value: string): string => normalizeBitcoinAddress(value)
-
 export const buildDestinationErrors = (
   addresses: string[],
   addressSummary: AddressSummary,
   t: TFunction,
 ): Array<string | undefined> => {
-  const normalizedAddresses = addresses.map((address) => normalizeAddress(address))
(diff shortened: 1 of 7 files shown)


src/lib/formValidation.ts
```diff
@@ -10,12 +10,22 @@ import type { AddressSummary } from '@/context/JamWalletInfoContext'
 import type { BitcoinAddress, BlockHeight, JarIndex } from '@/types/global'
 import { isValidInteger } from './utils'
 
+const BECH32_HRP_PREFIXES = ['bc1', 'tb1', 'bcrt1']
```
theborakompanioni: Can we use https://github.com/ruigomeseu/bitcoin-address-validation#address-information to detect `bech32`, or overkill?

src/components/sweep/destinationValidation.ts
```diff
@@ -1,8 +1,8 @@
 import type { TFunction } from 'i18next'
 import type { AddressSummary } from '@/context/JamWalletInfoContext'
-import { isReusedAddress, isValidAddress } from '@/lib/formValidation'
+import { isReusedAddress, isValidAddress, normalizeBitcoinAddress } from '@/lib/formValidation'
 
-const normalizeAddress = (value: string): string => value.trim()
+const normalizeAddress = (value: string): string => normalizeBitcoinAddress(value)
```
theborakompanioni: Just use `normalizeBitcoinAddress` directly?

theborakompanioni (dismissed):
lgtm! Only nits. 🚀 
Awesome fix.

theborakompanioni (approved):

Issue #1469, opened 2026-08-31
Logged out while the refresh token is still valid


CapThunder19:
I was reading through the auth code and I think `RefreshApiToken` (`src/App.tsx:309-359`) can end a session that the backend is still perfectly happy with. Two ways:

**The renewal timer resets on page reload.** The access token lasts 30 minutes and Jam renews it on a ~16.9 minute timer. But the timer starts from zero on every reload, while the token itself survives in `sessionStorage` and we never record when it was issued. So if you log in at 0:00 and reload at 0:16, the next renewal is scheduled for 0:33 and the token dies at 0:30. The next request 401s, `unauthorizedResponseInterceptor` clears auth, and you're back at the login screen. The renewal never fires. Reload more often than every ~17 minutes and it never fires at all, so you get logged out every 30 minutes no matter what.

This is invisible in development, which is probably why it hasn't come up: `.env.development` sets `VITE_JM_API_AUTH_TOKEN_EXPIRY_SECONDS=15`, so renewal runs every ~11 seconds and a reload can't land between two of them. With the `.env` value of 1800 the window is 17 minutes wide.

**A single failed renewal ends the session.** The client isn't configured with `throwOnError`, so a dropped connection resolves as `{ error }` rather than throwing, and lands here:

```ts
if (!response.data) {
  clearAuthAndQueryCache()
```

A Tor hiccup, a node restart, or a laptop waking from sleep is treated the same as "your refresh token was rejected" no retry, and the toast is behind `isDevMode`, so in a release build you just get moved to the login screen with no explanation.

Both look avoidable given what jmwalletd already tells us. `expires_in` comes back on every grant and we drop it in favour of a hardcoded constant, and with `ACCESS_TOKEN_EXPIRY_SECONDS = 1800` / `REFRESH_TOKEN_EXPIRY_SECONDS = 14400` there are usually hours of refresh validity left at the point we give up. The `/token` endpoint is even declared `Depends(require_auth_allow_expired)`, so an expired access token is no obstacle to refreshing we clear auth instead of trying.

Rough idea for a fix: store `expiresAt` from `expires_in` alongside the tokens (falling back to the current constant when a backend omits it, since it's optional in the API types), schedule the renewal off that instead of a fixed interval, and only clear auth when the refresh call actually comes back 401 `invalid_token` — retry with backoff on transport errors while there's still refresh validity left.

[...]

PR #1471, opened 2026-08-31
fix(auth): schedule token renewal from expiry, not a fixed interval

CapThunder19:
## What does this PR do?

Fixes the token-refresh loop so it doesn't log users out while their refresh token is still valid.

Two related problems in `RefreshApiToken`:

1. **Renewal timer resets on reload.** It ran on a fixed interval counted from component mount, with no record of when the current token was actually issued. Since the token itself survives in `sessionStorage` across a reload but the timer doesn't, reloading late enough into the token's life reschedules the next renewal *after* the token has already expired logging the user out even though the refresh token is still good for hours.
2. **Any failed renewal cleared auth locally**, treating a dropped connection or the node restarting the same as an explicit rejection. A genuinely rejected refresh token already comes back as a 401 `invalid_token`, which the shared response interceptor handles this was piling redundant, overly aggressive logout on top of that.

Fix: `authStore` now tracks `expiresAt` (from the grant response's `expires_in`, with the existing constant as fallback), renewal is scheduled off the actual remaining time instead of a fixed interval, and a failed renewal just retries next cycle instead of clearing auth.

#1469 

- new unit tests covering the reload-timing scenario and the no-clear-on-transient-failure behavior (`authStore.test.ts`, `utils.test.ts`, `App.test.tsx`)
- full suite (917 tests), typecheck, lint, and build all passing


src/App.tsx
```diff
@@ -329,10 +344,9 @@ function RefreshApiToken() {
         })
 
         if (!response.data) {
-          clearAuthAndQueryCache()
-
+          const message = getErrorReason(response.error, 'Unknown error.')
```
theborakompanioni, line 347: What are the consequences of removing `clearAuthAndQueryCache`?
Will a user see lots of error messages as a result instead of being logged out?

CapThunder19: no visible change for users either way the toast was always dev-only

CapThunder19: real change is that network blip no longer force logs out, it just retries a dead token still gets caught right away by the existing 401 interceptor.

theborakompanioni: k. Thanks, I understand. What do you think about clearing it nonetheless and informing the user about it (now dev only)?

But.. can be done in a follow-up.. if we decide to do it.

theborakompanioni (approved):
Good work @CapThunder19 
Thank you 🙏

PR #1473, opened 2026-09-02
feat(sweep): show days and hours in wait duration

### LINKED ISSUE #1472: ui: human readable "wait period before next action"
Display the value for "wait period before next action" on the Sweep page in a more human friendly way

Image

Currently only shown in minutes and seconds.
Should also show days and hours.
Can drop the "seconds" part if  >=1h

Value in the screenshot is 9,119m 3s, that is over 6 days.

NormanSMA:
Fixes #1472

### Problem

On the Sweep page a long wait before the next action rendered as `9,119m 3s`. You cannot read that without doing the arithmetic — it is about six days.

`formatDuration()` in `src/components/sweep/scheduleUtils.ts` only handled minutes and seconds, so everything accumulated into minutes no matter how long the wait was.

### Change

Show at most the two largest non-zero units. Seconds are dropped from an hour upwards, minutes from a day upwards, as the issue asks.

| Duration | Before | After |
|---|---|---|
| 45s | `45s` | `45s` |
| 2m 5s | `2m 5s` | `2m 5s` |
| 2h 30m 15s | `150m 15s` | `2h 30m` |
| ~6 days | `9,119m 3s` | `6d 7h` |

Short durations render exactly as before. The existing behaviour is preserved: negative input is still clamped to zero, fractional seconds still round up, and numbers still go through `toLocaleString()`.

### Translations

The four new keys are added to `en` only, matching how the existing `duration_*` keys are handled — they live in `en/translation.json` alone and the other locales fall back to it.

### Tests

`formatDuration` had no tests, so this adds the first ones — seven cases covering the boundaries (below a minute, negative input, fractional rounding, the hour boundary, the day boundary) and the exact duration from the issue.

The `t()` double renders the key together with its interpolated values, so each assertion shows both which key was selected and what was passed to it. If key selection regresses, the failure names the key.

### Verification

```
npx vitest run --project unit src/components/sweep/scheduleUtils.test.ts
  → 11 passed (4 existing + 7 new)
npx tsc -b        → clean
npx prettier      → clean
npx eslint        → clean
```

### Note

I left the `// TODO: move to utils and use Intl.DurationFormat (es2025) if available` comment in place — this change doesn't address it, and `Intl.DurationFormat` would be a separate refactor.

### DIFF
--- a/src/components/sweep/scheduleUtils.test.ts
+++ b/src/components/sweep/scheduleUtils.test.ts
@@ -565,8 +565,12 @@ describe('scheduleUtils', () => {
 describe('formatDuration', () => {
   // Minimal stand-in for i18next: renders the key with its interpolated values,
   // so each assertion shows both which key was picked and what was passed to it.
+  // Keys are sorted so the assertions stay focused on which translation key was
+  // selected and with what values -- i18next treats option key order as
+  // irrelevant, so reordering the object literal must not break these tests.
   const t = ((key: string, values: Record<string, string> = {}) =>
     `${key}(${Object.entries(values)
+      .toSorted(([a], [b]) => a.localeCompare(b))
       .map(([k, v]) => `${k}=${v}`)
       .join(',')})`) as unknown as Parameters<typeof formatDuration>[1]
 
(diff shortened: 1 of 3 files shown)


theborakompanioni (approved):
Great 🚀 
Thank you  @NormanSMA 🙏

PR #1479, opened 2026-09-08
fix(sweep): keep stale sweep plans visible

### LINKED ISSUE #1346: feat(sweep): ability to display stale sweep plans
Could we keep this response visible instead of dropping it? cos the the backend’s first post-restart response is `stale:true`, `status:'failed`, with the restart error. Dropping it exposes the fresh-plan form, whose `force:true` request can overwrite the failed plan before the user sees or acknowledges it.

_Originally posted by @parrth20 in https://github.com/joinmarket-webui/jam/pull/1339#discussion_r3632511257_

hoowertsg19:
## What does this PR do?

Keeps stale sweep plans visible instead of falling back to the fresh-plan form.

### Summary

- Displays stale failed plans and their backend error.
- Prevents Start and Renew actions for stale plans while keeping Stop available when the scheduler is actually running.
- Keeps the existing clear action available before another plan can be created.
- Adds regression coverage for the stale failed-plan flow.

Fixes #1346

### Tests

- Added regression coverage for `stale: true` + `status: failed`.
- Verified the fresh-plan form is not rendered for stale plans.
- Verified Start and Renew are unavailable for stale plans.
- Verified Stop remains available when the scheduler is actually running.
- Verified no replacement plan is created.
- ESLint passed.
- Unit tests passed.
- Coverage tests passed.
- Production build passed.
- Storybook build passed.

## Visual Demo

Current JoinMarket NG reconciles stale running plans during startup, so the
`stale: true` response is no longer naturally observable after a normal restart.

The screenshots below use a controlled regtest reproduction of the same backend
state covered by the regression test.

### After

The stale failed plan remains visible so the restart error can be inspected
before the plan is cleared.

image

The corresponding backend response contains `stale: true` and the restart error.

image

image


src/components/sweep/SweepPage.tsx
```diff
@@ -201,10 +201,10 @@ export const SweepPage = ({ walletFileName }: SweepPageProps) => {
   const currentSchedule = useMemo(() => {
     if (getScheduleQuery.error) return
     if (getScheduleQuery.data === undefined) return
-    if (getScheduleQuery.data.stale === true) return
 
     return toSchedule(getScheduleQuery.data, walletInfo.jars)
   }, [getScheduleQuery.error, getScheduleQuery.data, walletInfo.jars])
+  const isCurrentScheduleStale = getScheduleQuery.data?.stale === true
```
theborakompanioni: Can `stale` be added to `currentSchedule.summary`?

src/components/sweep/SweepPage.tsx
```diff
@@ -355,9 +355,13 @@ export const SweepPage = ({ walletFileName }: SweepPageProps) => {
...
 
-                <SweepScheduleProgress schedule={currentSchedule} debug={isDeveloperMode} />
+                <SweepScheduleProgress
+                  schedule={currentSchedule}
+                  error={getScheduleQuery.data?.error}
+                  debug={isDeveloperMode}
+                />
 
-                {schedulerRunning && currentSchedule.summary.status.running ? (
+                {!isCurrentScheduleStale && schedulerRunning && currentSchedule.summary.status.running ? (
```
theborakompanioni: Can a schedule in state `running` ever be `stale`?
If the scheduler is still running - can a stale schedule be stopped?

src/components/sweep/SweepPage.tsx
hoowertsg19: Done — stale is now part of ScheduleSummary and is populated by toSchedule() using plan.stale === true. SweepPage now uses currentSchedule.summary.stale instead of reading the stale state separately from the raw API response. I also added coverage for true, false, and undefined.

src/components/sweep/SweepPage.tsx
hoowertsg19: Good point. A stale response normally represents a persisted plan for which no live runner exists, so stale + running scheduler should not normally occur. I removed the stale guard from Stop: Stop is now controlled only by schedulerRunning && currentSchedule.summary.status.running.

Start and Renew remain unavailable for stale plans. I also updated the regression test to verify the defensive case that Stop remains available if the scheduler is reported as running even when the plan is marked stale.

theborakompanioni (approved):
lgtm! 🚀 
Thank you  @hoowertsg19  🙏

PR #1481, opened 2026-09-10
feat(send): list the utxos a sweep will spend in the confirm dialog

GuTS805:
## What does this PR do?

Follow-up to the review on #1463, where it was suggested it'd be nice to show the utxos in the payment confirm dialog before broadcasting (see #807 for prior art).

The sweep already pins its exact inputs via `input_utxos`, but the confirm dialog only showed the total amount, so there was no way to review which specific coins were about to be spent. This resolves the pinned utxos against the source jar's known utxo list and lists each one with its address and value, falling back to the raw outpoint if a utxo isn't found in the jar (e.g. if it was consumed elsewhere in the meantime).

Only shows up for sweeps (`sweepUtxos`); direct sends are unaffected.

## Visual Demo

The dialog now shows a "UTXOs (n)" row between the amount and the collaborator count, listing address + balance for each utxo the sweep will spend.


theborakompanioni (approved):
tACK. 🚀 

Would it be nicer to include the label and hide it by default in an accordion?

Current:

Image

Previous: 

Image


But ACK nonetheless! Awesome. 🧡

PR #1482, opened 2026-09-10
fix(send): don't claim a collaborative send completed based on utxo hash alone

### LINKED ISSUE #1475: Send: a collaborative transaction is reported as successful whenever the utxo set changed, even if no CoinJoin happened
**Expected behavior**

After a collaborative send, Jam should only claim the transaction succeeded if it has some evidence that a CoinJoin transaction was actually made.

**Actual behavior**

When the taker stops, Jam decides between "completed" and "ended" purely by comparing the wallet's utxo hash against the hash captured when the taker was started (`SendPage.tsx:459`):

```tsx
{currentPaymentAttempt.utxosHashHex === utxosHashHex ? (
  <Alert variant="warning">  {/* collaborative flow ended */}
) : (
  <Alert variant="success">  {/* collaborative transaction completed */}
)}
```

The hash is snapshotted when the coinjoin is started (`SendPage.tsx:322-327`):

```tsx
setCurrentPaymentAttempt({
  createdAt: Date.now(),
  walletFileName,
  utxosHashHex,
  data,
})
```

So the actual test is "did the wallet's utxo set change at any point while the taker was running", not "did my coinjoin produce a transaction". Anything that changes the utxo set during that window flips a failed attempt into a success message. The window is the entire taker run, which can be many minutes while it waits for counterparties and retries.

The success text is unconditional:

> **Collaborative transaction completed**
> JoinMarket finished the collaborative transaction successfully.

The failure text is appropriately hedged ("no new completed transaction was confirmed yet. Please verify the recipient and transaction history."), so the strong claim sits on the unreliable side of the branch.

1. Start a collaborative send.
2. While the taker is running, have any unrelated change land in the wallet's utxo set. Receiving a deposit is enough, and needs no action from the user.
3. Let the coinjoin fail (no counterparties, aborted, etc).
4. When the taker stops, Jam shows "Collaborative transaction completed - JoinMarket finished the collaborative transaction successfully", even though nothing was sent.

**Specifications**

- Version: 2.0.0-beta.3 (devel @ b4d0ca22)
- Platform: n/a, this is code-level and backend agnostic
- Browser: n/a

---

**This is not theoretical**

The existing suite has a test for the failure branch ("shows the collaborative ended alert when utxos are unchanged") but none for the success branch. I patched `SendPage.test.tsx` locally to make the wallet's `utxosHashHex` differ from the attempt's, with `takerRunning: false` and nothing else changed:

```tsx
it('PROBE: reports success purely because the utxo hash changed', async () => {
  mocks.currentPaymentAttemptPresent = true
  // nothing coinjoin-related happened; only the wallet utxo set changed
  mocks.currentUtxosHashHex = 'hash-after-unrelated-deposit'

  render(<SendPage walletFileName="wallet.jmdat" />)

  expect(screen.getByText('send.alert_collaborative_completed_title')).toBeInTheDocument()
  expect(screen.queryByText('send.alert_collaborative_ended_title')).not.toBeInTheDocument()
})
```

It passes. No transaction, no txid, no coinjoin of any kind, and Jam reports success. (Reverted, not part of any PR.)

**The direct send path in the same file does this properly**

The non-collaborative branch has a real transaction to work with and uses it (`SendPage.tsx:274-293`):

```ts
onSuccess: (result: DirectSendResult, data: SendFormValues) => {
  const tx = result.response.txinfo as Required<JmTxInfo>
  jmTxStore.getState().add(tx)
  ...
  description: t('send.alert_payment_successful', { amount: ..., address: ..., txid: tx.txid }),
```

So a direct send reports success against an actual txid, while the collaborative send reports the same level of confidence against "some utxo changed". That's the gap.

**A definitive signal already exists and isn't used**

The comment above the block says clientserver has no direct way to verify success, which is fair, but Jam already receives transaction notifications over the websocket and stores them (`JmWebsocketContextProvider.tsx:27-37`):

```ts
const isKnownTx = jmTxStore.getState().get(message.txid) !== undefined
jmTxStore.getState().add(message.txdetails)
```

`jmTxStore` is keyed by txid and holds the full `txinfo`. `App.tsx` already watches it to trigger `AFTER_TX_MESSAGE` refetches, so the plumbing is there. A transaction arriving during the attempt window, with inputs from the source jar, is much closer to evidence than "some utxo somewhere changed".

**Two secondary notes on the hash itself**

`combinedUtxosHash` (`JamWalletInfoContextProvider.tsx:132-136`) is worth a look independently of the above:

```ts
const combinedUtxosHash = (utxos: Utxo[]) => {
  const utxoIds = utxos.map((it) => hexToBytes(it.utxo.split(':', 1)[0]))
  const combinedUtxoIds = concatBytes(...utxoIds)
  return sha256(combinedUtxoIds)
}
```

- It concatenates in array order and nothing sorts the utxos (`useQueryUtxos` returns the API array as-is), so the hash is order dependent. If `listutxos` ever returns the same set in a different order, the hash changes with no state change, which flips this branch to success and also invalidates the `displaywallet` query key.
- `split(':', 1)[0]` drops the vout, so the hash identifies a multiset of txids rather than a set of outpoints. I could not construct a reachable collision that hides a real transition, so this looks harmless today, but it makes the value weaker than it reads.

**Impact**

Telling someone their payment went through when it did not is the wrong way for this to fail. The user closes the alert, and only finds out later, if they check the transaction history. Given the failure branch already tells the user to go and verify, the success branch making a stronger claim on weaker evidence seems backwards.

This isn't new, the comparison came in with #1245 and #1413 only touched the surrounding presentation, so it's long standing rather than a recent regression. Not a duplicate of #1447 or #1464 (both about redundant refetches driven by the same value), #516 or #1374 (both about how the success screen looks, not how it is decided), #1188, or #996.

**Proposed fix direction**

Require some transaction evidence before claiming success, rather than treating any utxo change as proof. Something like: a tx recorded in `jmTxStore` after `currentPaymentAttempt.createdAt` that spends from the attempt's source jar. If no such evidence exists, fall back to the hedged "ended" message rather than the success one, since that wording is already correct for the unknown case.

GuTS805:
## What does this PR do?

Fixes #1475.

The collaborative-send outcome alert was deciding "completed" vs "ended" purely by comparing the wallet's utxo hash before/after the taker run. Any unrelated utxo change during that window (an incoming deposit, for example) flips a failed attempt into a "transaction completed successfully" message, since the hash isn't actual evidence a coinjoin happened. See the issue for a reproduction.

This is an interim fix: until there's a real signal to check the outcome against (opened joinmarket-ng/joinmarket-ng#627 for a proper status endpoint), both branches are collapsed into one neutral "flow stopped" message that points the user at their transaction history instead of asserting success or failure.

## Visual Demo

Same alert now shows regardless of whether the utxo hash changed. No more unconditional "Collaborative transaction completed" text without evidence.


src/components/send/SendPage.tsx
```diff
@@ -455,46 +455,31 @@ export const SendPage = ({ walletFileName }: SendPageProps) => {
...
+                  <AlertTriangleIcon />
+                  <AlertTitle>{t('send.alert_collaborative_stopped_title')}</AlertTitle>
+                  <AlertDescription className="flex flex-col gap-2">
+                    <div>{t('send.alert_collaborative_stopped_description')}</div>
+                    <div>
+                      <Button
+                        variant="outline"
+                        onClick={() => {
+                          clearCurrentPaymentAttempt()
+                        }}
```
theborakompanioni, line 477:
```suggestion
                        onClick={() => {
                          setFormId((current) => current + 1)
                          clearCurrentPaymentAttempt()
                        }}
```

Should the form be cleared afterwards?
(Before it was only cleared if "assumed" successful)

GuTS805: Done👍

theborakompanioni (approved):
> @theborakompanioni Ready for review. this is the interim fix discussed in #1475, tracked the backend follow-up separately ([joinmarket-ng/joinmarket-ng#627](https://github.com/joinmarket-ng/joinmarket-ng/issues/627)).

This is the right move. Thank you @GuTS805  🙏

PR #1489, opened 2026-09-20
fix(fb): pin the bond utxo when renewing or moving a fidelity bond

### LINKED ISSUE #1488: Renew / Move-to-jar still sweep
## Expected behavior

Renewing or moving a fidelity bond spends exactly that bond UTXO.nothing else in the jar should be affected, and if the bond can't be spent the action should fail rather than report success.

## Actual behavior

useFidelityBondSweep still does the old "freeze everything else, then direct-send with mixdepth + amount_sats: 0" dance, with no input_utxos. The freeze list is a snapshot of Jam's cached UTXOs and nothing re-checks the jar before broadcast the same shape #1460 described for the create wizard, which #1467 fixed by pinning input_utxos. This hook was left out

It's actually worse here than in #1460. In joinmarket-ng's sweep path (jmwallet/wallet/spend.py, present in 0.35 through 0.39.2), a mixdepth sweep spends regular unfrozen coins and only falls back to an expired bond when there are no regular coins at all. Jam's approach only works because it froze everything else. So if any unfrozen coin lands in the jar between the snapshot and the broadcast 

Renew: the new coin gets time-locked at the chosen date, the actual bond is not renewed and is left unfrozen in the jar, Jam freezes txid:0 (the wrong output), shows the success step and toasts "renewed". The confirm screen showed utxo.value the bond amount which isn't what got locked. The leftover unfrozen expired bond then also blocks Send from that jar.
Move to jar: the new coin moves, the bond stays.
The window is bounded (backend tx-monitor polls every 5 s + Jam's refetch + N sequential freeze round-trips, each over Tor on a typical node), not user-paced like the create wizard was but the outcome is a wrong coin locked for months/years and a false success.

Have a jar with an expired fidelity bond and at least one other unfrozen UTXO.
Open Renew, pick a lock date, confirm.
While the freeze requests are in flight, have a new unfrozen UTXO arrive in that jar (deposit to a receive address of that jar, or hold the freeze calls in devtools and send the deposit).
The broadcast spends the new UTXO to the timelock address. The bond is untouched and now unfrozen; the dialog reports success.
Also reproducible as a unit test in the hook's existing harness: mock freezeUtxo to add a UTXO to the backend state on each call, model direct-send with the selection rule above, run sweep() the request body has no input_utxos, the spent input is the deposit, the hook returns a result.

## Specifications

Version: devel @ b731b078 (v2.0.0-beta.3+), joinmarket-api-ts 0.37.1
Platform: any; backend joinmarket-ng ≥ 0.37 (has input_utxos)
Browser: n/a
Additional context

## Relevant code:
 src/components/earn/fidelity-bond/useFidelityBondSweep.ts L36-39 (snapshot), L63-87 (freeze loop + sweep without input_utxos). Compare with the fixed create flow at useCreateFidelityBondWizard.ts L283-292, which even carries a // TODO: refactor: use the sweep functionality from useFidelityBondSweep here so this hook is meant to become the canonical one.

## Suggested fix: 
pass input_utxos: [utxo.utxo], keep only "unfreeze bond → send → re-freeze on failure", drop the freeze/unfreeze-others loops (the source of the #1048/#1397 rollback bugs), bail out if the bond is no longer in the jar, and add the two regression tests #1467 added for create. resolve_input_utxos on the backend admits expired bonds and requires them unfrozen, so this is supported without a backend change.

CapThunder19:
## What does this PR do?

Renew and move-to-jar (useFidelityBondSweep) still froze every other utxo in the jar and then swept the whole mixdepth the snapshot problem #1460 described and #1467 fixed for the create wizard, but this shared hook was left out.

It's worse here: joinmarket-ng's mixdepth sweep spends regular unfrozen coins first and only falls back to an expired bond when there are none. So if a coin lands in the jar after the snapshot (deposit, coinjoin change, stale cache), the backend spends that coin instead of the bond on renew the deposit gets time-locked, the bond stays unfrozen in the jar, and Jam reports success.

The sweep now passes input_utxos: [bond] (backend ≥ 0.37, already used by the Send page and create wizard), so exactly the bond is spent. The freeze/unfreeze-others loops and their rollback go away; only "unfreeze the bond, re-freeze on failure" remains. If the bond has left the jar when the user confirms, the hook errors out instead of broadcasting (same guard as #1467).

Fixes #1488 

## Changes
- useFidelityBondSweep.ts: pin input_utxos, drop freeze-others loops, add bond-gone guard.
- ranslation.json: new key earn.fidelity_bond.error_bond_not_in_jar.
- useFidelityBondSweep.test.ts: existing tests updated; two new tests mirroring #1467 (utxo arriving mid-flow is not spent; no broadcast when the bond is gone). Both fail on devel, pass here.


theborakompanioni (approved):
tACK! Awesome 🚀

